PR #219, opened 2024-02-14
feat: BTree example

AlpinYukseloglu:
Adds an example for running insertions on a standard BTree.


examples/btree/Cargo.lock
gakonst, line 1: Cargo lock I think shouldn't be needed here.

puma314:
Closing this for now, I think someone else is making a Merkle Tree example!

PR #252, opened 2024-02-15
fix: sha compress

hidenori-shinohara:
https://github.com/succinctlabs/sp1/pull/216 introduces a range check for `Add5Operation`.

Sha compress doesn't call `eval` though it calls `populate`. This causes an interaction bug. 

It is a TODO item that we need to call `eval`, and we're trying to get it to asap. this is a temporary fix.

### DIFF
--- a/core/src/syscall/precompiles/sha256/compress/trace.rs
+++ b/core/src/syscall/precompiles/sha256/compress/trace.rs
@@ -139,6 +139,13 @@ impl<F: PrimeField> MachineAir<F> for ShaCompressChip {
                 // index, which is definitely feasible but we haven't gotten to yet. If we call
                 // Add5Operation::populate, it creates interactions, and without the matching eval
                 // call, it will cause a panic. We plan to fix this really soon.
+                //
+                // This is how we should call populate here.
+                //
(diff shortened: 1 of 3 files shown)


core/src/syscall/precompiles/sha256/compress/trace.rs
```diff
@@ -134,9 +134,17 @@ impl<F: PrimeField> MachineAir<F> for ShaCompressChip {
                 let e_not_and_g = cols.e_not_and_g.populate(output, e_not, g);
                 let ch = cols.ch.populate(output, e_and_f, e_not_and_g);
 
-                let temp1 = cols
-                    .temp1
-                    .populate(output, h, s1, ch, event.w[j], SHA_COMPRESS_K[j]);
+                // TODO: This is a hack to avoid calling Add5Operation::populate. We currently don't
```
puma314, line 137: nit can you leave the old "correct" code in the comment so people who what it should be fixed to?

hidenori-shinohara: good point! added.

puma314 (approved):

PR #253, opened 2024-02-16
feat: automatic ordering of execution

tamirhemo:


core/src/stark/air.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// An table for `less than` operation on field elements.
+    FieldLTU(FieldLTUChip),
+    /// A table for initializing the memory state.
+    MemoryInit(MemoryGlobalChip),
+    /// A table for finalizing the memory state.
+    MemoryFinal(MemoryGlobalChip),
+    /// A table for initializing the program memory.
+    ProgramMemory(MemoryGlobalChip),
+
+    ShaExtend(ShaExtendChip),
```
jtguibas, line 74: nit: the spacing here is a bit weird, can we just have a comment per element in the enum?

PR #261, opened 2024-02-16
feat: Add program build script

preston-evans98:
This draft adds a build script which automatically compiles the program binary and enforces that the "native" rust compiler toolchain and the succinct toolchain have the same version. This check currently fails, since the repo-wide native version is set to `1.77` and `succinct` toolchain is `1.69` or `1.75` depending on the platform.


ctian1:
hey @preston-evans98, thanks for the PR! Working off of it now in #296.

PR #262, opened 2024-02-17
feat: field extension air support

kevjue:


jtguibas:
I moved a lot of this stuff into my latest PR

Issue #263, opened 2024-02-17
Documentation: sp1 relies on OpenSSL 3 as well as OpenSSL 1.1.1


SadPencil:
[Installation](https://succinctlabs.github.io/sp1/getting-started/install.html#installation) says `Make sure you have Rust and OpenSSL 1.1 installed.` But this is not correct. It ~~actually~~ also requires `libssl.so.3` file which indicates the requirement of OpenSSL 3 (package `libssl3` in Debian/Ubuntu).

Suggestion:

```diff
- Make sure you have Rust and OpenSSL 1.1 installed.
+ Make sure you have Rust, OpenSSL 1.1 and OpenSSL 3 installed.
```

The rest of the contents is the "proof" of my suggestion.

-----------------------------------------------------

The following error message reported by `sp1up` indicates the need of `libssl.so.3` file:
```txt
sp1up: installing sp1 (version nightly, tag nightly)
sp1up: downloading latest cargo-prove
################################################################################################################# 100.0%
/home/user/.sp1/bin/cargo-prove: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
sp1up: command failed: /home/user/.sp1/bin/cargo-prove prove --version
sp1up: installed -
sp1up: installing rust toolchain
/home/user/.sp1/bin/cargo-prove: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
sp1up: command failed: /home/user/.sp1/bin/cargo-prove prove install-toolchain
```

------

This web page shows the file list of the `libssl3` package of version 3.1.5 (on which `libssl-dev` relies), where `libssl.so.3` file is contained.

https://packages.debian.org/trixie/amd64/libssl3/filelist

I've reproduce the installation procedure on Debian 11 (whose OpenSSL version is 1.1.1) and Debian 13 (OpenSSL 3.1.5 for now). Both Linux distributions have `libssl-dev` and rust toolchain installed. In Debian 13 it behaves correctly while in Debian 11 the error prompt above shows.

Debian 13  (has OpenSSL 3, so `sp1up` does not fail):

```bash
user@localhost:~$ apt list --installed | grep -i libssl
libssl-dev/testing,now 3.1.5-1 amd64 [installed]
libssl3/testing,now 3.1.5-1 amd64 [installed]

user@localhost:~$ openssl version
OpenSSL 3.1.5 30 Jan 2024 (Library: OpenSSL 3.1.5 30 Jan 2024)

[...]

PR #268, opened 2024-02-18
docs: correct the OpenSSL version requirement

SadPencil:
It appears that both OpenSSL 1.1 and OpenSSL 3 are required. `sp1up` requires OpenSSL 3 (`libssl.so.3`) while `cargo prove build` requires OpenSSL 1.1 (`libssl.so.1.1`)

This PR enhances the documentation about OpenSSL dependency.

See issue #263 for more details.

### DIFF
--- a/book/SUMMARY.md
+++ b/book/SUMMARY.md
@@ -29,3 +29,7 @@
 - [Basics](./generating-proofs/basics.md)
 
 - [Advanced](./generating-proofs/advanced.md)
+
+# FAQ
+
+- [Installing OpenSSL](./faq/installing-openssl.md)
--- a/book/SUMMARY.md
+++ b/book/SUMMARY.md
@@ -6,6 +6,8 @@
 
 - [Install](./getting-started/install.md)
 
+- [Install OpenSSL](./getting-started/openssl.md)
+
 - [Quickstart](./getting-started/quickstart.md)
 
 # Writing Programs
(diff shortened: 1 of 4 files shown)


SadPencil:
```
These commands are borrowed from this [StackOverflow post](https://askubuntu.com/questions/1102803/how-to-upgrade-openssl-1-1-0-to-1-1-1-in-ubuntu-18-04).
```

@jtguibas 

well, in fact, these commands are mostly borroweded and improved from [this Chinese post](https://www.cnblogs.com/eagle6688/p/17471170.html), since the key thing to adding OpenSSL 1.1.1 library is not to mess up with the shipped one on which all system packages rely, the parameters of the `./config` command matters. It sets the output of OpenSSL binary as `/usr/local/ssl-1.1` which is a good design. This is different from the [StackOverflow post](https://askubuntu.com/questions/1102803/how-to-upgrade-openssl-1-1-0-to-1-1-1-in-ubuntu-18-04): it just **upgrades** the system OpenSSL (1.1.0->1.1.1, acceptable) instead of **adding** a new OpenSSL (1.1 and 3 are not compatible with each other).

Also, are you suggesting deleting [book/getting-started/install.md](https://github.com/succinctlabs/sp1/pull/268/commits/86b695cdf9ac228922144d9dac6562f415959d29#diff-5651f9ad97a91583bb6f443ce583e8ef4300691162da6e54ee7a702329e5293f) file and using your less wordy one [book/faq/installing-openssl.md](https://github.com/succinctlabs/sp1/pull/268/commits/86b695cdf9ac228922144d9dac6562f415959d29#diff-cc9e7d61ba43d20742cb694c638356777ee7c4646c34aaecdd6550b1f43c64e1)? If so I would delete the formmer file.

ctian1:
This should be fixed with recent changes (https://github.com/succinctlabs/sp1/pull/315), the toolchain and cargo-prove are compiled statically now! Openssl should not be necessary anymore.

PR #270, opened 2024-02-19
fix(cli): get-target

0xYYY:
## Issue

To build from source on platforms without pre-built toolchain (e.g. x86_64 macOS), the [Book](https://succinctlabs.github.io/sp1/getting-started/install.html#option-2-building-from-source) instruct to run the `cargo prove build-toolchain` command.

In the `build_toolchain` command, the function `get_target()` will be called.
https://github.com/succinctlabs/sp1/blob/8ac9e54ceb049d79030f2f8831e039af0ca61b30/cli/src/commands/build_toolchain.rs#L121

But it only returns the target triplet for the platforms with pre-built binaries and panic otherwise.
https://github.com/succinctlabs/sp1/blob/8ac9e54ceb049d79030f2f8831e039af0ca61b30/cli/src/lib.rs#L65-L77

Meaning that building from source on other platforms will encounter the following error.
```
Build completed successfully in 0:07:59
info: no toolchain installed for 'succinct'
Succesfully removed existing toolchain.
Found built toolchain directory at ./rust/build/x86_64-apple-darwin/stage2.
Succesfully linked the toolchain to rustup.
thread 'main' panicked at cli/src/lib.rs:74:5:
Unsupported architecture. Please build the toolchain from source.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Solution
This PR uses the [target_lexicon](https://docs.rs/target-lexicon/latest/target_lexicon/index.html) crate to get the target triplet of the host machine.


tamirhemo:
Seems like we may want different functions: one to notify users that the platform is not supported, and one for getting the target when building from source. Any thoughts @ctian1 , @jtguibas ?

0xYYY:
> one to notify users that the platform is not supported

https://github.com/succinctlabs/sp1/blob/8ac9e54ceb049d79030f2f8831e039af0ca61b30/cli/src/lib.rs#L79-L91

Can this function fit the need?

For example, in the `install_toolchain` command, both `get_target()` and `get_toolchain_download_url()` are called. `get_target()` determine the host machine's architecture, while `get_toolchain_download_url()` determine whether the host machine's architecture is supported.
https://github.com/succinctlabs/sp1/blob/81f01f96fd95e262c3ca79a37f8ebe0cc9c6121c/cli/src/commands/install_toolchain.rs#L57-L61

ctian1 (approved):
Thanks for the fix! LGTM

PR #271, opened 2024-02-19
fix: inputs for RSA example 

brozorec:
The RSA example's keys and signature are hardcoded in the program: https://github.com/succinctlabs/sp1/blob/8ac9e54ceb049d79030f2f8831e039af0ca61b30/examples/rsa/program/src/main.rs#L18-L27
That makes this example program look "rigid" by generating an elf file for a specific key and message. 

It seems more natural to have those elements in the script as inputs. That allows us to generalize the program and reuse the generated elf file for the verification of any key-message pair.


examples/rsa/program/src/main.rs
```diff
@@ -3,47 +3,40 @@ sp1_zkvm::entrypoint!(main);
...
-const RSA_2048_PUB_DER: &[u8] = include_bytes!("rsa2048-pub.der");
 
 pub fn main() {
-    let private_key = RsaPrivateKey::from_pkcs8_der(RSA_2048_PRIV_DER).unwrap();
-    let public_key = RsaPublicKey::from_public_key_der(RSA_2048_PUB_DER).unwrap();
-    println!("{:?}{:?}", private_key, public_key);
+    // Read an input to the program.
+    //
+    // Behind the scenes, this compiles down to a custom system call which handles reading inputs
+    let pk_der = sp1_zkvm::io::read::<Vec<u8>>();
```
puma314, line 14: Does this program actually work? I don't think this should work because the io reader doesn't know how many bytes to actually read, if I'm not mistaken. You can either use `read_slice` and pass in a mutable slice with a particular length or you can just read a `RsaPublicKey` if that implements `Serde, Deserde`

brozorec: Well, it does when running in dev mode with `SP1Prover::execute(REGEX_IO_ELF, stdin)` but when trying with `SP1Prover::prove(REGEX_IO_ELF, stdin)` it prints the message from the program and gets killed:
image

I thought it was something related to my machine (Ubuntu 22) because it gets killed too when running the initial RSA example.

> because the io reader doesn't know how many bytes to actually read

This was my understanding as well but then it works when `String` is passed along and `String` boils down to `Vec<u8>` :thinking:

PR #278, opened 2024-02-21
feat: derive `MachineAir`, chip and machine cleanup

tamirhemo:


core/src/stark/air.rs
```diff
@@ -0,0 +1,198 @@
...
+    Secp256k1Add(WeierstrassAddAssignChip<SWCurve<Secp256k1Parameters>>),
+    /// A precompile for doubling a point on the Elliptic curve secp256k1.
+    Secp256k1Double(WeierstrassDoubleAssignChip<SWCurve<Secp256k1Parameters>>),
+    /// A precompile for the Keccak permutation.
+    KeccakP(KeccakPermuteChip),
+    /// A precompile for the Blake3 compression function.
+    Blake3Compress(Blake3CompressInnerChip),
+}
+
+impl<F: PrimeField32> RiscvAir<F> {
```
kevjue, line 96: I feel like if you are going to go the `derive` route, then `included` should also be derived?

core/src/stark/machine.rs
```diff
@@ -56,223 +41,46 @@ pub struct VerifyingKey<SC: StarkGenericConfig> {
...
-        }
-
-        if !shard.last_memory_record.is_empty() {
-            chips.push(self.memory_finalize.as_ref());
-        }
-
-        if !shard.program_memory_record.is_empty() {
-            chips.push(self.program_memory_init.as_ref());
-        }
+    pub fn dyn_chips(&self) -> Vec<ChipRef<SC>> {
```
kevjue, line 69: I feel like there def. should be a comment that this is used for non critical functions (e.g. debug_interactions).

Or maybe just do away with it?  Like go all in with derive?

kevjue (approved):
LGTM

PR #294, opened 2024-02-22
feat: prover memory optimizations

jtguibas:
Significantly decreases prover memory usage, among other performance improvements such as optimized trace generation.


core/src/stark/types.rs
```diff
@@ -87,6 +87,26 @@ pub enum ShardMainDataWrapper<SC: StarkGenericConfig> {
 }
 
 impl<SC: StarkGenericConfig> ShardMainDataWrapper<SC> {
+    pub fn materialize_inplace(&mut self) -> Result<&ShardMainData<SC>, Error>
```
tamirhemo, line 90: nit: maybe `materialize_in_place`

Also it's a bit of a strange name

tamirhemo (approved):

PR #310, opened 2024-02-26
refactor: introduce a union type for `opcode_specific_columns`

dlubarov:
This may be a bit questionable since union access is unsafe, leading to additional unsafe code. I would argue that that the usual safety concern with union simply doesn't apply here, since all interpretations of the underlying data are simultaneously valid. Still, any use of `unsafe` is not ideal.

The use of unions also requires `T: Copy`, but I think that should be fine.

On the plus side, `size_of_opcode_specific_columns()` can now be replaced with a simple `size_of()` call, and code that interacts with opcode-specific columns becomes a bit cleaner.

Given the pros and cons, this is a tentative suggestion; if you prefer the old style I think that's perfectly reasonable also.


jtguibas:
Thanks for the PR @dlubarov! Taking a look now, sorry we missed this.

jtguibas (approved):

Issue #231, opened 2024-02-14
Precompile: implement blake2 precompile


puma314:
We have precompiles for many popular hash functions, including keccak256, sha256, and blake3. Some blockchain infra uses the [blake2 hash function](https://en.wikipedia.org/wiki/BLAKE_(hash_function)). It would be good to create a precompile for this hash function.

Godspower-Eze:
I would like to work on this

PR #321, opened 2024-02-27
feat: add blake2b compression precompile

0xkanekiken:
The implementation is based on reference #180 and partially addresses issue #231. In the Blake2b hash function, the state and message elements are represented as `u64` values, whereas the sp1 standard supports `u32` words. As a result, each `u64` is implemented using two `u32` limbs, with appropriate handling for this format.

Furthermore, two operations for `u64` data types have been introduced:

- Fixed right shift for `u64`.
- Addition for `u64`.

The XOR operation for `u64` can be derived from the XOR operations of individual `u32` limbs.

To-Do:

- [x] Add tests.
- [x] Integrate with sp1.


0xkanekiken:
encountered the following verification error and need to figure out how to resolve it

image

kevjue:
@0xkanekiken - I see that your PR is modeled off of the Blake3 precompile.  I recently [fixed a bug](https://github.com/succinctlabs/sp1/pull/336) in that precompile which I believe also affects this PR.

0xkanekiken:
@kevjue- thank you. I had the same question when I was taking a look at the constraints of blake3. However, I noticed that the `round_index` was updated to `ROUND_COUNT`, which seemed to make the constraint valid. In fact, the tests for blake3 were also passing.

https://github.com/succinctlabs/sp1/blob/4cf63779809d0da483822b897073d3c4bb13dcfe/core/src/syscall/precompiles/blake3/compress/trace.rs#L108-L116

I attempted to incorporate the fix you introduced, but I am still encountering the `NonZeroCumulativeSum` error. Is there a more systematic approach to evaluating all the AIR and determining whether it satisfies the constraints or not?

kevjue:
@0xkanekiken  - Ya, the fix PR removed the need for setting `round_index`.  That was there to deal with the very last real row of blake3 where it expected the next (which would be a padded row) row to have a value of `round_index + 1`, which would be ROUND_COUNT.

Regarding the `NonZeroCumulativeSum` issue, that means that something is inconsistent with your interactions (a term we use for communication between chips).

You can try adding this function (https://github.com/succinctlabs/sp1/blob/049410580e3945f1f90767f740ed6c7696baefdf/core/src/lookup/debug.rs#L112) in your test case which will output any inconsistencies.  Here is an example of how to use it:  https://github.com/succinctlabs/sp1/blob/049410580e3945f1f90767f740ed6c7696baefdf/core/src/memory/global.rs#L200.

0xkanekiken:
Thank you for all the guidance and support! Indeed, the problem stemmed from how I defined the memory access AIR constraints. I've addressed this issue in https://github.com/succinctlabs/sp1/pull/321/commits/54a416ec651e9ae8e586035dce7147d5e0dfb032. Consequently, the tests are passing, and I'm able to generate the proof. It's ready for a review now.

kevjue:
Thanks for the PR @0xkanekiken !  In general, it looks great!  But we'll take a closer look at it soon.

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+        let a = (a_hi as u64) << 32 | a_lo as u64;
+        let b = (b_hi as u64) << 32 | b_lo as u64;
+        let expected = a.wrapping_add(b);
+        self.lo = Word::from(expected as u32);
+        self.hi = Word::from((expected >> 32) as u32);
+        let a = a.to_le_bytes();
+        let b = b.to_le_bytes();
+
+        let mut carry = [0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8];
+        if (a[0] as u32) + (b[0] as u32) > 255 {
```
puma314, line 44: nit: can we use a for loop here instead of doing this manually?

tests/blake2b-compress/src/main.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    unsafe {
+        syscall_blake2b_compress_inner(
+            input_state.as_mut_ptr() as *mut u32,
+            input_message.as_ptr() as *const u32,
+        );
+    }
+
+    // The expected output state is the result of compress_inner.
+    let output_state: [u8; 128] = [
```
puma314, line 36: would it be possible to import a pure rust implementation of blake2b in this program and then get the `output_state` from calling that so that we have a ground truth reference for the code?

core/src/syscall/precompiles/blake2b/compress/trace.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn generate_trace(
+        &self,
+        input: &ExecutionRecord,
+        output: &mut ExecutionRecord,
+    ) -> RowMajorMatrix<F> {
+        let mut rows = Vec::new();
+
+        let mut new_field_events = Vec::new();
+
+        for i in 0..input.blake2b_compress_inner_events.len() {
```
puma314, line 33: For maximum performance, we generally recommend that you use `par_iter` to iterate over the events here so that trace generation for each row can be done in parallel. @jtguibas @ratankaliani might have a good example for you to take a look at.

core/src/syscall/precompiles/blake2b/compress/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+#[allow(dead_code)]
+pub(crate) const MSG_NUM_WORDS: usize = MSG_SIZE * 2;
+
+/// The number of rounds in the compress inner operation.
+pub(crate) const NUM_MIX_ROUNDS: usize = 12;
+
+/// The number of time we call `mix` in the compress inner operation in each mix round.
+pub(crate) const OPERATION_COUNT: usize = 8;
+
+/// The number of `Word`s in the state that we pass to `mix`. Each `Word` is 8 bytes.
```
puma314, line 41: nit: is "Word" here correct? don't you mean u64?

puma314: maybe it's better to just call this `STATE_SIZE` to be consistent with `MSG_SIZE`

core/src/syscall/precompiles/blake2b/compress/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+pub(crate) const STATE_ELE_PER_CALL: usize = 4;
+
+/// Each state word is 8 bytes and our words size is 4 bytes. So we need to double the size of the
+/// state.
+pub(crate) const NUM_STATE_WORDS_PER_CALL: usize = STATE_ELE_PER_CALL * 2;
```
puma314, line 46: and same here with calling it `STATE_NUM_WORDS`

core/src/syscall/precompiles/blake2b/compress/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+
+/// The number of `Word`s in the message that we pass to `mix`. Each `Word` is 8 bytes.
```
puma314, line 48: nit: I think same comment as above -- is word technically correct? I think you mean u64?

core/src/syscall/precompiles/blake2b/compress/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+    [12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11],
+    [13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10],
+    [6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5],
+    [10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0],
+    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15],
+    [14, 10, 4, 8, 9, 15, 13, 6, 1, 12, 0, 2, 11, 7, 5, 3],
+];
+
+#[inline(always)]
+pub(crate) fn mix(input: [u64; MIX_INPUT_SIZE]) -> [u64; STATE_ELE_PER_CALL] {
```
puma314, line 90: so this is the function that is each row of the trace? would be nice to have a comment about that

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+use crate::air::WORD_SIZE;
+use crate::air::{SP1AirBuilder, Word};
+use crate::operations::{AddOperationU64, FixedRotateRightOperationU64, XorOperation};
+use crate::runtime::ExecutionRecord;
+
+use super::mix;
+use super::{MIX_INPUT_SIZE, STATE_ELE_PER_CALL};
+/// A set of columns needed to compute the `mix` of the input state.
+///  ``` ignore
+/// fn mix(state: &mut BlockWords, a: u64, b: u64, c: u64, d: u64, x: u64, y: u64) {
```
puma314, line 17: nit: do you mean to have `state[a]` here or just `a`? I thnk you likely just want to have `a, b, c...` not `state[a], state[b], ...`

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+///     state[c] = state[c].wrapping_add(state[d]);
+///     state[b] = (state[b] ^ state[c]).rotate_right(63);
+/// }
+///  ```
+#[derive(AlignedBorrow, Default, Debug, Clone, Copy)]
+#[repr(C)]
+pub struct MixOperation<T> {
+    pub a_plus_b: AddOperationU64<T>,
+    pub a_plus_b_plus_x: AddOperationU64<T>,
+    pub d_xor_a_lo: XorOperation<T>,
```
puma314, line 33: hm, perhaps we should have a `Xor64Operation` that is very simple and just wraps 2 `XorOperation`'s, what do you think?

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub c_plus_d_2: AddOperationU64<T>,
+    pub b_xor_c_2_lo: XorOperation<T>,
+    pub b_xor_c_2_hi: XorOperation<T>,
+    pub b_xor_c_2_rotate_right_63: FixedRotateRightOperationU64<T>,
+    /// `state[a]`, `state[b]`, `state[c]`, `state[d]` after all the steps.
+    pub result: [Word<T>; STATE_ELE_PER_CALL * 2],
+}
+
+impl<F: Field> MixOperation<F> {
+    pub fn populate(&mut self, record: &mut ExecutionRecord, input: [u32; 12]) -> [u32; 8] {
```
puma314, line 54: I think here, you should keep everything in `u64` and then only do the decomposition within the internal of the operation itself. instead of having to do `a_lo, a_hi` everywhere, just have all the populate methods also take `u64`

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+            (c_hi as u64) << 32 | c_lo as u64,
+            (d_hi as u64) << 32 | d_lo as u64,
+            (x_hi as u64) << 32 | x_lo as u64,
+            (y_hi as u64) << 32 | y_lo as u64,
+        ];
+
+        // First 4 steps.
+        {
+            // a = a + b + x.
+            (a_lo, a_hi) = self.a_plus_b.populate(record, a_lo, a_hi, b_lo, b_hi);
```
puma314, line 81: yes, here your life would be a lot easier if all `populate` took in `u64` and also returned `u64`

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+        result
+    }
+
+    pub fn eval<AB: SP1AirBuilder>(
+        builder: &mut AB,
+        input: [Word<AB::Var>; MIX_INPUT_SIZE * 2],
+        cols: MixOperation<AB::Var>,
+        is_real: AB::Var,
+    ) {
+        builder.assert_bool(is_real);
```
puma314, line 152: hm, do we typically have these in the operation? I think these columns are globally asserted as `bool` within their respective table. so i would suggest removing here.

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut d_hi = input[7];
+        let x_lo = input[8];
+        let x_hi = input[9];
+        let y_lo = input[10];
+        let y_hi = input[11];
+
+        // First 4 steps.
+        {
+            // a = a + b + x.
+            AddOperationU64::<AB::F>::eval(builder, a_lo, a_hi, b_lo, b_hi, cols.a_plus_b, is_real);
```
puma314, line 169: I think your life would be easier if there was an `U64` type similar to `Word` that basically had 8 AB::Var elements instead of 4. Then you wouldn't have to have all this hi/lo stuff everywhere and could just keep passing around the `U64`

core/src/syscall/precompiles/blake2b/compress/mix.rs
```diff
@@ -0,0 +1,281 @@
...
+                cols.b_xor_c_2_rotate_right_63,
+                is_real,
+            );
+            (b_lo, b_hi) = (
+                cols.b_xor_c_2_rotate_right_63.lo,
+                cols.b_xor_c_2_rotate_right_63.hi,
+            );
+        }
+
+        let results = [a_lo, a_hi, b_lo, b_hi, c_lo, c_hi, d_lo, d_hi];
```
puma314, line 270: so here, one thing you can do to save on these "result" columns within this mix operation, is that you can just instead have a method `.result()` on the `MixOperation` columns that basically just returns the relevant columns instead of `builder.assert_eq(cols.result[i][j], results[i][j]` because that is redundant columns, if that makes sense.

core/src/syscall/precompiles/blake2b/compress/execute.rs
```diff
@@ -0,0 +1,99 @@
...
+                            rt.mr(message_ptr + (message_index[i] as u32 * 2) * 4);
+                        let (record_hi, hi) =
+                            rt.mr(message_ptr + (message_index[i] as u32 * 2) * 4 + 4);
+                        message_reads[round][operation][2 * i] = record_lo;
+                        message_reads[round][operation][2 * i + 1] = record_hi;
+                        input.push(lo as u64 + ((hi as u64) << 32));
+                    }
+                }
+
+                // println!(
```
puma314, line 56: nit: remove?

core/src/syscall/precompiles/blake2b/compress/execute.rs
```diff
@@ -0,0 +1,99 @@
...
+                //     "round: {} operation: {} input: {:?}",
+                //     round, operation, input
+                // );
+
+                // Call mix.
+                let results = mix(input.try_into().unwrap());
+
+                // println!(
```
puma314, line 64: nit: remove

core/src/syscall/precompiles/blake2b/compress/columns.rs
```diff
@@ -0,0 +1,52 @@
...
+    mix::MixOperation, NUM_MIX_ROUNDS, NUM_MSG_WORDS_PER_CALL, NUM_STATE_WORDS_PER_CALL,
+    OPERATION_COUNT,
+};
+use super::{MSG_ELE_PER_CALL, STATE_ELE_PER_CALL};
+
+pub const NUM_BLAKE2B_COMPRESS_INNER_COLS: usize = size_of::<Blake2bCompressInnerCols<u8>>();
+
+#[derive(AlignedBorrow, Default, Debug, Clone, Copy)]
+#[repr(C)]
+pub struct Blake2bCompressInnerCols<T> {
```
puma314, line 17: how many cols does this end up being? rust analyzer should be able to tell you

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+        is_real: AB::Var,
+    ) {
+        let one = AB::Expr::one();
+        let base = AB::F::from_canonical_u32(256);
+
+        let mut builder_is_real = builder.when(is_real);
+
+        // For each limb, assert that difference between the carried result and the non-carried
+        // result is either zero or the base.
+        let overflow_0 = a_lo[0] + b_lo[0] - cols.lo[0];
```
puma314, line 104: for loop possible?

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+        builder_is_real.assert_zero((cols.carry[0] - one.clone()) * overflow_0.clone());
+        builder_is_real.assert_zero((cols.carry[1] - one.clone()) * overflow_1.clone());
+        builder_is_real.assert_zero((cols.carry[2] - one.clone()) * overflow_2.clone());
+        builder_is_real.assert_zero((cols.carry[3] - one.clone()) * overflow_3.clone());
+        builder_is_real.assert_zero((cols.carry[4] - one.clone()) * overflow_4.clone());
+        builder_is_real.assert_zero((cols.carry[5] - one.clone()) * overflow_5.clone());
+        builder_is_real.assert_zero((cols.carry[6] - one.clone()) * overflow_6.clone());
+
+        // Assert that the carry is either zero or one.
+        builder_is_real.assert_bool(cols.carry[0]);
```
puma314, line 140: for loop please

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+        builder_is_real.assert_bool(cols.carry[1]);
+        builder_is_real.assert_bool(cols.carry[2]);
+        builder_is_real.assert_bool(cols.carry[3]);
+        builder_is_real.assert_bool(cols.carry[4]);
+        builder_is_real.assert_bool(cols.carry[5]);
+        builder_is_real.assert_bool(cols.carry[6]);
+        builder_is_real.assert_bool(is_real);
```
puma314, line 147: i think we don't need to check `assert_bool` within an operation since that should be done globally at the table level

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+        builder_is_real.assert_zero(cols.carry[0] * (overflow_0.clone() - base));
+        builder_is_real.assert_zero(cols.carry[1] * (overflow_1.clone() - base));
+        builder_is_real.assert_zero(cols.carry[2] * (overflow_2.clone() - base));
+        builder_is_real.assert_zero(cols.carry[3] * (overflow_3.clone() - base));
+        builder_is_real.assert_zero(cols.carry[4] * (overflow_4.clone() - base));
+        builder_is_real.assert_zero(cols.carry[5] * (overflow_5.clone() - base));
+        builder_is_real.assert_zero(cols.carry[6] * (overflow_6.clone() - base));
+
+        // If the carry is not one, then the overflow must be zero.
+        builder_is_real.assert_zero((cols.carry[0] - one.clone()) * overflow_0.clone());
```
puma314, line 131: for loop ser?

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+        builder_is_real.assert_zero(overflow_1.clone() * (overflow_1.clone() - base));
+        builder_is_real.assert_zero(overflow_2.clone() * (overflow_2.clone() - base));
+        builder_is_real.assert_zero(overflow_3.clone() * (overflow_3.clone() - base));
+        builder_is_real.assert_zero(overflow_4.clone() * (overflow_4.clone() - base));
+        builder_is_real.assert_zero(overflow_5.clone() * (overflow_5.clone() - base));
+        builder_is_real.assert_zero(overflow_6.clone() * (overflow_6.clone() - base));
+        builder_is_real.assert_zero(overflow_7.clone() * (overflow_7.clone() - base));
+
+        // If the carry is one, then the overflow must be the base.
+        builder_is_real.assert_zero(cols.carry[0] * (overflow_0.clone() - base));
```
puma314, line 122: for loop?

core/src/operations/addu64.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Trace.
+    pub carry: [T; 7],
+}
+
+impl<F: Field> AddOperationU64<F> {
+    pub fn populate(
+        &mut self,
+        record: &mut ExecutionRecord,
+        a_lo: u32,
```
puma314, line 30: I think your life would be a lot better if this took in `u64, u64` and returned a u64? a later comment also points this out

core/src/operations/fixed_rotate_rightu64.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    pub fn carry_multiplier(rotation: usize) -> u32 {
+        let nb_bits_to_shift = Self::nb_bits_to_shift(rotation);
+        1 << (8 - nb_bits_to_shift)
+    }
+
+    pub fn populate(
+        &mut self,
+        record: &mut ExecutionRecord,
+        input_lo: u32,
```
puma314, line 49: this might be a lot better if you took in a `u64` instead of 2 `u32`

core/src/syscall/precompiles/blake2b/compress/air.rs
```diff
@@ -0,0 +1,248 @@
...
+        self.constrain_memory(builder, local);
+
+        self.constrain_mix_operation(builder, local);
+    }
+}
+
+impl Blake2bCompressInnerChip {
+    /// Constrains the given index is correct for the given selector. The `selector` is an
+    /// `n`-dimensional boolean array whose `i`-th element is true if and only if the index is `i`.
+    fn constrain_index_selector<AB: SP1AirBuilder>(
```
puma314, line 44: do we not already have a generic thing of doing this? this seems like a really good utility to have in airbuilder in general tbh, so maybe we should move it to a move general crate with all the existing air builder methods

core/src/syscall/precompiles/blake2b/compress/trace.rs
0xkanekiken: done! please see https://github.com/succinctlabs/sp1/commit/dae55542cd38371935a43e3c8421badded586abb

core/src/syscall/precompiles/blake2b/compress/mod.rs
0xkanekiken: I've updated the comments in the commit https://github.com/succinctlabs/sp1/commit/dae55542cd38371935a43e3c8421badded586abb

core/src/syscall/precompiles/blake2b/compress/mod.rs
0xkanekiken: done

core/src/syscall/precompiles/blake2b/compress/mix.rs
0xkanekiken: yeah, it makes sense! I've introduced it in latest commit https://github.com/succinctlabs/sp1/commit/dae55542cd38371935a43e3c8421badded586abb

core/src/syscall/precompiles/blake2b/compress/mix.rs
0xkanekiken: done, thanks

core/src/syscall/precompiles/blake2b/compress/mix.rs
0xkanekiken: yes- it was there in `g`, `add` and `add4` which I used as a reference. I've removed it from the `add_u64` and `mix` in the latest commit. We can get it removed from other parts in a separate PR?

core/src/syscall/precompiles/blake2b/compress/mix.rs
0xkanekiken: I've created a new struct `WordU64` for this.

core/src/syscall/precompiles/blake2b/compress/columns.rs
0xkanekiken: It has `351` columns.

core/src/syscall/precompiles/blake2b/compress/air.rs
0xkanekiken: should I address this issue in the current PR, or should it be handled in a separate PR by creating an issue?

rsoeldner:
@puma314 do you mind reviving this PR? I would be very interested in the precompile :-)

XuJiandong:
Hi, we've built an application that involves a large amount of blake2b hashing. Without precompile support, it's nearly impossible to achieve. Could you please revive this PR?

PR #323, opened 2024-02-27
perf: enable reconstruct trace/commit by default instead of using disk

jtguibas:


core/src/stark/types.rs
```diff
@@ -64,7 +64,7 @@ impl<SC: StarkGenericConfig> ShardMainData<SC> {
     where
         ShardMainData<SC>: Serialize,
     {
-        let mut writer = BufWriter::new(&file);
+        let mut writer = BufWriter::with_capacity(1024 * 1024 * 512, &file);
```
tamirhemo, line 67: Should this capacity be some env variable as well?

core/src/utils/prove.rs
```diff
@@ -476,6 +476,12 @@ pub(super) mod baby_bear_blake3 {
         }
     }
 
+    impl Clone for BabyBearBlake3 {
```
tamirhemo, line 479: where do we need to clone the config?

jtguibas: Right now it's required in prove_with_config, but maybe it's not necessary?

tamirhemo: I just removed the bound and it seems fine.

PR #329, opened 2024-02-28
fix: S-box in Poseidon2 config

dlubarov:
x^5 works with Mersenne31, but x^7 is the smallest permutation monomial over BabyBear. I think this mistake originates in some Plonky3 example code which had the same issue.


jtguibas:
Thanks for catching this!

tamirhemo (approved):

PR #341, opened 2024-02-29
refactor: logging

ratankaliani:
Update logging levels to be more informative.

Dependent on https://github.com/succinctlabs/plonky3/pull/29


core/src/syscall/write.rs
```diff
@@ -56,7 +56,7 @@ impl Syscall for SyscallWrite {
                         u32_to_comma_separated(rt.state.global_clk - start)
                     );
                 } else {
-                    log::info!("stdout: {}", s.trim_end());
+                    log::debug!("stdout: {}", s.trim_end());
```
puma314, line 59: I think this should be changed back to `log::info!()`?

ratankaliani: fixed

PR #343, opened 2024-03-01
refactor(breaking): move `SP1Prover` and `SP1Verifier` to `sp1` crate

ratankaliani:
In preparation for remote proving, move `SP1Prover` and `SP1Verifier` into `sp1_core`. User entrypoints into core should go through the `sp1` crate, and the `sp1-core` crate should not have dependencies on libraries such as `reqwest`.

Note: This change should be accompanied with changes to the documentation, which previously specified importing user-facing dependencies from `sp1_core`.


sp1/src/io.rs
```diff
@@ -125,10 +125,10 @@ pub mod proof_serde {
 
     #[cfg(test)]
     mod tests {
-        use crate::{
-            utils::{tests::FIBONACCI_IO_ELF, BabyBearBlake3},
-            SP1ProofWithIO, SP1Prover, SP1Stdin, SP1Verifier,
-        };
+        use crate::{utils::BabyBearBlake3, SP1ProofWithIO, SP1Prover, SP1Stdin, SP1Verifier};
+
+        pub const FIBONACCI_IO_ELF: &[u8] =
```
ratankaliani, line 130: Note: Didn't want to re-export all of the ELF's from `programs::tests` as it would slow down compilation time (would force `sp1-core` to compile all of the ELF's before `sp1` could import). Instead, just include the only ELF required by the tests in `io.rs`

PR #357, opened 2024-03-08
feat: recursive DSL initial commit

tamirhemo:


recursion/compiler/examples/fibonacci.rs
```diff
@@ -0,0 +1,34 @@
+use p3_baby_bear::BabyBear;
+use p3_field::AbstractField;
+use sp1_core::stark::StarkGenericConfig;
+use sp1_core::utils::BabyBearPoseidon2;
+use sp1_recursion_compiler::prelude::*;
+use sp1_recursion_core::runtime::Runtime;
+
+fn main() {
+    let mut builder = AsmBuilder::<BabyBear>::new();
+    let a: Felt<_> = builder.constant(BabyBear::zero());
```
jtguibas, line 10: It's weird to me that Felt has a generic in F, but Bool doesn't.

recursion/compiler/examples/logic.rs
```diff
@@ -0,0 +1,24 @@
+use p3_baby_bear::BabyBear;
+use sp1_recursion_compiler::prelude::*;
+
+fn main() {
+    let mut builder = AsmBuilder::<BabyBear>::new();
+
+    let a: Bool = builder.constant(false);
+    let b: Bool = builder.constant(true);
+
+    let and: Bool = builder.uninit();
```
jtguibas, line 10: uninit is a weird name, maybe we can have something similar to plonky2x where it's just builder.constant and builder.init?

recursion/compiler/src/asm/instruction.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Multiply, dst = lhs * rhs.
+    MUL(i32, i32, i32),
+    /// Multiply immediate.
+    MULI(i32, i32, F),
+    /// Divide, dst = lhs / rhs.
+    DIV(i32, i32, i32),
+    /// Divide immediate, dst = lhs / rhs.
+    DIVI(i32, i32, F),
+    /// Divide value from immediate, dst = rhs / lhs.
+    DIVIN(i32, i32, F),
```
jtguibas, line 40: It's a bit weird here that b/c get reordered from the VM-side of things, maybe better to just do DIVIN(i32, F, i32)

jtguibas (approved):

recursion/compiler/examples/fibonacci.rs
tamirhemo: That's because we need to know how to make arithmetic operations on Felt<F>

recursion/compiler/src/util.rs
```diff
@@ -0,0 +1,9 @@
+use p3_field::PrimeField32;
+
+pub fn canonical_i32_to_field<F: PrimeField32>(x: i32) -> u32 {
+    if x < 0 {
```
dlubarov, line 4: It seems like this assumes `-P < x < P` which I think seems fine based on how it's currently used, but maybe good to add an assert in case it gets used for other things later

tamirhemo: added assert

PR #358, opened 2024-03-09
feat: runtime optimizations

ctian1:


core/src/utils/prove.rs
```diff
@@ -26,30 +29,40 @@ pub trait StarkUtils: StarkGenericConfig {
...
+    let shards = tracing::info_span!("runtime.run(...)").in_scope(|| {
+        let mut runtime = Runtime::new(program, &mut receiver);
         runtime.run();
-        runtime
+        receiver.close()
     });
-    let config = BabyBearBlake3::new();
-    prove_core(config, runtime)
+    let record = &shards[0];
+    println!("stats: {:?}", record.stats());
```
jtguibas, line 48: nit: fix

core/src/runtime/event.rs
```diff
@@ -0,0 +1,654 @@
...
+
+/// An event handler that take in events from program execution.
+pub trait EventHandler {
+    fn handle(&mut self, event: RuntimeEvent);
+}
+
+/// An event handler that sends events to a buffer on another thread, periodically runs
+/// dependency generation, and finally puts them into ExecutionRecord shards.
+pub struct AsyncEventRecorder<SC: StarkGenericConfig> {
+    s: Option<mpsc::Sender<RuntimeEvent>>,
```
jtguibas, line 49: why does s need to be option?

jtguibas:
let's revisit this later, i think it's too complicated rn

PR #359, opened 2024-03-09
feat: core recursion air constraints

jtguibas:
- basic clock constraints
- memory + immediates
- field alu 
- field alu <-> cpu
- extension alu
- eq/neq
- fp branching


core/src/air/extension.rs
```diff
@@ -0,0 +1,65 @@
+use std::ops::{Add, Mul, Neg, Sub};
+
+use p3_field::AbstractField;
+use sp1_derive::AlignedBorrow;
+
+const DEGREE: usize = 4;
+
+#[derive(AlignedBorrow, Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+#[repr(C)]
+pub struct Extension<T>(pub [T; DEGREE]);
```
tamirhemo, line 10: I don't think we have to use a struct for the extension, we could instead use the build in trait in Plonky3 and just case from an array.

recursion/core/src/memory/air.rs
```diff
@@ -0,0 +1,155 @@
...
+
+use super::columns::MemoryInitCols;
+use crate::air::Word;
+use crate::memory::MemoryChipKind;
+use crate::memory::MemoryGlobalChip;
+use crate::runtime::ExecutionRecord;
+
+pub(crate) const NUM_MEMORY_INIT_COLS: usize = size_of::<MemoryInitCols<u8>>();
+
+// pub(crate) const MEMORY_INIT_COL_MAP: MemoryInitCols<usize> = make_col_map();
```
tamirhemo, line 20: nit: remove

recursion/core/src/memory/air.rs
```diff
@@ -0,0 +1,155 @@
...
+
+// const fn make_col_map() -> MemoryInitCols<usize> {
```
tamirhemo, line 22: nit: remove

recursion/core/src/memory/air.rs
```diff
@@ -0,0 +1,155 @@
...
+//     let indices_arr = indices_arr::<NUM_MEMORY_INIT_COLS>();
+//     unsafe { transmute::<[usize; NUM_MEMORY_INIT_COLS], MemoryInitCols<usize>>(indices_arr) }
+// }
+
+#[allow(dead_code)]
+impl MemoryGlobalChip {
```
tamirhemo, line 28: It would be nice if we could have only a single memory chip generic in the word but maybe that is not necessary and even bad if RISV-AIR and Recursion Air are completely separate crates

tamirhemo (approved):

PR #369, opened 2024-03-11
fix: ci test fix

0xkanekiken:


jtguibas:
It might be better to specify sp1-recursion-core more explicitly. We have a lot of random crates in the repo.

0xkanekiken:
I'm not sure if cargo supports executing tests from multiple packages with a single command. I've set up a separate action for sp1-recursion-core. However, I believe it should be possible to combine the following into a single action, though I'll have to check how to accomplish this:

```
cargo test -p sp1-core --release && cargo test -p sp1-recursion-core --release
```

@jtguibas - Would you prefer the second approach over creating a separate action?

PR #374, opened 2024-03-13
fix: RUST_LOG log level in cycle-tracking section of book

dynm:
The book currently instructs users to set RUST_LOG to 'info' for the cycle-tracker. However, in #363, the log level for the cycle-tracker was changed to 'debug'. 

This Pull Request updates the RUST_LOG level in the book from 'info' to 'debug', matching the actual setting in the code. This ensures that users following the book's instructions will see the expected logs.


jtguibas (approved):

jtguibas:
Hi, thanks we did indeed change this.

PR #389, opened 2024-03-15
feat: checkpoint runtime for constant memory usage

ctian1:
Proving large programs (>500m cycles?) is currently impossible on machines with limited RAM (<= 32 GB) because all runtime events are stored in memory at once. This PR changes the runtime/prover to have a configurable constant max memory usage.

* Adds the env var SHARD_BATCH_SIZE which causes the runtime to save a checkpoint every N shards.
  * Setting it to 0 (default) results in a single batch, so proving everything at once like before
* Runtime now has `Runtime::recover(program, state)`, `execute_record` and `execute_state` for events / checkpointing respectively
* `utils::run_and_prove`: During proving, each checkpoint will be loaded and proved, 1 at a time


core/src/utils/prove.rs
```diff
@@ -89,9 +86,123 @@ pub fn run_test(program: Program) -> Result<(), crate::stark::ProgramVerificatio
     Ok(())
 }
 
-pub fn prove_elf(elf: &[u8]) -> crate::stark::Proof<BabyBearBlake3> {
-    let program = Program::from(elf);
-    prove(program)
+fn trace_checkpoint(program: Program, file: &File) -> ExecutionRecord {
```
jtguibas, line 89: It feels like this method should actually be impl'd on Runtime. It's also weird that the "file-handling" is mixed with the actual logic.

What do you think of:

let runtime = Runtime::recover(program, state);
runtime.execute_record();

core/src/utils/prove.rs
```diff
@@ -89,9 +86,123 @@ pub fn run_test(program: Program) -> Result<(), crate::stark::ProgramVerificatio
...
+    });
+    events
+}
+
+fn reset_seek(file: &mut File) {
+    file.seek(std::io::SeekFrom::Start(0))
+        .expect("failed to seek to start of tempfile");
+}
+
+pub fn run_and_prove<SC: StarkGenericConfig + StarkUtils + Send + Sync + Serialize>(
```
jtguibas, line 111: Is the behavior actually the same if the env var is set to 0? Won't it execute the runtime twice?

jtguibas (approved):

core/src/utils/prove.rs
ctian1: yes it will, first time without events though. maybe we should add a special case?

core/src/utils/prove.rs
ctian1: yea that seems good

core/src/utils/prove.rs
jtguibas: Yeah I think it's probably worth doing that if you can do it in a clean way

Issue #230, opened 2024-02-14
Precompile: Add secp256r1 precompiles


puma314:
SP1 already has support for secp256k1 precompiles used in ECDSA signature verification. It would be good to add similar precompiles for secp256r1. This would involve

- implementing the secp256r1 parameters, similar to secp256k1 parameters
- adding a new chip with the secp256r1 parameters
- creating extern syscalls to utilize these new precompiles
- writing tests for these syscalls to verify they work correctly

The precompile system of SP1 is not well documented at present, so this would require the contributor to be able to figure out how things work with minimal documentation.

shreyas-londhe:
I'm interested in working on this!

nfurfaro:
I'm interested in tackling this!

PR #400, opened 2024-03-18
feat: add secp256r1 add/double precompiles

shuklaayush:
Adds `SECP256R1_ADD` and `SECP256R1_DOUBLE` precompiles for [secp256r1](https://neuromancer.sk/std/secg/secp256r1) (P-256) operations. The implementation is similar to secp256k1 and I reused the `WeierstrassAdd` and `WeierstrassDouble` chips

Run the tests:

```bash
cargo t -r test_secp256r1_add_simple && cargo t -r test_secp256r1_double_simple && cargo t -r test_secp256r1_mul_simple
```

Since the constant $a \neq 0$ for this curve, the constraints in the `WeierstrassDouble` chip were failing on padding rows.  I modified it to only check the constraints on rows where the `is_real` column is set

Addresses #230


shuklaayush:
@puma314 For full signature verification, I'll have to add a precompile for secp256r1 point decompression (akin to `SECP256K1_DECOMPRESS`). The `SECP256K1_DECOMPRESS` precompile assumes that the curve constant $a = 0$ which is not the case for secp256r1.
Should I generalize this precompile for all SW curves (by adding more columns to the air) or should I create a new precompile specifically for secp256r1? I'm thinking of following up with that in a separate PR

shuklaayush:
@puma314 Done! Should be good for review

PR #404, opened 2024-03-19
chore:  Move memory access timestamp check inline

kevjue:


core/src/air/builder.rs
```diff
@@ -308,79 +281,103 @@ pub trait MemoryAirBuilder: BaseAirBuilder {
...
+    ) where
+        Eb: Into<Self::Expr> + Clone,
+        Everify: Into<Self::Expr>,
+    {
+        let current_ts: Self::Expr = mem_access.current_ts.clone().into();
+        let prev_ts: Self::Expr = mem_access.prev_ts.clone().into();
+        let do_check: Self::Expr = do_check.into();
+
+        // Verify that ts_diff is calculated correctly correctly.
+        let max_shard_size = env::shard_size();
```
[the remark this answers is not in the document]
kevjue: max_shard_size is retrieved from the user setting via the env var.  I'll rename `max_shard_size` to something more apt.

core/src/memory/columns.rs
```diff
@@ -35,16 +37,63 @@ pub struct MemoryAccessCols<T> {
...
+        builder
+            .when(do_check.clone())
+            .assert_bool(compare_clk.clone());
+        builder
+            .when(do_check.clone())
+            .when(compare_clk.clone())
+            .assert_eq(current_shard.clone(), prev_shard.clone());
+
+        // Verify self.prev_time_value's value.
+        let expected_prev_time_value = compare_clk.clone() * prev_clk.clone()
```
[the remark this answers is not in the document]
kevjue: Actually, looks like we can remove some of these materialized columns.  I recall now that the main reason was b/c they were being sent to the field table, and sending an interaction takes up 2 degrees.

core/src/memory/columns.rs
```diff
@@ -34,17 +34,18 @@ pub struct MemoryAccessCols<T> {
...
-
-    // This materialized column is equal to use_clk_comparison ? prev_timestamp : current_shard
-    pub prev_time_value: T,
     // This will be true if the current shard == prev_access's shard, else false.
-    pub use_clk_comparison: T,
-    // This materialized column is equal to use_clk_comparison ? current_clk : current_shard
-    pub current_time_value: T,
+    pub compare_clk: T,
+
+    // The following columns are decomposed limbs for the different between the current access's timestamp
```
jtguibas, line 40: nit: should be "///"

core/src/memory/trace.rs
```diff
@@ -106,22 +119,28 @@ impl<F: Field> MemoryAccessCols<F> {
...
-        // Add a field op event for the prev_time_value < current_time_value constraint.
-        let field_event = FieldEvent::new(true, prev_time_value, current_time_value);
-        new_field_events.push(field_event);
+        let diff_minus_one = current_time_value - prev_time_value - 1;
+        let diff_16bit_limb = diff_minus_one & 0xffff;
+        self.diff_16bit_limb = F::from_canonical_u32(diff_16bit_limb);
+        let diff_8bit_limb = (diff_minus_one >> 16) & 0xff;
+        self.diff_8bit_limb = F::from_canonical_u32(diff_8bit_limb);
+
+        // Add a byte table lookup with the 16Range op
```
jtguibas, line 140: nit: should have period

core/src/air/builder.rs
```diff
@@ -368,79 +350,121 @@ pub trait MemoryAirBuilder: BaseAirBuilder {
...
+        // Get the comparison timestamp values for the current and previous memory access.
+        let prev_comp_value = self.if_else(
+            mem_access.compare_clk.clone(),
+            mem_access.prev_clk.clone(),
+            mem_access.prev_shard.clone(),
+        );
+
+        let current_comp_val = self.if_else(compare_clk.clone(), clk.into(), shard);
+
+        // Assert `current_comp_val > prev_comp_val`. We check this by asserting that
```
jtguibas, line 433: nit: long comments should be written in paragraph form, similar to function declarations.

```
// Assert `current_comp_val > prev_comp_val`. We check this by asserting that
// `0 <= current_comp_val-prev_comp_val-1 < 2^24`.
//
// The equivalence of these statements comes from the fact that if `current_comp_val <= prev_comp_val`,
// then `current_comp_val-prev_comp_val-1 < 0` and will underflow in the prime field,
// resulting in a value that is `>= 2^24` as long as both `current_comp_val, prev_comp_val` are
// range-checked to be `<2^24` and as long as we're working in a field larger than `2 * 2^24`
// (which is true of the BabyBear and Mersenne31 prime).

PR #410, opened 2024-03-20
chore:  Merge sub chip into add chip

kevjue:


tamirhemo (approved):
LGTM!

I am wondering if there is a way to avoid the extra interaction but it seems not possible without adding more word columns which would make it not worth it.

kevjue:
> LGTM!
> 
> I am wondering if there is a way to avoid the extra interaction but it seems not possible without adding more word columns which would make it not worth it.

Ya, I was trying to figure out a way too.  As you said, it would require additional words b/c of the degree requirement for interactions.

PR #422, opened 2024-03-21
feat: dsl derive macro + fri pow witness verify

jtguibas:


recursion/derive/src/lib.rs
```diff
@@ -0,0 +1,144 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DeriveInput, Fields};
+
+#[proc_macro_derive(DslVariable)]
+pub fn derive_variable(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = input.ident; // Struct name
```
tamirhemo, line 10: Seems like there no generics support so maybe we can panic if the list of generics in the type is not empty, and we can fix it later if needed.

tamirhemo (approved):

recursion/derive/src/lib.rs
jtguibas: Sounds good

PR #427, opened 2024-03-22
Add example attested-image, example on (de)serializing vector 

ultraspicy:
A simplified example on how to use sp1 for (de)serializing vector input.


ultraspicy:
I asked the question in telegram group about the best practice to take vectors as input. Not sure this is what you expect, but at least it works.

ultraspicy:
fixed the diff title

PR #455, opened 2024-03-29
feat: public values

kevjue:


core/src/cpu/air/mod.rs
```diff
@@ -272,7 +274,6 @@ impl CpuChip {
         let syscall_id = syscall_code[0];
         let send_to_table = syscall_code[1]; // Does the syscall have a table that should be sent.
         let num_cycles = syscall_code[2]; // How many extra cycles to increment the clk for the syscall.
-        let is_halt = syscall_code[3]; // Whether or not the syscall is a halt.
```
jtguibas, line 275: nit: if you're going to remove this, update the comment above SyscallCode

core/src/runtime/mod.rs
```diff
@@ -84,6 +86,10 @@ pub struct Runtime {
     pub max_syscall_cycles: u32,
 
     pub emit_events: bool,
+
+    pub pi_hasher: Option<Sha256>,
```
jtguibas, line 90: nit: seems unecessary, can you just instantiate a Sh256 instance locally where you need it?

zkvm/entrypoint/src/syscalls/io.rs
```diff
@@ -59,17 +64,25 @@ pub extern "C" fn syscall_read(fd: u32, read_buf: *mut u8, nbytes: usize) {
...
+                    in("t0") crate::syscalls::WRITE,
+                    in("a0") fd,
+                    in("a1") write_buf,
+                    in("a2") nbytes,
+                );
+            }
 
-    #[cfg(not(target_os = "zkvm"))]
-    unreachable!()
+            // Public Inputs writes
```
jtguibas, line 79: nit: comment format

zkvm/entrypoint/src/lib.rs
```diff
@@ -32,22 +37,81 @@ macro_rules! entrypoint {
 #[cfg(all(target_os = "zkvm", feature = "libm"))]
 mod libm;
 
+const PI_DIGEST_NUM_WORDS: usize = 8;
+
+#[derive(Serialize, Deserialize)]
+pub struct PiDigest<T>(pub [T; PI_DIGEST_NUM_WORDS]);
```
jtguibas, line 43: seems unnecessary to make this generic in T

kevjue: The digest will be represented either as `u32s` (in the guest program and the runtime) or Word<F> (in the air and prover/verifier).  

So the generic could be useful?

jtguibas: Ah I see, it's a bit weird to import `sp1-zkvm-entrypoint` into Runtime though.

kevjue: Ya, agreed.  I'll see if there is a shared library between `core` and `zkvm` (or maybe it makes sense for `zkvm` to import `core`).

core/src/runtime/mod.rs
kevjue: I don't keep a buffer for the public input bytes.  Instead, it's streamed into this hasher.  lmk if you think the raw bytes should be buffered instead.

core/src/runtime/syscall.rs
```diff
@@ -23,12 +24,11 @@ use strum_macros::EnumIter;
 /// - The second byte is 0/1 depending on whether the syscall has a separate table. This is used
 /// in the CPU table to determine whether to lookup the syscall using the syscall interaction.
 /// - The third byte is the number of additional cycles the syscall uses.
-/// - The fourth byte is 0/1 depending on whether the syscall is the HALT syscall.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, EnumIter)]
 #[allow(non_camel_case_types)]
 pub enum SyscallCode {
     /// Halts the program.
-    HALT = 0x01_00_00_00,
+    HALT = 0x00_00_00_00,
```
jtguibas, line 31: Make sure to also update the entries in zkvm/syscalls or whatever

core/src/air/builder.rs
```diff
@@ -66,6 +67,25 @@ pub trait BaseAirBuilder: AirBuilder + MessageBuilder<AirInteraction<Self::Expr>
...
+        array: &[I],
+        index_bitmap: &[EIndex],
+    ) -> Self::Expr
+    where
+        I: Into<Self::Expr> + Clone,
+        EIndex: Into<Self::Expr> + Clone,
+    {
+        let mut result = Self::Expr::zero();
+
+        for (value, i) in array.iter().zip(index_bitmap) {
```
jtguibas, line 83: zip_eq might be better than zip here for safety

core/src/runtime/mod.rs
jtguibas: I do feel like it would be better if raw bytes were buffered instead, that way we can inspect the bytes more easily and avoid doing this take/unwrap thing

core/src/cpu/columns/ecall.rs
```diff
@@ -16,4 +17,9 @@ pub struct EcallCols<T> {
 
     /// Whether the current ecall is HALT.
     pub is_halt: IsZeroOperation<T>,
```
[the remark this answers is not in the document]
kevjue: I actually removed that encoding, and now the `is_halt` field is used to determine if the ecall is a `halt`.

The reason is to make it more consistent with how we identify other ecalls.  E.g. the `ENTER_UNCONSTRAINED` and the new `COMMIT` ecalls.

core/src/runtime/mod.rs
```diff
@@ -990,6 +996,14 @@ impl Runtime {
         program_memory_events.sort_by_key(|event| event.addr);
 
         self.record.program_memory_events = program_memory_events;
+
+        // Set the public input digest.
+        let mut pi_hasher = Sha256::new();
```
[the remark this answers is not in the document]
kevjue: This is used to calculate the `pi_digest` to attach it to the generated proof.

core/src/runtime/record.rs
```diff
@@ -434,15 +438,29 @@ impl MachineRecord for ExecutionRecord {
             .program_memory_events
             .extend_from_slice(&self.program_memory_events);
 
+        // Set the pi_digest for all shards.
```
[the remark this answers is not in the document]
kevjue: ya, as I mentioned in slack, we actually need to add the `pi_digest` to the last 2 shards b/c the access of that data may be needed for multiple shards in boundary case.

I'll remove the comment about the last shard.

examples/tendermint/program/src/main.rs
```diff
@@ -35,6 +35,16 @@ fn main() {
         light_block_2.validators.validators().len()
     );
 
+    println!("cycle-tracker-start: header hash");
+    let header_hash_1 = light_block_1.signed_header.header.hash();
+    let header_hash_2 = light_block_2.signed_header.header.hash();
+    println!("cycle-tracker-end: header hash");
+
+    println!("cycle-tracker-start: public input headers");
+    sp1_zkvm::io::public_input(header_hash_1.as_bytes());
```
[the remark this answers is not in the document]
jtguibas: Yeah agreed. Maybe this is easier to discuss in Slack, but what if we just stick to `read/write` and then there's public/private versions? Seems the most intuitive to users.

core/src/air/public_input.rs
```diff
@@ -0,0 +1,89 @@
...
+impl<T> IntoIterator for PiDigest<T> {
+    type Item = T;
+    type IntoIter = IntoIter<T, PI_DIGEST_NUM_WORDS>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+/// Convertion into a byte vec
```
jtguibas, line 64: nit: period

core/src/air/public_input.rs
```diff
@@ -0,0 +1,89 @@
...
+impl From<PiDigest<u32>> for Vec<u8> {
+    fn from(val: PiDigest<u32>) -> Self {
+        val.0
+            .iter()
+            .flat_map(|word| word.to_le_bytes().to_vec())
+            .collect::<Vec<u8>>()
+    }
+}
+
+/// Convertion into a field vec
```
jtguibas, line 74: nit: period

core/src/air/public_input.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn from(val: PiDigest<Word<T>>) -> Self {
+        val.0.iter().flat_map(|word| word.0).collect::<Vec<T>>()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    /// Check that the PI_DIGEST_NUM_WORDS number match the zkVM crate's.
```
jtguibas, line 85: nit: comment is below derive but above it's above

core/src/runtime/record.rs
```diff
@@ -434,15 +438,29 @@ impl MachineRecord for ExecutionRecord {
...
+        for shard in shards.iter_mut() {
+            shard.set_pi_digest(self.pi_digest.unwrap());
+        }
+
         shards
     }
+
+    fn pi_digest(&self) -> Option<PiDigest<u32>> {
```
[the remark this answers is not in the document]
kevjue: `pi_digest` and `set_pi_digest` was added to the `MachineRecord` trait so that both the `core` and `recursion`'s `ExecutionRecord` will implement it.

This actually was a suggestions from @jtguibas

core/src/runtime/record.rs
```diff
@@ -434,15 +438,29 @@ impl MachineRecord for ExecutionRecord {
...
+    }
+
+    fn set_pi_digest(&mut self, digest: PiDigest<u32>) {
+        self.pi_digest = Some(digest);
+    }
 }
 
 impl ExecutionRecord {
-    pub fn new(index: u32, program: Arc<Program>) -> Self {
+    pub fn new(index: u32, program: Arc<Program>, pi_digest: Option<PiDigest<u32>>) -> Self {
```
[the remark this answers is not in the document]
kevjue: Ya, good point.  Will change.

core/src/stark/verifier.rs
```diff
@@ -228,7 +228,7 @@ impl<SC: StarkGenericConfig, A: MachineAir<Val<SC>>> Verifier<SC, A> {
         selectors: &LagrangeSelectors<SC::Challenge>,
         alpha: SC::Challenge,
         permutation_challenges: &[SC::Challenge],
-        pi_digest: PiDigest<Word<Val<SC>>>,
+        pv_digest: Option<PublicValuesDigest<Word<Val<SC>>>>,
```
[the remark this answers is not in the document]
kevjue: It's an option for compatibility with the `recursion` crate.  Right now, it's still a TODO for that to implement public values. 

The `Option` should be deprecated soon.

I'll add a comment.

core/src/stark/verifier.rs
```diff
@@ -250,8 +250,14 @@ impl<SC: StarkGenericConfig, A: MachineAir<Val<SC>>> Verifier<SC, A> {
             local: unflatten(&opening.permutation.local),
             next: unflatten(&opening.permutation.next),
         };
-
-        let public_values: Vec<Val<SC>> = pi_digest.into();
+        if pv_digest.is_none() {
```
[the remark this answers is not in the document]
kevjue: oh weird. I'm not sure how that got into there.  ya, this is wrong.

core/src/stark/verifier.rs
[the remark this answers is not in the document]
kevjue: Actually, found a way to not make it optional, and removed the `Option`.

zkvm/entrypoint/src/syscalls/sys.rs
```diff
@@ -1,10 +1,13 @@
...
+    syscalls::{syscall_halt, syscall_write},
+    PI_DIGEST_NUM_WORDS,
+};
 
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
 pub unsafe extern "C" fn sys_panic(msg_ptr: *const u8, len: usize) -> ! {
     sys_write(2, msg_ptr, len);
-    syscall_halt(1);
+    syscall_halt(1, &[0u32; PI_DIGEST_NUM_WORDS]);
```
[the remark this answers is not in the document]
kevjue: I ended up modifying it so that it would calculate the digest even for a non zero exit.

core/src/runtime/mod.rs
```diff
@@ -997,13 +994,13 @@ impl Runtime {
 
         self.record.program_memory_events = program_memory_events;
 
-        // Set the public input digest.
-        let mut pi_hasher = Sha256::new();
-        pi_hasher.update(&self.pi_buffer);
-        let pi_digest_bytes: &[u8] = &pi_hasher.finalize();
+        // Calculate and set the public values digest.
```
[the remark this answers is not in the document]
kevjue: Ya, that works.

core/src/air/public_values.rs
```diff
@@ -0,0 +1,103 @@
...
+// TODO:  Create a config struct that will store the num_words setting and the hash function
+//        and initial entropy used.
+const PV_DIGEST_NUM_WORDS: usize = 8;
+
+/// The PublicValuesDigest struct is used to represent the public input digest.  This is the hash of all the
+/// bytes that the guest program has written to public input.
+#[derive(Serialize, Deserialize, Clone, Copy, Default, Debug)]
+pub struct PublicValuesDigest<T>([T; PV_DIGEST_NUM_WORDS]);
+
+/// Convertion from a byte array into a PiDigest<u32>.
```
ctian1, line 22: nit: Convertion -> Conversion + this file still mentions PiDigest / public inputs, should be changed to PublicValuesDigest

core/src/cpu/columns/ecall.rs
```diff
@@ -16,4 +17,9 @@ pub struct EcallCols<T> {
...
+
+    /// Whether the current ecall is a COMMIT.
+    pub is_commit: IsZeroOperation<T>,
+    pub digest_word: Word<T>,
+    pub index_bitmap: [T; PV_DIGEST_NUM_WORDS],
```
ctian1, line 24: Please comment what these columns are for and maybe use more descriptive names

core/src/cpu/air/mod.rs
```diff
@@ -420,6 +434,50 @@ impl CpuChip {
...
+
+        // Verify the index bitmap.
+        let mut bitmap_sum = AB::Expr::zero();
+        for bit in ecall_columns.index_bitmap.iter() {
+            builder.when(local.selectors.is_ecall).assert_bool(*bit);
+            bitmap_sum += (*bit).into();
+        }
+        builder.when(is_commit.clone()).assert_one(bitmap_sum);
+
+        // Verify the public_values_digest_word.  Note that for the interaction builder, it will not have
```
kevjue, line 466: Actually, this comment can be more clear.  Will push a better one soon.

zkvm/entrypoint/src/lib.rs
```diff
@@ -29,13 +29,21 @@ macro_rules! entrypoint {
     };
 }
 
+pub const PV_DIGEST_NUM_WORDS: usize = 8;
```
ctian1, line 32: please comment what this value means

zkvm/entrypoint/src/lib.rs
```diff
@@ -29,13 +29,21 @@ macro_rules! entrypoint {
...
 #[cfg(target_os = "zkvm")]
 mod zkvm {
     use crate::syscalls::syscall_halt;
+
     use getrandom::{register_custom_getrandom, Error};
+    use once_cell::sync::Lazy;
+    use sha2::{Digest, Sha256};
+    use std::sync::Mutex;
+
+    pub static PUBLIC_VALUES_HASHER: Lazy<Mutex<Sha256>> = Lazy::new(|| Mutex::new(Sha256::new()));
```
ctian1, line 46: since the zkvm is singlethreaded I don't think a mutex is necessary, we can assume we will access it safely. Maybe just Lazy<Sha256> + `static mut` will work, which is like what is done in sys_alloc_aligned

zkvm/entrypoint/Cargo.toml
```diff
@@ -9,9 +9,11 @@ bincode = "1.3.3"
 cfg-if = "1.0.0"
 getrandom = { version = "0.2.12", features = ["custom"] }
 k256 = { version = "0.13.3", features = ["ecdsa", "std", "bits"] }
+once_cell = "1.8.0"
 rand = "0.8.5"
 serde = { version = "1.0.196", features = ["derive"] }
 libm = { version = "0.2.8", optional = true }
+sha2 = "0.10.8"
```
ctian1, line 16: we should use our patched sha2 crate

core/src/cpu/air/mod.rs
```diff
@@ -420,6 +434,53 @@ impl CpuChip {
...
+                bytes.iter().map(|byte| (*byte).into()).collect::<Vec<_>>();
+
+            digest_words.push(Word::<AB::Expr>(bytes_expr_vec.try_into().unwrap()));
+        }
+
+        // Get the ecall specific columns.
+        let ecall_columns = local.opcode_specific_columns.ecall();
+
+        // Verify the index bitmap.
+        let mut bitmap_sum = AB::Expr::zero();
```
jtguibas, line 459: nit: it's kind of weird here that you do pv stuff -> ecall/index bitmap stuff -> pv stuff. it should be 1) pv stuff and then 2) ecall/index bitmap stuff or vice-versa.

core/src/cpu/air/mod.rs
```diff
@@ -420,6 +434,53 @@ impl CpuChip {
...
+    pub(crate) fn commit_eval<AB: SP1AirBuilder>(
+        &self,
+        builder: &mut AB,
+        local: &CpuCols<AB::Var>,
+        is_commit: AB::Expr,
+    ) {
+        let public_values = builder.public_values();
+
+        // Convert public values into words.
+        let mut digest_words = Vec::new();
```
jtguibas, line 447: It seems like this type of logic won't work when we start adding new public inputs. Do we need a higher level abstraction for PublicInputs?

jtguibas: Maybe we should have PublicValuesCols with transmute so that everything is labeled.

core/src/cpu/air/mod.rs
```diff
@@ -420,6 +434,53 @@ impl CpuChip {
...
+
+        // Verify the index bitmap.
+        let mut bitmap_sum = AB::Expr::zero();
+        for bit in ecall_columns.index_bitmap.iter() {
+            builder.when(local.selectors.is_ecall).assert_bool(*bit);
+            bitmap_sum += (*bit).into();
+        }
+        builder.when(is_commit.clone()).assert_one(bitmap_sum);
+
+        // Retrieve the public_values_digest_word to check against the one passed into the commit ecall.
```
jtguibas, line 466: This also seems a bit jank TBH because you're using public values as a way to detect if it's an interaction builder. I wonder if there's a better way we can do it?

Maybe we should expose something like `builder.is_interaction_builder()`?

zkvm/entrypoint/src/syscalls/halt.rs
```diff
@@ -1,11 +1,42 @@
...
+        // Convert the digest bytes into words, since we will be calling COMMIT ecall with
+        // the words as a parameter.
+        let pv_digest_words: [u32; PV_DIGEST_NUM_WORDS] = pv_digest_bytes
+            .chunks_exact(4)
+            .map(|chunk| u32::from_le_bytes(chunk.try_into().unwrap()))
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
+
+        // For each digest word, call COMMIT ecall.  In the runtime, this will store the digest words
```
jtguibas, line 33: nit: weird double spaces afters sentences

core/src/cpu/air/mod.rs
kevjue: I feel like this should be done in the follow up PRs to this one that adds other public inputs.  This PR is already very big.

zkvm/entrypoint/src/lib.rs
```diff
@@ -29,6 +29,7 @@ macro_rules! entrypoint {
     };
 }
 
+// The number of 32 bit words that the public values digest is composed of.
```
ctian1, line 32: nit: comments like this describing a variable should have triple slash

zkvm/entrypoint/Cargo.toml
```diff
@@ -13,7 +13,7 @@ once_cell = "1.8.0"
 rand = "0.8.5"
 serde = { version = "1.0.196", features = ["derive"] }
 libm = { version = "0.2.8", optional = true }
-sha2 = "0.10.8"
+sha2 = { git = "https://github.com/succinctbot/RustCrypto-hashes.git" }
```
ctian1, line 16: this should be `https://github.com/sp1-patches/RustCrypto-hashes` with branch `patch-v0.10.8`

core/src/cpu/columns/ecall.rs
```diff
@@ -20,6 +20,10 @@ pub struct EcallCols<T> {
 
     /// Whether the current ecall is a COMMIT.
     pub is_commit: IsZeroOperation<T>,
+    /// The digest word passed into the COMMMIT ecall.  This is the public values digest word calculated
```
ctian1, line 23: nit: double space

core/src/air/public_values.rs
```diff
@@ -8,22 +8,22 @@ use std::{
...
 
-/// The PublicValuesDigest struct is used to represent the public input digest.  This is the hash of all the
-/// bytes that the guest program has written to public input.
+/// The PublicValuesDigest struct is used to represent the public values digest.  This is the hash of all the
+/// bytes that the guest program has written to public values.
 #[derive(Serialize, Deserialize, Clone, Copy, Default, Debug)]
 pub struct PublicValuesDigest<T>([T; PV_DIGEST_NUM_WORDS]);
 
-/// Convertion from a byte array into a PiDigest<u32>.
+/// Conversion from a byte array into a PiDigest<u32>.
```
ctian1, line 24: nit: still see PiDigest in this file a few places

core/src/cpu/columns/ecall.rs
jtguibas: triple M

examples/fibonacci/script/Cargo.toml
```diff
@@ -5,6 +5,8 @@ name = "fibonacci-script"
 edition = "2021"
 
 [dependencies]
+p3-baby-bear = { git = "https://github.com/Plonky3/Plonky3.git", branch = "sp1" }
```
jtguibas, line 8: why do we need p3 in fibonacci script?

core/src/cpu/air/mod.rs
jtguibas: Sounds good.

jtguibas (approved):

PR #458, opened 2024-03-31
feat: relay proofs

mattstam:
After remote proving, the proof can be broadcast to each NetworkGateway on the specified chains:
```
PROVER_NETWORK_ACCESS_TOKEN=... CHAINS=11155111,17000 CALLBACKS=0xDEd0000E32f8F40414d3ab3a830f735a3553E18e,0xDEd0000E32f8F40414d3ab3a830f735a3553E18e CALLBACK_DATAS=0x,0x cargo run --release

[4/5] Proving shards...
⠸ [26s] [-------------------------------------] 0/1 (eta 0s)
Proof succeeded

2024-03-31T16:43:56.772252Z  INFO CHAINS is set, proving on-chain    
Broadcast to chain 11155111 succeeded with tx hash: "0x9139145061b67d3fb70c4b4c1444388ba6a0d169331f3790933ffa482ef9d199"
Broadcast to chain 17000 succeeded with tx hash: "0x4620fe537b5144ba454034f0ff41fd73f45258bab17db501653a21946da2605b"
2024-03-31T16:44:08.355864Z  INFO Proofs broadcasted successfully    
successfully generated and verified proof for the program!
```


sdk/src/io.rs
```diff
@@ -1,12 +1,12 @@
...
 use sp1_core::utils::Buffer;
 
 /// Standard input for the prover.
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Deserialize)]
 pub struct SP1Stdin {
-    pub buffer: Vec<Vec<u8>>,
-    #[serde(skip)]
-    pub ptr: usize,
+    pub buffer: Buffer,
```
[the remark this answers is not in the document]
mattstam: rebased

sdk/src/client.rs
```diff
@@ -7,26 +7,29 @@ use reqwest::{
...
+        GetRelayStatusResponse, NetworkServiceClient, ProofStatus, RelayProofRequest,
+        SubmitProofRequest, TransactionStatus,
     },
-    SP1ProofWithIO, SP1Stdin,
+    SP1ProofWithIO,
 };
 
-pub struct SP1ProverServiceClient {
+const DEFAULT_PROVER_NETWORK_RPC: &str = "https://rpc.succinct.xyz/";
+pub const SP1_VERIFIER_ADDRESS: &str = "0x9a39f368676f7a5cbbfe8ea33c258c4536b5398f";
```
ctian1, line 25: should this be configurable by env variable as well?

sdk/src/io.rs
```diff
@@ -140,45 +142,41 @@ pub mod proof_serde {
...
             stdin.write(&3u32);
             let proof = SP1Prover::prove(FIBONACCI_IO_ELF, stdin).unwrap();
             let json = serde_json::to_string(&proof).unwrap();
-            let output = serde_json::from_str::<SP1ProofWithIO<BabyBearPoseidon2>>(&json).unwrap();
+            let output = serde_json::from_str::<SP1ProofWithIO<_>>(&json).unwrap();
             SP1Verifier::verify(FIBONACCI_IO_ELF, &output).unwrap();
         }
 
-        /// Tests serialization with a binary encoding
+        /// Tests serialization with MsgPack encoding
```
ctian1, line 170: we should also have a bincode test

sdk/src/lib.rs
```diff
@@ -109,6 +117,100 @@ impl SP1Prover {
...
+                let access_token = env::var("PROVER_NETWORK_ACCESS_TOKEN")?;
+
+                let chain_ids: Vec<u32> = Self::parse_env_var_array("CHAINS")?;
+                let callbacks: Vec<String> = Self::parse_env_var_array("CALLBACKS")?;
+                let callback_datas: Vec<String> = Self::parse_env_var_array("CALLBACK_DATAS")?;
+
+                if !(chain_ids.len() == callbacks.len() && callbacks.len() == callback_datas.len())
+                {
+                    anyhow::bail!("CHAINS, CALLBACKS, and CALLBACK_DATAS must be of the same size");
+                }
```
ctian1, line 189: I don't think it makes sense to configure all of this through env

ctian1 (changes requested):

sdk/src/lib.rs
mattstam: Yeah I agree, in a subsequent PR I can add functions where you just pass in stuff like this so that an operator can use this lib. 

I don't know the best design for those interfaces yet (probably will start with tendermint) so this PR will just focus on the base mechanics.

sdk/src/io.rs
mattstam: I removed the dependency though. Why is bincode important?

sdk/src/client.rs
mattstam: updated so that you can override with SP1_VERIFIER_ADDRESS env var

sdk/src/io.rs
mattstam: updated to leave it

ctian1 (approved):

sdk/src/lib.rs
```diff
@@ -109,6 +117,100 @@ impl SP1Prover {
         }
     }
 
+    pub async fn relay_remote(
+        access_token: String,
+        proof_id: &str,
+        chain_ids: Vec<u32>,
+        callbacks: Vec<&str>,
+        callback_datas: Vec<&str>,
```
ctian1, line 125: these should not be string, and they should be slice (not owned) instead of vec (owned), maybe it should be `&[[u8;32]]` and `&[Vec<u8>]`

mattstam: [u8; 20] but yeah sgtm, updated

PR #466, opened 2024-04-02
feat: shard transition public values

kevjue:


core/src/stark/machine.rs
```diff
@@ -240,13 +240,25 @@ impl<SC: StarkGenericConfig, A: MachineAir<Val<SC>>> MachineStark<SC, A> {
         tracing::debug_span!("observe challenges for all shards").in_scope(|| {
             proof.shard_proofs.iter().for_each(|proof| {
                 challenger.observe(proof.commitment.main_commit.clone());
+                challenger.observe(proof.public_values.shard);
```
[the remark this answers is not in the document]
kevjue: The commitment_digest would be observed for all shards though.  And that would increase the verification cost (for both the rust one and the recursive one).

core/src/cpu/air/mod.rs
```diff
@@ -485,12 +526,23 @@ impl CpuChip {
...
+            .when(is_halt.clone() + local.selectors.is_unimpl)
             .assert_zero(next.is_real);
+
+        // // If we're halting, the public values next pc should be 0.
+        builder
+            .when(is_halt.clone())
+            .assert_zero(public_values.last_row_next_pc);
+
+        // // Verify the exit code.
+        builder
```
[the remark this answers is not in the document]
kevjue: ya, it's op_b_val

core/src/runtime/record.rs
```diff
@@ -280,6 +280,22 @@ impl MachineRecord for ExecutionRecord {
...
+                // read it (e.g. when the halt instruction is the only instruction in the last shard).
+                // It seems overly complex to set the public_values_digest for the last two shards, so we just set it
+                // for all of the shards.
+                shard.public_values.committed_value_digest =
+                    self.public_values.committed_value_digest;
+
+                shard.public_values.shard = shard.cpu_events[0].shard;
+                shard.public_values.first_row_clk = shard.cpu_events[0].clk;
+                shard.public_values.first_row_pc = shard.cpu_events[0].pc;
+                shard.public_values.last_row_next_clk =
```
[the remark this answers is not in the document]
kevjue: I ended up keeping `next_pc` in the CPU event since that will need to be an actual column in the cpu event.

core/src/air/public_values.rs
```diff
@@ -1,96 +1,75 @@
...
+use std::iter::once;
 
 use super::Word;
 
-// TODO:  Create a config struct that will store the num_words setting and the hash function
-//        and initial entropy used.
 pub const PV_DIGEST_NUM_WORDS: usize = 8;
 
-/// The PublicValuesDigest struct is used to represent the public values digest.  This is the hash of all the
+/// The PublicValues struct is used to represent the public values digest.  This is the hash of all the
```
ctian1, line 11: nit: comment should be updated since it's not just digest anymore

core/src/cpu/air/mod.rs
```diff
@@ -506,8 +541,53 @@ impl CpuChip {
...
+        // Verify the public value's first row pc.
+        builder
+            .when_first_row()
+            .assert_eq(public_values.first_row_pc.clone(), local.pc);
+
+        // If the last real row is not the last row, verify the public value's next pc.
+        builder
+            .when_transition()
+            .when(local.is_real - next.is_real)
+            .assert_eq(public_values.last_row_next_pc.clone(), local.next_pc);
```
ctian1, line 573: comment is a bit confusing to me. suggestion: "If this row is the last real row, verify public_values.last_row_next_pc".

Also, do we not have to worry about when local.is_real = 0 and next.is_real = 1? seems like that shouldn't happen because first row probably needs to be real?

core/src/cpu/event.rs
```diff
@@ -15,6 +15,10 @@ pub struct CpuEvent {
     /// The current program counter.
     pub pc: u32,
 
+    /// The value of the next instruction's program counter. This value needs to be make public for
```
ctian1, line 18: nit: make -> made

core/src/cpu/air/mod.rs
```diff
@@ -230,9 +255,21 @@ impl CpuChip {
...
             .when_transition()
             .when(next.is_real)
-            .when(local.selectors.is_jal + local.selectors.is_jalr)
+            .when(is_jump_instruction.clone())
             .assert_eq(jump_columns.next_pc.reduce::<AB>(), next.pc);
 
+        // When is_jump_instruction is true, assert that local.is_real is true.
+        builder
+            .when(is_jump_instruction.clone())
+            .assert_one(local.is_real);
```
ctian1, line 264: Do we need the checks like this one? and for is_branching, is_sequential_instr. Seems like the constraints on is_real by itself (first row is real, once not real can't become real again) should be enough.

core/src/cpu/air/mod.rs
kevjue: We don't have to worry about that.  There are constraints that check that there is at most 1 at_real transition and that at_real starts at 1.

core/src/cpu/air/mod.rs
kevjue: ya, agreed.

core/src/cpu/air/branch.rs
```diff
@@ -67,6 +77,16 @@ impl CpuChip {
                 .when(next.is_real)
                 .when(local.not_branching)
                 .assert_eq(local.pc + AB::Expr::from_canonical_u8(4), next.pc);
+
+            // When local.not_branching is true, assert that local.is_real is true.
+            builder.when(local.not_branching).assert_one(local.is_real);
```
ctian1, line 78: nit: I think this is also not needed?

core/src/cpu/air/branch.rs
```diff
@@ -44,13 +44,23 @@ impl CpuChip {
...
-            // When we are branching, assert that local.pc <==> branch_columns.next_pc as Word.
+            // When we are branching, assert that next.pc <==> branch_columns.next_pc as Word.
             builder
                 .when_transition()
                 .when(next.is_real)
                 .when(local.branching)
                 .assert_eq(branch_cols.next_pc.reduce::<AB>(), next.pc);
 
+            // When local.branching is true, assert that local.is_real is true.
+            builder.when(local.branching).assert_one(local.is_real);
```
ctian1: nit: this is also not needed I think?

core/src/cpu/air/mod.rs
```diff
@@ -433,22 +462,33 @@ impl CpuChip {
         local: &CpuCols<AB::Var>,
         next: &CpuCols<AB::Var>,
         is_branch_instruction: AB::Expr,
+        is_halt: AB::Expr,
     ) {
+        // Verify that if is_sequential_instr is true, assert that local.is_real is true.
+        // This is needed for the following constraint, which is already degree 3.
+        builder
+            .when(local.is_sequential_instr)
+            .assert_one(local.is_real);
```
ctian1, line 471: nit: not needed?

core/src/cpu/air/mod.rs
```diff
@@ -506,8 +535,53 @@ impl CpuChip {
...
+        // If the last real row is the last row, verify the public value's next pc.
+        builder
+            .when_last_row()
+            .when(local.is_real)
+            .assert_eq(public_values.next_pc.clone(), local.next_pc);
+
+        // If we're halting, verify that the public value's next pc == 0.
+        builder
+            .when(is_halt.clone())
+            .assert_one(public_values.last_instr_halt.clone());
```
ctian1, line 578: I think we also need to constrain when (1-is_halt), last_instr_halt == 0

core/src/cpu/air/branch.rs
```diff
@@ -44,13 +44,19 @@ impl CpuChip {
...
                 .when_transition()
                 .when(next.is_real)
                 .when(local.branching)
                 .assert_eq(branch_cols.next_pc.reduce::<AB>(), next.pc);
 
+            // When the last row is real and local.branching, assert that local.next_pc <==> branch_columns.next_pc as Word.
+            builder
+                .when_last_row()
+                .when(local.branching)
+                .assert_eq(branch_cols.next_pc.reduce::<AB>(), local.next_pc);
```
ctian1, line 58: is local.next_pc constrained properly when the last real row is not last row?

core/src/cpu/air/branch.rs
```diff
@@ -44,13 +44,19 @@ impl CpuChip {
...
 
-            // When we are branching, assert that local.pc <==> branch_columns.next_pc as Word.
+            // When we are branching, assert that next.pc <==> branch_columns.next_pc as Word.
             builder
                 .when_transition()
                 .when(next.is_real)
                 .when(local.branching)
                 .assert_eq(branch_cols.next_pc.reduce::<AB>(), next.pc);
 
+            // When the last row is real and local.branching, assert that local.next_pc <==> branch_columns.next_pc as Word.
```
ctian1, line 54: nit: "last row" -> "current row"

ctian1 (approved):

PR #475, opened 2024-04-04
feat: full prover crate

ctian1:


prover/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+use std::time::Instant;
+
+use sp1_core::{
+    runtime::Program,
+    stark::{Proof, RiscvAir, StarkGenericConfig, VerifyingKey},
+    utils::{run_and_prove, BabyBearPoseidon2},
+};
+use sp1_recursion_core::runtime::Runtime;
+use sp1_recursion_program::compress::build_compress;
+
```
jtguibas, line 13: Eventually, I think it would be nicer to have these methods on a struct to make it more object oriented.

jtguibas (approved):

PR #476, opened 2024-04-04
feat: add support for witness in programs

jtguibas:


core/src/stark/machine.rs
```diff
@@ -94,6 +95,13 @@ impl<SC: StarkGenericConfig, A: MachineAir<Val<SC>>> MachineStark<SC, A> {
             .sorted_by_key(|chip| chip_ordering.get(&chip.name()))
     }
 
+    pub fn chips_sorted_indices(&self, proof: &ShardProof<SC>) -> Vec<usize> {
+        self.chips()
+            .iter()
+            .map(|chip| *proof.chip_ordering.get(&chip.name()).unwrap_or(&9999))
```
ctian1, line 101: why is this .unwrap_or(&9999)?

recursion/core/src/runtime/mod.rs
```diff
@@ -533,6 +538,24 @@ where
                     }
                     (a, b, c) = (a_val, b_val, c_val);
                 }
+                Opcode::HintLen => {
+                    let (a_ptr, b_val, c_val) = self.alu_rr(&instruction);
+                    self.mr(a_ptr, MemoryAccessPosition::A);
+                    let a_val: Block<F> =
+                        F::from_canonical_usize(self.witness_stream[0].len()).into();
+                    self.mw(a_ptr, a_val, MemoryAccessPosition::A);
```
ctian1, line 546: I don't think you can do .mr and then .mw, I think there should only be .mw?

ctian1 (approved):

PR #483, opened 2024-04-06
feat: sdk using secp256k1 auth

mattstam:
The `rpc.succinct.xyz` calls that could use computational resources from the network (e.g. `CreateProof`) requires a signature instead of an access key. The account address gets recovered from the signature, and the account must have a legitimate balance for the request to succeed.

The messages themselves follow [EIP-712](https://eips.ethereum.org/EIPS/eip-712), where the domain is `"succinct"` and the typestruct changes depending on which endpoint is being used. This standard is best practice and if we ever want to expose this functionality on the frontend (like to make a demo proof request) then it will show nicely on wallets like Metamask.

Also cleans up our SDK library a bit by letting operators initialize a `ProverClient` that can be configured to prove locally or remotely. There can be a switch from local to remote proving with **zero code change** by specifying a `PRIVATE_KEY`.


book/generating-proofs/advanced.md
```diff
@@ -64,23 +63,26 @@ You can either use `utils::setup_logger()` or `utils::setup_tracer()` to enable
...
 utils::setup_logger();
 ```
 
 You must run your command with:
+
 ```bash
 RUST_LOG=info cargo run --release
-```
\ No newline at end of file
+```
```
mattstam, line 88: random markdown auto-formatting

core/src/lib.rs
```diff
@@ -115,8 +111,8 @@ impl SP1Prover {
     }
 }
 
-impl SP1Verifier {
-    /// Verify a proof generated by `SP1Prover`.
+impl VerifierClient {
+    /// Verify a proof generated by `ProverClient`.
```
ratankaliani, line 115: IMO, it's more clear to combine the methods from the `ProverClient` and `VerifierClient` into an `SP1Client`, as there's no hosted/local delineation in the `VerifierClient`.

ratankaliani: And also simplifies imports.

mattstam: [merged VeriferClient into ProverClient](https://github.com/succinctlabs/sp1/pull/483/commits/a1c37d74a46d0f569fe85c909dafff5f7d4211ef)

sdk/src/auth.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        Ok(signature.as_bytes().to_vec())
+    }
+
+    pub async fn sign_submit_proof_message(&self, proof_id: &str) -> Result<Vec<u8>> {
+        let domain_seperator = Self::get_domain_separator();
+
+        let type_struct = SubmitProof {
+            proof_id: proof_id.to_string(),
+        };
```
ratankaliani, line 61: nit: Add docs for `NetworkAuth`, as this will be commonly used & it's important to understand what's being signed.

sdk/src/client.rs
```diff
@@ -22,30 +21,24 @@ use crate::{
 };
 
 const DEFAULT_PROVER_NETWORK_RPC: &str = "https://rpc.succinct.xyz/";
-const DEFAULT_SP1_VERIFIER_ADDRESS: &str = "0x9a39f368676f7a5cbbfe8ea33c258c4536b5398f";
+const DEFAULT_SP1_VERIFIER_ADDRESS: &str = "0xed2107448519345059eab9cddab42ddc78fbebe9";
```
ratankaliani, line 24: Is this across all chains?

mattstam:
`SP1_VERIFIER_ADDRESS` is the default address for the verifier on all chains, which can be overriden.

sdk/src/auth.rs
mattstam: added

PR #492, opened 2024-04-08
perf(simd): avx512 by dyxushuai

jtguibas:


ctian1 (approved):

weikengchen:
We should let a16z know that SP1 just gets 2x faster :)

PR #499, opened 2024-04-10
feat: recursion optimizations

jtguibas:


recursion/program/src/compress.rs
```diff
@@ -0,0 +1,164 @@
+use std::time::Instant;
```
ctian1, line 1: I think this file came back from merge conflicts, I renamed it to reduce.rs

ctian1 (approved):

PR #505, opened 2024-04-11
feat(sdk): add `prove_async`

ratankaliani:
Add `async` context for `prove` for use in `async` runtimes, as the current `prove` function is incompatible due to the use of `block_on`.


sdk/src/lib.rs
```diff
@@ -85,68 +86,92 @@ impl ProverClient {
...
+            self.prove_remote_sync(elf, stdin)
         } else {
             println!("Proving locally");
             self.prove_local(elf, stdin, BabyBearPoseidon2::new())
         }
     }
 
-    pub fn prove_remote(
+    /// Generate a proof for the execution of the ELF with the given public inputs for an async context
+    pub async fn prove_async(
```
mattstam, line 97: kind of weird to have *_async and *_sync functions - I think we should name the async version of `prove_remote` to `prove_remote_async`, and leave the sync version as just `prove_remote`.

sdk/src/lib.rs
```diff
@@ -85,68 +86,92 @@ impl ProverClient {
...
+        &self,
+        elf: &[u8],
+        stdin: SP1Stdin,
+    ) -> Result<SP1ProofWithIO<BabyBearPoseidon2>> {
+        if self.client.is_some() {
+            println!("Proving remotely");
+            self.prove_remote(elf, stdin).await
+        } else {
+            println!("Proving locally");
+            self.prove_local(elf, stdin, BabyBearPoseidon2::new())
```
mattstam, line 107: I think `prove_async` is is technically inaccurate in this context because `prove_local` is sync, maybe there is a better way to do it?

mattstam (changes requested):

sdk/src/lib.rs
ratankaliani: What do you think the interface should be for `prove_async`?

Either:
- Add `prove_local_async`
- Remove `prove_local` from `prove_async` and only expose `prove_remote_async`

sdk/src/lib.rs
ratankaliani: Updated!

sdk/src/lib.rs
ratankaliani: Removed `prove_async` for now.

mattstam (approved):

PR #507, opened 2024-04-11
feat: nonce in signed messages

mattstam:
RPC messages now accept a signed nonce to verify them against the stored nonce. The goal of signing and verifying nonces is to prevent replay attacks and out-of-order messages.

This is achieved by first calling `get_nonce()` before signing the message, then signing the message that includes it.


sdk/src/client.rs
```diff
@@ -70,6 +70,17 @@ impl NetworkClient {
...
+    pub async fn get_nonce(&self) -> u64 {
+        let res = self
+            .rpc
+            .get_nonce(GetNonceRequest {
+                address: self.auth.get_address().to_vec(),
+            })
+            .await
+            .unwrap();
+        res.nonce
+    }
```
ratankaliani, line 82: docs: Add a more clear specification for what a nonce is and where you're querying it from? i.e. querying from sequencer

sdk/src/auth.rs
```diff
@@ -79,6 +87,7 @@ impl NetworkAuth {
     /// Signs a message to relay a proof.
     pub async fn sign_relay_proof_message(
         &self,
```
ratankaliani, line 89: docs: This is unclear. This should specify that a user is requesting for a hosted relay of a proof.

sdk/src/auth.rs
```diff
@@ -63,10 +70,11 @@ impl NetworkAuth {
     }
 
     /// Signs a message to submit a proof.
```
ratankaliani, line 72: docs: Signs a message to mark a proof as ready for proof generation.

sdk/src/auth.rs
```diff
@@ -50,11 +53,15 @@ impl NetworkAuth {
         }
     }
 
+    pub fn get_address(&self) -> [u8; 20] {
+        *self.wallet.address().0
+    }
+
     /// Signs a message to create a proof.
```
ratankaliani, line 60: docs: Clarify what this fn does.

ratankaliani (approved):
LGTM, left some small comments on adding more docs for each stage of requesting proof generation which will be helpful as the protocol evolves.

sdk/src/client.rs
mattstam: added

sdk/src/auth.rs
mattstam: added

sdk/src/auth.rs
mattstam: added

PR #514, opened 2024-04-12
feat: verify pc and shard transition in recursive proofs

kevjue:


recursion/program/src/reduce.rs
```diff
@@ -164,15 +177,56 @@ pub fn build_reduce_program(setup: bool) -> RecursionProgram<Val> {
...
+
+                let pv_next_pc = felt_to_var(builder, pv.next_pc);
+                builder.if_eq(pv_next_pc, zero).then_or_else(
+                    |builder| {
+                        builder.assert_felt_eq(shard_next_shard, zero_felt);
+                    },
+                    |builder| {
+                        builder.assert_felt_eq(shard_next_shard, pv_shard_plus_one);
+                    },
+                );
```
ctian1, line 224: do we need to witness/commit next_shard ? don't we only need shard?

recursion/program/src/reduce.rs
```diff
@@ -208,7 +260,27 @@ pub fn build_reduce_program(setup: bool) -> RecursionProgram<Val> {
...
             // Recursive proof
             |builder| {
-                // TODO: Verify proof public values
+                let mut pv_elements = Vec::new();
+                for i in 0..PROOF_MAX_NUM_PVS {
+                    let element = builder.get(&proof.public_values, i);
+                    pv_elements.push(element);
+                }
+
+                let proof_pv = RecursionPublicValues::<Felt<_>>::from_vec(pv_elements);
```
ctian1, line 269: are we planning to remove the hashing part and just commit to all the public values directly?

ctian1 (approved):

recursion/program/src/reduce.rs
kevjue: I plan to still keep the hashing part, but the individual public values will be more structured instead of an array.

recursion/program/src/reduce.rs
kevjue: We don't for the first layer of verifying SP1 proofs, but I believe we need them for the higher layers.  Basically knowing what is the shard range for a recursive proof.

kevjue: To your point, the `next_shard` data is not really needed for the first layer of proof verification, so this constraint can be removed.

PR #524, opened 2024-04-15
fix: use REMOTE_PROVE to check local or remote

mattstam:
Instead of directly choosing to remote prove off of `PRIVATE_KEY`, check for `REMOTE_PROVING=true`.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -57,10 +57,11 @@ impl ProverClient {
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
         dotenv::dotenv().ok();
-        let private_key = env::var("PRIVATE_KEY").unwrap_or_default();
-        if private_key.is_empty() {
+        if env::var("REMOTE_PROVING").is_err() {
             Self { client: None }
         } else {
+            let private_key =
+                env::var("PRIVATE_KEY").unwrap_or_else(|_| panic!("PRIVATE_KEY is not set"));
             Self {
                 client: Some(NetworkClient::new(&private_key)),
             }
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -57,9 +57,12 @@ impl ProverClient {
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
         dotenv::dotenv().ok();
-        let remote_proving = env::var("REMOTE_PROVING").unwrap_or_else(|_| String::from("false"));
+        let remote_proving = env::var("REMOTE_PROVING")
+            .unwrap_or_else(|_| String::from("false"))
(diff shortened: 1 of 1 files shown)


sdk/src/lib.rs
```diff
@@ -57,13 +57,16 @@ impl ProverClient {
...
         dotenv::dotenv().ok();
-        let private_key = env::var("PRIVATE_KEY").unwrap_or_default();
-        if private_key.is_empty() {
-            Self { client: None }
-        } else {
+        let remote_proving = env::var("REMOTE_PROVING").unwrap_or_else(|_| String::from("false"));
+
+        if remote_proving == "true" {
+            let private_key =
+                env::var("PRIVATE_KEY").unwrap_or_else(|_| panic!("PRIVATE_KEY is not set"));
```
ratankaliani, line 64: nit: `PRIVATE_KEY` must be set for remote proving.

sdk/src/lib.rs
```diff
@@ -57,13 +57,16 @@ impl ProverClient {
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
         dotenv::dotenv().ok();
-        let private_key = env::var("PRIVATE_KEY").unwrap_or_default();
-        if private_key.is_empty() {
-            Self { client: None }
-        } else {
+        let remote_proving = env::var("REMOTE_PROVING").unwrap_or_else(|_| String::from("false"));
+
+        if remote_proving == "true" {
```
ratankaliani, line 62: nit: Better practice to read this into a `bool`, but don't feel too strongly.

mattstam: updated

sdk/src/lib.rs
mattstam: updated

sdk/src/lib.rs
```diff
@@ -57,13 +57,19 @@ impl ProverClient {
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
         dotenv::dotenv().ok();
-        let private_key = env::var("PRIVATE_KEY").unwrap_or_default();
-        if private_key.is_empty() {
-            Self { client: None }
-        } else {
+        let remote_proving = env::var("REMOTE_PROVING")
```
ratankaliani, line 60: nit: `REMOTE_PROVE`

ratankaliani (commented):
LGTM, we can revisit in the future if users are confused how to delineate between a relaying `PRIVATE_KEY` and a `PRIVATE_KEY` for remote requests.

ratankaliani (approved):

sdk/src/lib.rs
mattstam: added

PR #530, opened 2024-04-16
feat: execute before `prove_remote_async`

ratankaliani:
Execute before `prove_remote_async` in the SDK. If a user attempts to request a proof with a failing runtime, it will be caught at this point.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -100,8 +100,8 @@ impl ProverClient {
     }
 
     // Generate a proof remotely using the Succinct Network in an async context.
-    // TODO: If the simulation of the runtime is expensive for user programs, add an optional flag to skip it. This
-    // shouldn't be the case for the vast majority of user programs.
+    // Note: If the simulation of the runtime is expensive for user programs, we can add an optional
+    // flag to skip it. This shouldn't be the case for the vast majority of user programs.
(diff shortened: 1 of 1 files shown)


sdk/src/lib.rs
```diff
@@ -110,6 +110,11 @@ impl ProverClient {
             .as_ref()
             .ok_or_else(|| anyhow::anyhow!("Network client not initialized"))?;
 
+        // Execute the runtime before creating the proof request.
+        let mut runtime = Runtime::new(Program::from(elf));
+        runtime.write_vecs(&stdin.buffer);
+        runtime.run();
```
mattstam, line 116: maybe we should log `simulation complete` or something, that way people have a better understanding if this does take a significant period of time

sdk/src/lib.rs
```diff
@@ -110,6 +110,11 @@ impl ProverClient {
...
+
```
mattstam, line 117: could be overkill, but an optional `skip_simulation` bool that is false by default might be nice

mattstam (approved):
lgtm just a few optional things to consider

sdk/src/lib.rs
ratankaliani: Will skip this for now, left a TODO about how we can add this for later if there's demand.

PR #538, opened 2024-04-17
chore:  re-organise cpu air constraints

kevjue:
This PR re-organised the CPU air constraints.  It has the following changes:

1) Create a new file `register.rs` and moved all the register memory access constraints to there.
2) Created a new file `ecall.rs` and moved all the ecall related methods and constraints to there.
3) Moved all memory related constraints in `mod.rs` to `memory.rs`.
4) Created `is_real_eval` method in `mod.rs` and moved `is_real` related constraints to there.


core/src/cpu/air/ecall.rs
```diff
@@ -0,0 +1,224 @@
...
+        builder.send_syscall(
+            local.shard,
+            local.clk,
+            syscall_id,
+            local.op_b_val().reduce::<AB>(),
+            local.op_c_val().reduce::<AB>(),
+            local.ecall_mul_send_to_table,
+        );
+
+        // Constrain EcallCols.is_enter_unconstrained.result == syscall_id is ENTER_UNCONSTRAINED.
```
jtguibas, line 52: Not sure if this return pattern is really getting you anything, maybe we should just use ecall_cols directly?

core/src/cpu/air/ecall.rs
```diff
@@ -0,0 +1,224 @@
...
+impl CpuChip {
+    /// Whether the instruction is an ECALL instruction.
+    pub(crate) fn is_ecall_instruction<AB: SP1AirBuilder>(
+        &self,
+        opcode_selectors: &OpcodeSelectorCols<AB::Var>,
+    ) -> AB::Expr {
+        opcode_selectors.is_ecall.into()
+    }
+
+    /// Constraints related to the ECALL opcode.
```
jtguibas, line 21: Here i think it would be helpful to have a more detailed description of what's going on like the comment doesn't really tell you anything, like you can be like

1) Handle whether we need to send an ecall interaction to another table.
2) Compute the flags in EcallCols.
3) ...

jtguibas: It needs to come across as more organized as rather than just a list of random constraints for it to be maximally understandable

core/src/cpu/air/memory.rs
```diff
@@ -43,6 +43,61 @@ impl CpuChip {
         opcode_selectors.is_sb + opcode_selectors.is_sh + opcode_selectors.is_sw
     }
 
+    /// Constrains the addr_aligned, addr_offset, and addr_word memory columns.
+    ///
+    /// This method will first constrain the addr_word memory column by ensuring it equals to
```
jtguibas, line 48: This comment here also really doesn't say much / it's too low level (same abstraction level as the code itself), like I think it's better to be like

1) Calculate that the unaligned address is correctly computed to be `op_b.value + op_c.value`.
2) Calculate that the address offset is address % 4.
3) Assert the validity of the aligned address given the address offset and the unaligned address.

jtguibas (commented):
Looks a lot cleaner! I left some nits. They are not that important functionality wise but I think it would be good to note for future PRs.

jtguibas (approved):

PR #545, opened 2024-04-18
refactor: fixing dep tree for `prover`, `recursion`, `core` and `sdk`

ratankaliani:
## Overview
`prover` will contain the majority of the proof/verification logic and imports `recursion` and `core`. `sdk` is now a user-facing crate that uses methods from `prover`.


jtguibas (approved):

jtguibas:
@ratankaliani I'm merging this but I still think there is more work on fixing the rexports.

PR #546, opened 2024-04-18
feat: update network client with claim and fulfill

mattstam:
Updates the SDK so that provers can call `claimProof()` and `fulfillProof()` on the prover network RPC.


sdk/src/client.rs
```diff
@@ -87,6 +89,71 @@ impl NetworkClient {
...
+                .await?
+                .bytes()
+                .await?;
+            Some(bincode::deserialize(&proof).expect("Failed to deserialize proof"))
+        } else {
+            None
+        };
+
+        Ok((res, proof))
+    }
```
ratankaliani, line 117: Add doc comments for each of these functions (as they're public-facing).

sdk/src/lib.rs
```diff
@@ -123,33 +121,24 @@ impl ProverClient {
         let proof_id = client.create_proof(elf, &stdin).await?;
         println!("proof_id: {:?}", proof_id);
 
-        let mut pb = StageProgressBar::new();
```
ratankaliani, line 126: Why did we remove the progress bar?

sdk/src/lib.rs
```diff
@@ -123,33 +121,24 @@ impl ProverClient {
...
 
             match status.status() {
-                ProofStatus::ProofSucceeded => {
+                ProofStatus::ProofFulfilled => {
                     println!("Proof succeeded");
-                    pb.finish();
                     if let Some(proof) = maybe_proof {
                         return Ok(proof);
                     } else {
                         return Err(anyhow::anyhow!("Proof succeeded but no proof available"));
```
ratankaliani, line 135: The logic here is a bit unclear. 

If we fail to fetch the proof with proof status fulfilled, we return an error from the `get_proof_status` function, which then returns another error due to the `?` keyword. If the proof fails to deserialize, we also return an error: `bincode::deserialize(&proof).expect("Failed to deserialize proof")`. 

```rust
let proof = if res.status() == ProofStatus::ProofFulfilled {
            let proof = self
                .http
                .get(res.proof_url.clone())
                .send()
                .await?
                .bytes()
                .await?;
            Some(bincode::deserialize(&proof).expect("Failed to deserialize proof"))
        } else {
            None
        };
```

There doesn't seem to be a state where the error for "proof succeeded but no proof available" occurs, as if the proof is fulfilled, we should always be able to fetch the proof OR we should return an error via `?`.

Given this is the case, we should just unwrap the proof if `ProofFulfilled`, as that's the behavior we expect.

sdk/src/auth.rs
```diff
@@ -85,6 +95,37 @@ impl NetworkAuth {
...
+        let type_struct = ClaimProof {
+            nonce,
+            proof_id: proof_id.to_string(),
+        };
+
+        let message_hash = type_struct.eip712_signing_hash(&domain_seperator);
+        let signature = self.wallet.sign_hash(&message_hash).await?;
+
+        Ok(signature.as_bytes().to_vec())
+    }
```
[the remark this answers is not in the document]
mattstam: updated, this is super nice

sdk/src/client.rs
mattstam: added docs to all of the NetworkClient functions

sdk/src/lib.rs
mattstam: the network no longer has information about how far a proof has progressed

sdk/src/lib.rs
mattstam: added some better error handling to this fn

ratankaliani: My point here was that we should just do `proof.unwrap()` here. By this point, we can assume that any error from the underlying function was handled with the `?` operators instead of:

```rust
if let Some(proof) = maybe_proof {
    ...
```

sdk/src/auth.rs
```diff
@@ -58,31 +68,46 @@ impl NetworkAuth {
         *self.wallet.address().0
     }
 
+    // Generic function to sign any message based on the SolStruct
```
ratankaliani, line 71: nit: add . at end

sdk/src/client.rs
```diff
@@ -131,33 +210,43 @@ impl NetworkClient {
...
-                .await?;
-            Some(bincode::deserialize(&proof).expect("Failed to deserialize proof"))
-        } else {
-            None
-        };
+        Ok(res)
+    }
 
-        Ok((res, proof))
+    // Fulfill a proof. Should only be called after the proof has been uploaded. Only callable by the
```
ratankaliani, line 229: This may be more clear:

Fulfills a proof. Returns an error if 1) the proof has not yet been uploaded or 2) the caller is not the proof claimer.

sdk/src/client.rs
```diff
@@ -131,33 +210,43 @@ impl NetworkClient {
...
+                nonce,
+                proof_id: proof_id.to_string(),
+                signature,
+            })
+            .await?;
+
+        Ok(res)
     }
 
+    // Relay a proof. Should only be called after a proof has been fulfilled.
```
ratankaliani, line 249: nit: Relay a proof. Returns an error if the proof has not been fulfilled yet.

sdk/src/proto/network.rs
```diff
@@ -1,125 +1,234 @@
 // This file is @generated by prost-build.
+/// The request for a nonce. Should be called before signing a message to get the correct nonce for
```
ratankaliani, line 2: nit: Should only -> Returns an error if

sdk/src/proto/network.rs
```diff
@@ -1,125 +1,234 @@
...
+    /// The address to get the nonce for.
     #[prost(bytes = "vec", tag = "1")]
     pub address: ::prost::alloc::vec::Vec<u8>,
 }
+/// The response for a nonce request.
 #[derive(serde::Serialize, serde::Deserialize)]
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct GetNonceResponse {
+    /// The nonce for the given address. Should be signed along with the rest of the message.
```
ratankaliani, line 17: nit: the rest of what message?

The nonce for the given address. Retrieved when signing new messages

sdk/src/proto/network.rs
```diff
@@ -1,125 +1,234 @@
...
+    /// The url to upload the elf to.
     #[prost(string, tag = "2")]
-    pub program_put_url: ::prost::alloc::string::String,
+    pub program_url: ::prost::alloc::string::String,
+    /// The url to upload the stdin to.
     #[prost(string, tag = "3")]
-    pub stdin_put_url: ::prost::alloc::string::String,
+    pub stdin_url: ::prost::alloc::string::String,
 }
+/// The request to submit a proof, the second step in requesting a proof. Should only be called after
```
ratankaliani, line 93: nit: Should only -> Returns an error if the program and stdin have not been uploaded yet.

sdk/src/proto/network.rs
```diff
@@ -1,125 +1,234 @@
...
-    #[prost(uint32, optional, tag = "6")]
-    pub stage_progress: ::core::option::Option<u32>,
-    #[prost(uint32, optional, tag = "7")]
-    pub stage_total: ::core::option::Option<u32>,
+    pub stdin_artifact_id: ::prost::alloc::string::String,
+    /// The artifact identifier for the proof location.
+    #[prost(string, tag = "3")]
+    pub proof_artifact_id: ::prost::alloc::string::String,
+}
+/// The request to fulfill a proof. Should only be called after the proof has been uploaded. Only
```
ratankaliani, line 145: nit: Should only -> Returns an error if

sdk/src/proto/network.rs
```diff
@@ -1,125 +1,234 @@
...
+/// The response for fulfilling a proof, empty on success.
+#[derive(serde::Serialize, serde::Deserialize)]
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct FulfillProofResponse {
+    /// The amount of seconds between proof claim and fulfillment.
+    #[prost(uint64, tag = "1")]
+    pub proving_seconds: u64,
 }
+/// The request to relay a proof through the NetworkGateway on a given chain. Should only be called
```
ratankaliani, line 170: nit: Should only -> Returns an error if the proof has not been fulfilled yet

sdk/src/client.rs
mattstam: Added some error conditions on `FulfillProof()` rpc. Note: it won't return an error if it hasn't been uploaded because that will be done by an async agent that penalizes incorrect proofs, resulting is loss of balance.

sdk/src/client.rs
mattstam: added

sdk/src/proto/network.rs
mattstam: Arguably it's better to follow https://www.ietf.org/rfc/rfc2119.txt standard but Im down to change it to error conditions instead.

sdk/src/proto/network.rs
mattstam: "Retrieved when signing new messages" I dont really think the server should be too strict about the behavior of the client. For instance, a client might want to cache the account's nonce and increment on their end to avoid additional RPC calls

sdk/src/proto/network.rs
mattstam: so yeah something like this `Claim a requested proof, commiting the proving and later fulfilling it. MUST be called proof is in a PROOF_REQUESTED state.` on the RPC spec, lmk if you have any issues with that

sdk/src/proto/network.rs
mattstam: set it as `Get the nonce for a given account. Used to check current nonce, which must match when signing and sending a message.`.

sdk/src/lib.rs
mattstam: Sounds good, updated to just do `maybe_proof.unwrap();`

sdk/src/proto/network.rs
mattstam: added this to client.rs

sdk/src/proto/network.rs
mattstam: added this to client.rs

sdk/src/proto/network.rs
mattstam: added this to client.rs

examples/fibonacci-io/script/src/main.rs
```diff
@@ -31,6 +30,7 @@ fn main() {
     assert_eq!(a, expected_a);
     assert_eq!(b, expected_b);
 
+    println!("n: {}", n);
```
ratankaliani, line 33: nit: Remove

sdk/src/client.rs
```diff
@@ -79,12 +82,86 @@ impl NetworkClient {
...
+
+        let simulation_url = match res.status() {
+            TransactionStatus::TransactionFailed => Some(res.simulation_url.clone()),
+            _ => None,
+        };
+
+        Ok((res, tx_hash, simulation_url))
+    }
+
+    /// Makes a proof request for the given ELF and stdin.
```
ratankaliani, line 164: nit: Create instead of "Makes"

ratankaliani (approved):
LGTM!

PR #551, opened 2024-04-19
chore: cleanup prover

jtguibas:


prover/src/lib.rs
```diff
@@ -1,128 +1,108 @@
...
-        .map(|chip_idx| {
-            let name = chips[chip_idx].name().clone();
-            let prep_sorted_idx = vk.chip_ordering[&name];
-            (prep_sorted_idx, vk.chip_information[prep_sorted_idx].1)
-        })
-        .unzip();
-    (prep_sorted_indices, prep_domains)
+/// A wrapper to abstract proofs representing a range of shards with multiple proving configs.
+#[derive(Serialize, Deserialize)]
+pub enum SP1ReduceProofWrapper {
```
ctian1, line 99: nit: maybe we can call this ReduceableProof

prover/src/lib.rs
```diff
@@ -136,129 +116,191 @@ impl SP1ProverImpl {
...
+                    },
+                })
+            })
+            .collect::<Vec<_>>();
+
+        // Keep reducing until we have only one shard. If we have only one shard, we still want to
+        // wrap it into a reduce shard.
+        while reduce_proofs.len() > 1 {
+            reduce_proofs = self.reduce_layer(vk, core_challenger.clone(), reduce_proofs, 2);
+        }
```
ctian1, line 186: I think we need some more logic here if there's only one shard

ctian1 (approved):

PR #561, opened 2024-04-22
feat(recursion): FRI-fold precompile

Sladuca:
- added unoptimized AIR for fri-fold
- memory lookups are all currently all dummy until we figure out a more optimal way to pack arguments.

Built atop #557.


Sladuca:
closed in favor of #581

PR #568, opened 2024-04-23
feat: a few simplifications to the Keccak AIR

dlubarov:


core/src/syscall/precompiles/keccak256/air.rs
```diff
@@ -64,34 +65,15 @@ where
         );
 
         // Constrain that the inputs stay the same throughout the 24 rows of each cycle
-        let mut not_last_in_cycle = local.keccak.step_flags[0].into();
```
dlubarov, line 67: My `not_final_step` is computed differently, but the two should be guaranteed to give the same result based on the `eval_round_flags` constraints in `KeccakAir`.

core/src/syscall/precompiles/keccak256/air.rs
```diff
@@ -64,34 +65,16 @@ where
...
-            .when(not_last_in_cycle.clone() * local.is_real)
-            .assert_eq(local.clk, next.clk);
-        builder
-            .when_transition()
-            .when(not_last_in_cycle.clone() * local.is_real)
-            .assert_eq(local.state_addr, next.state_addr);
-        builder
-            .when_transition()
-            .when(not_last_in_cycle)
-            .assert_eq(local.is_real, next.is_real);
```
dlubarov, line 86: Just removed filtering by `is_real`, I think dummy rows will still satisfy these consistency checks since they seem to consistently have zeros for these fields?

core/src/syscall/precompiles/keccak256/air.rs
```diff
@@ -64,34 +65,16 @@ where
...
-
-        // The last row must be nonreal because NUM_ROUNDS is not a power of 2. This constraint
-        // ensures that the table does not end abruptly.
-        builder.when_last_row().assert_zero(local.is_real);
```
dlubarov, line 90: I believe a real final row should be okay? It can't have a final 24th row, so it can't export a result.

core/src/syscall/precompiles/keccak256/air.rs
tamirhemo: Indeed! they do have zeros

tamirhemo (approved):
LGTM!

PR #574, opened 2024-04-23
feat(recursion): recursion air builder

kevjue:


recursion/core/src/air/builder.rs
```diff
@@ -0,0 +1,10 @@
+use p3_air::AirBuilderWithPublicValues;
+use sp1_core::air::{BaseAirBuilder, ExtensionAirBuilder};
+
+/// Builder for the SP1 recursion machine AIRs.
+pub trait SP1RecursionAirBuilder:
```
tamirhemo, line 5: I think it would be nice to have 
`MachineAirBuilder: BaseAirBuilder + ExtensionAirBuilder + AirBuilderWithPublicValues`
in core so that we don't need to give the same bounds again

kevjue: just to clarify, you are proposing that there is a "subtrait" that contains the common builder traits between `SP1AirBuilder` and `SP1RecursionAirBuilder`?

tamirhemo: super trait but yes.

tamirhemo (approved):
LGTM!

PR #594, opened 2024-04-25
feat: groth16 server

mattstam:
Intializing a Groth16Prover starts up a server, which loads the R1CS and PK once. Then, requests can be sent to it.


recursion/gnark/go.mod
```diff
@@ -1,19 +1,41 @@
 module github.com/succinctlabs/sp1-recursion-gnark
 
-go 1.20
+go 1.22.1
```
mattstam, line 3: note: I'm bumping this just because Go added some nice http features in std lib, but if it's a problem because of Gnark audits lmk

recursion/gnark-ffi/src/groth16.rs
```diff
@@ -26,9 +33,46 @@ pub struct Groth16Proof {
...
+        // Start the Gnark server in a separate thread.
+        let rt = Runtime::new().unwrap();
+        rt.block_on(async {
+            tokio::spawn(async move {
+                let mut child = Command::new("go")
+                    .args([
+                        "run",
+                        "main.go",
+                        "serve",
+                        "--bucket",
```
jtguibas, line 52: I don't think the AWS bucket stuff should be handled by the Go script.

recursion/gnark-ffi/src/groth16.rs
```diff
@@ -26,9 +33,46 @@ pub struct Groth16Proof {
...
+    /// Starts up the Gnark server using Groth16 on the given port.
     pub fn new() -> Self {
-        Groth16Prover
+        let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        let gnark_dir = manifest_dir.join("../gnark");
+        let port = env::var("HOST_PORT").unwrap_or_else(|_| "8080".to_string());
+        let port_clone = port.clone();
+
+        // Start the Gnark server in a separate thread.
+        let rt = Runtime::new().unwrap();
```
jtguibas, line 44: Also, instead of tokio/rt, can we just use a simple process/thread? Async should be unecessary.

recursion/gnark-ffi/src/groth16.rs
```diff
@@ -139,7 +183,8 @@ impl Groth16Prover {
         }
     }
 
-    pub fn prove<C: Config>(witness: Witness<C>, build_dir: PathBuf) -> Groth16Proof {
+    // Proves by using the files in the data directory.
+    pub fn prove_local<C: Config>(witness: Witness<C>, build_dir: PathBuf) -> Groth16Proof {
```
jtguibas, line 187: Just leave prove. The circuit downloading shouldn't be handled here

recursion/gnark-ffi/src/groth16.rs
```diff
@@ -201,6 +246,33 @@ impl Groth16Prover {
 
         deserialized
     }
+
+    /// Proves by sending a request to the Gnark server. If the server is not ready, it will wait
+    /// until it is before sending the request.
+    pub fn prove<C: Config>(&self, witness: Witness<C>) -> Groth16Proof {
+        self.wait_for_healthy_server().unwrap();
```
jtguibas, line 253: Can we just wait in the call to new()? That way benchmarking is more accurate.

recursion/gnark/go.mod
jtguibas: no problem

recursion/gnark-ffi/src/groth16.rs
mattstam: sure although it might be nice to do other stuff while the download occurs

recursion/gnark-ffi/src/groth16.rs
mattstam: you mean exposong it as a flag?

mattstam: removed the flag

recursion/gnark-ffi/src/groth16.rs
mattstam: updated

recursion/gnark-ffi/src/groth16.rs
mattstam: removed prove_local

recursion/gnark-ffi/src/groth16.rs
mattstam: moved it over to new()

recursion/gnark/server/load.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/consensys/gnark-crypto/ecc"
+	"github.com/consensys/gnark/backend/groth16"
+	"github.com/consensys/gnark/constraint"
+	"github.com/pkg/errors"
+)
+
+// LoadCircuit downloads the R1CS and proving key and loads it into memory.
+func LoadCircuit(ctx context.Context, circuitBucket, circuitType, circuitVersion string) (constraint.ConstraintSystem, groth16.ProvingKey, error) {
```
mattstam, line 24: downloads from S3

recursion/gnark-ffi/src/groth16.rs
```diff
@@ -26,9 +33,73 @@ pub struct Groth16Proof {
...
+
+        let prover = Self { port: port_clone };
+
+        prover.wait_for_healthy_server().unwrap();
+
+        prover
+    }
+
+    /// Checks if the server is ready to accept requests.
+    fn wait_for_healthy_server(&self) -> Result<(), Box<dyn std::error::Error>> {
```
mattstam, line 80: I'm still kinda bullish doing this in prove() instead of new(), because then it's really easy for this to just be async work. For example:
```
    let wrapper_client = Groth16Prover::new();
    let sp1_prover = SP1Prover::new();
    // plus some other stuff
    // plus core proof
    // then finally, groth16 prove()
    ```

File: recursion/gnark/server/load.go [hunk 8]
```diff
@@ -0,0 +1,133 @@
...
+	}
+	defer os.Remove(r1csTempFile.Name())
+
+	r1csFileName := fmt.Sprintf("build/circuit_%s.bin", circuitType)
+	pkFileName := fmt.Sprintf("build/pk_%s.bin", circuitType)
+
+	var r1csExtracted, pkExtracted bool
+	for {
+		if r1csExtracted && pkExtracted {
+			break
```
jtguibas, line 112: You should separate out downloading the artifacts to a folder with loading the circuit artifacts from a build directory. Right now this function does both.

recursion/gnark-ffi/src/groth16.rs
ctian1: maybe start it in new but expose wait

mattstam: maybe its fine for now dont want to expose all the async

jtguibas (approved):

recursion/gnark/server/load.go
mattstam: going to leave a TODO to refactor this since Uma wants to test

PR #595, opened 2024-04-25
refactor: `get_cycles`

ratankaliani:
Move `get_cycles` utility method into `sp1-prover` for use in benchmarking.


jtguibas:
LGTM but one thing is that this function should kinda be refactored because it assumes the program has no inputs.

jtguibas (approved):

PR #621, opened 2024-04-27
feat: add groth16 example

0xturboblitz:
Add example of groth16 verification in sp1 (not the other way around) using arkworks.

Proof generation time on Macbook Air M2: 3254s.


0xturboblitz:
Thanks! Makes sense. Not planning on adding precompiles within the next few weeks, but if anyone forks [arkworks-rs/algebra](https://github.com/arkworks-rs/algebra) to add them switching here will be super easy.

PR #622, opened 2024-04-27
fix: some SubBuilder tweaks

dlubarov:
Technically `row` seemed wrong before (just passed along the whole row). Also `row_slice` can be simplified.


kevjue (approved):
LGTM!  Thanks for the tweaks.

PR #630, opened 2024-04-29
feat: add multiple proving modes to network client

mattstam:
When `SP1_PROVER=network`, the calls to `prove()`, `prove_compressed()`, `prove_plonk()`, and `prove_groth16()` will instead make the request to the prover network and get a proof back in that format.


sdk/src/proto/network.rs
```diff
@@ -89,15 +92,15 @@ pub struct ClaimProofResponse {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct FulfillProofRequest {
+    /// The signature of the message.
```
mattstam, line 95: moved signature to first param in all of these because so it is safer to append fields later

PR #634, opened 2024-04-30
feat(recursion): poseidon2 max constraint degree const generic

kevjue:


core/src/utils/mod.rs
```diff
@@ -29,13 +29,13 @@ pub const fn indices_arr<const N: usize>() -> [usize; N] {
     indices_arr
 }
 
-pub fn pad_to_power_of_two<const N: usize, T: Clone + Default>(values: &mut Vec<T>) {
-    debug_assert!(values.len() % N == 0);
-    let mut n_real_rows = values.len() / N;
+pub fn pad_to_power_of_two<T: Clone + Default>(num_columns: usize, values: &mut Vec<T>) {
```
tamirhemo, line 32: Is there a reason this needed to change? it seems like we are not using the fact it's a compile time constant, is that true?

core/src/utils/prove.rs
```diff
@@ -431,11 +431,17 @@ pub mod baby_bear_poseidon2 {
         }
     }
 
+    enum BabyBearPoseidon2Type {
```
tamirhemo, line 434: Nice fix!

recursion/core/src/poseidon2_wide/external.rs
```diff
@@ -292,48 +285,56 @@ fn eval_external_round<AB: SP1AirBuilder>(
...
-            state[i].clone(),
-        )
+        builder.assert_eq(external_state[i], state[i].clone())
     }
 }
 
 impl<F, const DEGREE: usize> BaseAir<F> for Poseidon2WideChip<DEGREE> {
     fn width(&self) -> usize {
-        NUM_POSEIDON2_WIDE_COLS
+        let use_sbox_3 = DEGREE < 7;
```
tamirhemo, line 332: I think it would be nicer to use a match statement (although the compiler might already optimize this if away)

```
match DEGREE {
    d if d < 7 => NUM_POSEIDON2_SBOX_COLS
    _ => NUM_POSEIDON2_COLS
}
```

recursion/core/src/poseidon2_wide/external.rs
```diff
@@ -381,55 +382,67 @@ where
     AB: SP1RecursionAirBuilder,
 {
     fn eval(&self, builder: &mut AB) {
+        assert!(DEGREE > 3, "Minimum supported constraint degree is 3");
+        let use_sbox_3 = DEGREE < 7;
         let main = builder.main();
         let cols = main.row_slice(0);
-        let cols: &Poseidon2WideCols<AB::Var> = (*cols).borrow();
+        let cols = if use_sbox_3 {
```
tamirhemo, line 389: same comment here, but again, compiler is probably making this ifs go away anyway

core/src/utils/mod.rs
kevjue: Actually, looks like I don't need to change it by using a const function to determine the poseidon2 width.  Will revert this.

kevjue: I actually now remember why I had to make this change.

So a const func that uses the struct's DEGREE generic wouldn't work.  The reason is that a const variable (event if declared within a function of the struct) can't access that generic.

`In Rust, const and static items are considered separate, standalone items, and thus they cannot capture or use generic parameters from their enclosing scopes directly. This limitation also applies to nested functions that attempt to use generics declared in an outer scope.`

E.g. these (illustrative) lines of code in `generate_trace` wouldn't work:

```
        const WIDTH: usize = if DEGREE > 3 { 16 } else { 4 };
        pad_to_power_of_two::<WIDTH, F>(&mut trace.values);
```

So that means that the poseidon2's chip's calculated width can't be a const, and can't be used as a const generic for `pad_to_power_of_two`.

PR #642, opened 2024-05-01
feat: `mulmod` uint256 precompile

puma314:
- [ ] Add logic in case that `modulus=0` for case of no reduction (i.e. modulus = 1 << 256)
- [ ] Fix division test


zkvm/precompiles/src/uint256_div.rs
```diff
@@ -4,12 +4,42 @@ use crate::syscall_uint256_mul;
 use crate::unconstrained;
 use num::{BigUint, Integer};
 
+const BIGINT_WIDTH_WORDS: usize = 8;
+
+#[allow(unused_variables)]
+#[no_mangle]
+pub extern "C" fn sys_bigint(
```
ctian1, line 11: pls rename and move to correct spot

zkvm/precompiles/src/uint256_div.rs
```diff
@@ -4,12 +4,42 @@ use crate::syscall_uint256_mul;
...
+    result: *mut [u32; BIGINT_WIDTH_WORDS],
+    op: u32,
```
ctian1, line 13: what is op for ?

ctian1 (approved):

PR #651, opened 2024-05-02
feat: program refactor 

tamirhemo:


prover/src/lib.rs
```diff
@@ -201,501 +232,379 @@ impl SP1Prover {
...
+                if i == 0 && j == 0 {
+                    // Truncate top 3 bits
+                    committed_values_digest =
+                        Bn254Fr::from_canonical_u32(byte.as_canonical_u32() & 0x1f);
+                } else {
+                    committed_values_digest *= Bn254Fr::from_canonical_u32(256);
+                    committed_values_digest += Bn254Fr::from_canonical_u32(byte.as_canonical_u32());
+                }
+            }
+        }
```
ctian1, line 602: this code was moved out already, should be as it is in main

prover/src/lib.rs
```diff
@@ -201,501 +232,379 @@ impl SP1Prover {
...
 
         let mut witness = Witness::default();
         proof.write(&mut witness);
-        witness.write_commited_values_digest(committed_values_digest);
-        witness.write_vkey_hash(vkey_hash);
+        witness.commited_values_digest = committed_values_digest;
+        witness.vkey_hash = vkey_hash;
```
ctian1, line 607: this should be as it was

prover/src/lib.rs
```diff
@@ -201,501 +232,379 @@ impl SP1Prover {
...
+
+                    let mut recursive_challenger = self.compress_machine.config().challenger();
+                    let mut proof = self.compress_machine.prove::<LocalProver<_, _>>(
+                        &self.compress_pk,
+                        runtime.record,
+                        &mut recursive_challenger,
+                    );
+
+                    debug_assert_eq!(proof.shard_proofs.len(), 1);
+                    (proof.shard_proofs.pop().unwrap(), ReduceProgramType::Reduce)
```
ctian1, line 450: shouldn't this be a call to verify_batch or some function

kevjue (approved):
LGTM

PR #696, opened 2024-05-10
chore: disable nightly release

jtguibas:
makes it so latest tagged release is always pinned on homepage of repo. our releases will always have working groth16 artifacts


kevjue (approved):
LGTM

PR #697, opened 2024-05-10
perf: get_cycles don't need emit events

lispc:
So it can be faster and use less memory


jtguibas (approved):

jtguibas:
Thanks for your PR. I will merge it in and it will be included in the next release.

PR #729, opened 2024-05-14
feat: add proof verification

ratankaliani:
## Overview

Adds Groth16 proof verification to the SDK.

- [x] Add `hash_public_values` to `SP1PublicValues` https://github.com/succinctlabs/sp1/pull/730
- [x] Re-add verification in Gnark for `Groth16` proofs

In a future PR, I'll add `mock` mode for `verify_groth16`.

PR #730, opened 2024-05-14
feat: add `hash_public_values`

ratankaliani:
Add `hash_public_values` to `core` to mirror the implementation of `hashPublicValues` in the Solidity verifier. This will be useful in https://github.com/succinctlabs/sp1/pull/729, where we'll add proof verification.

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -137,15 +137,16 @@ impl SP1PublicValues {
     /// return sha256(publicValues) & bytes32(uint256((1 << 253) - 1));
     /// ```
     pub fn hash_public_values(&self) -> BigUint {
+        // Hash the public values.
         let mut hasher = Sha256::new();
         hasher.update(self.buffer.data.as_slice());
         let hash_result = hasher.finalize();
-
         let mut hash = hash_result.to_vec();
 
         // Mask the top 3 bits.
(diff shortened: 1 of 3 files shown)


core/src/io.rs
```diff
@@ -127,6 +129,26 @@ impl SP1PublicValues {
     pub fn write_slice(&mut self, slice: &[u8]) {
         self.buffer.write_slice(slice);
     }
+
+    /// Hash the public values, mask the top 3 bits and return a BigUint.
+    /// Matches the implementation of hashPublicValues in the Solidity verifier.
+    ///
+    /// ```
+    /// return sha256(publicValues) & bytes32(uint256((1 << 253) - 1));
```
ratankaliani, line 137: Is it possible to add Solidity code inside of a Rust comment?

core/src/io.rs
```diff
@@ -127,6 +129,26 @@ impl SP1PublicValues {
     pub fn write_slice(&mut self, slice: &[u8]) {
         self.buffer.write_slice(slice);
     }
+
+    /// Hash the public values, mask the top 3 bits and return a BigUint.
```
jtguibas, line 133: nit: comment doens't seem to span 100 columns?

jtguibas (approved):

PR #738, opened 2024-05-15
refactor: use actual ffi for gnark

ctian1:
* Current gnark interface relies on the go code actually being available locally, so it won't work if you try to `cargo build` and use the executable somewhere else.
* This PR changes recursion/gnark-ffi to use real ffi and statically links the gnark code used to build/prove groth16.


prover/scripts/e2e.rs
```diff
@@ -77,17 +77,17 @@ pub fn main() {
...
+    let groth16_prover = Groth16Prover::new();
 
     tracing::info!("gnark prove");
-    let proof = groth16_prover.prove(witness.clone());
+    let proof = groth16_prover.prove(witness.clone(), args.build_dir.clone().into());
 
-    tracing::info!("verify gnark proof");
-    let verified = verify(proof.clone(), &args.build_dir.clone().into());
-    assert!(verified);
+    // tracing::info!("verify gnark proof");
```
jtguibas, line 85: nit: supposed to be commented?

jtguibas (approved):

PR #742, opened 2024-05-15
Invalid test case

nyunyunyunyu:


kevjue:
This test case has been incorporated in this https://github.com/succinctlabs/sp1/pull/747.

PR #745, opened 2024-05-15
feat: Implement `verify_groth16` & `prove_groth16` on `MockProver`

ratankaliani:
Add implementations for `verify_groth16` and `prove_groth16` to the MockProver.

These implementations simply check that the `public_values` and `vkey_hash` are correct.

This ensures that the following user flow is correct:
1. User generates a proof in mock mode and saves it locally.
2. User loads in the proof and calls `verify_groth16` on it in mock mode. If they are using the correct `vkey` (i.e. their program hasn't changed), then the proof will verify.

### DIFF
--- a/prover/src/verify.rs
+++ b/prover/src/verify.rs
@@ -6,15 +6,25 @@ use p3_baby_bear::BabyBear;
 use p3_field::{AbstractField, PrimeField};
 use sp1_core::{
     air::PublicValues,
+    io::SP1PublicValues,
     stark::{MachineProof, MachineVerificationError, StarkGenericConfig},
     utils::BabyBearPoseidon2,
 };
 use sp1_recursion_core::{air::RecursionPublicValues, stark::config::BabyBearPoseidon2Outer};
 use sp1_recursion_gnark_ffi::{Groth16Proof, Groth16Prover};
+use thiserror::Error;
 
 use crate::{
     CoreSC, HashableKey, OuterSC, SP1CoreProofData, SP1Prover, SP1ReduceProof, SP1VerifyingKey,
 };
+
+#[derive(Error, Debug)]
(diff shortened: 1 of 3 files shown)


prover/src/verify.rs
```diff
@@ -6,16 +6,26 @@ use p3_baby_bear::BabyBear;
...
 };
 use sp1_recursion_core::{air::RecursionPublicValues, stark::config::BabyBearPoseidon2Outer};
 use sp1_recursion_gnark_ffi::{Groth16Proof, Groth16Prover};
+use thiserror::Error;
 
 use crate::{
     CoreSC, HashableKey, OuterSC, SP1CoreProofData, SP1Prover, SP1ReduceProof, SP1VerifyingKey,
 };
 
+#[derive(Error, Debug)]
```
jtguibas, line 21: nit: iirc all error enums in the repo have errors that are lowercase with no periods

prover/src/verify.rs
```diff
@@ -218,12 +222,44 @@ impl SP1Prover {
...
-        let vk_hash = vk.hash_bn254().as_canonical_biguint();
-        if vk_hash != vkey_hash {
-            return Err(anyhow::Error::msg("vk hash mismatch"));
-        }
-
         Ok(())
     }
 }
+
+/// Verify the vk_hash and public_values_hash in the public inputs of the Groth16Proof match the expected values.
```
jtguibas, line 229: I think it might make more sense to implement this code directly on proof as a impl function on self.

Generally speaking, it seems like all of these functions are always going to be called against the object.

jtguibas (approved):

prover/src/verify.rs
ratankaliani: updated

PR #748, opened 2024-05-15
feat(recursion): public values constraints

kevjue:


jtguibas (approved):

jtguibas:
Can you write a small test to test that the public values are properly working?

PR #754, opened 2024-05-17
Use 24-element Poseidon2 for leaf hashing

nhukc:
Use 24-element Poseidon2 for leaf hashing. This yields a 10-second speedup in overall proof time on my machine.


jtguibas:
Fascinating! I'm curious if it will slow down recursion at all, given we'll have to have more columns to express the 24-wide permutation.

jtguibas:
Closing this for now because the chances we'll merge this are low due to it affecting recursion too much.

PR #763, opened 2024-05-20
fix(core): Fix benches warning for unstable features

Nashtare:
Compiling with `cargo build --release --all-targets` would otherwise fail with the following: `error: failed to evaluate generic const expression`


jtguibas:
Thank you! Will merge this into dev and we'll see it on main in the next release.

jtguibas (approved):

PR #786, opened 2024-05-22
feat: Implement an optimized SP1 Serializer

### LINKED ISSUE #724: Implement an optimized SP1 Serializer
Right now, when a user does `sp1_zkvm::io::read::<T>(...)`, underneath the hood we use `bincode` as our serializer to read from the buffer. Reference is [here](https://github.com/succinctlabs/sp1/blob/2f57e1e77f7c88396b571bd961d3cbe4b1291c7a/core/src/io.rs#L48).

However using `bincode` as a serializer is inefficient, since we don't want any compression and we don't want to pay zkVM cycles for reading in data. 

We should implement our own optimized zkVM serializer that just serializes/deserializes data ideally in `u32` aligned chunks.

malik672:
implemented a custom zkVm serializer

This PR implements a custom optimised zkVm serialised

Closes #724

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -107,13 +107,11 @@ impl SP1Stdin {
         let data_slice = &self.buffer[self.ptr];
 
         // Deserialize from the slice
-        let deserialized_data = serde_json::from_slice::<T>(data_slice)
-            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e));
+        serde_json::from_slice::<T>(data_slice)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))
 
         // Assuming you want to update `self.ptr` here to point to the next element, if needed
         // self.ptr += 1; // or adjust based on the actual size of T, if known
-
-        deserialized_data
     }
 
     /// Read a slice of bytes from the buffer.


matthiasgoergens (commented):
You might want to directly link to https://github.com/succinctlabs/sp1/issues/724 in your PR description, so that github also knows that you are talking about the issue.  And you might want to change your PR subject line to something informative like 'Implement an optimized SP1 Serializer' (plus the right prefix to make the PR linter happy).

malik672:
> You might want to directly link to #724 in your PR description, so that github also knows that you are talking about the issue. And you might want to change your PR subject line to something informative like 'Implement an optimized SP1 Serializer' (plus the right prefix to make the PR linter happy).

what's the right prefix

matthiasgoergens:
I'm not a maintainer.

I think you can find details about the checks in the `.github` folder. There should be something along the lines of PR lint or so. (Sorry, am on mobile, so can't check properly right now.)

core/src/io.rs
```diff
@@ -44,12 +91,27 @@ impl SP1Stdin {
...
+
+        // Correct access to the byte slice
+        let data_slice = &self.buffer[self.ptr];
+
+        // Deserialize from the slice
+        serde_json::from_slice::<T>(data_slice)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))
+
+        // Assuming you want to update `self.ptr` here to point to the next element, if needed
+        // self.ptr += 1; // or adjust based on the actual size of T, if known
```
drcapybara, line 114: looks like you might have left some chatgpt behind. Youll definitely want to increment that counter, otherwise you'll just be reading the same bytes every time you call read

drcapybara:
Just a thought but this PR kind of misses the point of the original issue that @puma314 was describing. The problem is that ```bincode``` consumes a substantial amount of cycles to perform deserialization. We shouldn't have to pay for these cycles in the zkVM because they add up extremely quickly and result in high proof time. 

```serde_json``` does nothing to help with this problem and in some cases might actually consume more cycles than ```bincode```. You should look into the use of zero-copy deserialization crates like ```rkyv``` which will perform safe zero-copy deserialization at a significantly lower cost than ```bincode```, or if you're feeling particularly daring, you can use pointer casts to deserialize embedded bytes in the guest for practically nothing.

matthiasgoergens:
Yes, rkyv is great.

You want to be very careful with pointer casting, though. You don't want a malicious prover to be able to subvert your program.

core/src/io.rs
malik672: nothing bad in using gpt though, just a tool to help

PR #793, opened 2024-05-23
chore: make linter and clippy happy

matthiasgoergens:
Via repeated application of
```bash
cargo fix --allow-dirty
cargo clippy --allow-dirty --fix --all -- -D warnings -D clippy::pedantic
cargo fmt
```
until we reach a fixed point.


matthiasgoergens:
I suggest reviewing this via re-running the same commands against `dev` and checking if you get the same result.  (Just to check that there's no funny business going on.)

jtguibas (approved):

jtguibas (approved):

jtguibas:
It seems like CI fails for some reason. I will investigate.

matthiasgoergens:
Looks like your clippy version has more suggestions.

jtguibas:
Closing this for now as the branch has gotten a bit out of sync.

PR #815, opened 2024-05-27
feat: sp1 core opts

jtguibas:


prover/src/lib.rs
```diff
@@ -578,16 +568,15 @@ impl SP1Prover {
         tracing::debug!("Compress program executed successfully");
 
         // Prove the compress program.
+        let opts = SP1CoreOpts::default();
```
ctian1: this should be recursion

ctian1 (approved):

PR #831, opened 2024-05-29
feat: execution record of program statistics

drcapybara:
Currently, when a RISCV program executes, we do not store detailed logs of the execution separate from the actual trace. When running a program within the provable setting, every cycle counts, and we want to be able to have a detailed view inside of our program that allows us to understand where compute resources are being used.

This PR introduces an execution record, which contains detailed information about the types of RISCV instructions and opcodes that are carried out during a programs run. It can be used in the following way:

```rust
use sp1_core::runtime::{Instruction, Opcode, Program, Register, Runtime, SP1CoreOpts};

let instructions = vec![
    Instruction::new(Opcode::ADD, 29, 0, 5, false, true),
    Instruction::new(Opcode::ADD, 30, 0, 37, false, true),
    Instruction::new(Opcode::ADD, 31, 30, 29, false, false),
];

// Setup a program and a runtime
let program = Program::new(instructions, 0, 0);
let mut runtime = Runtime::new(program, SP1CoreOpts::default());
// Runtime must execute before reporting statistics
let res = runtime.dry_run().unwrap();
assert_eq!(runtime.register(Register::X31), 42);

// Retrieve the number of ADD operations
let add_count = &res.get_metric("Arithmetic", "ADD").unwrap_or(0);
assert_eq!(*add_count, 3);
assert_eq!(res.cycles, 12);
assert_eq!(res.total_instruction_count, 3);

// You can also get a count of all instruction/opcode occurences by category
let add_operations_count = res.get_total_for_category("Arithmetic").unwrap_or(0);
assert_eq!(add_operations_count, 3,);
```

This will allow for cheap and easy program profiling to help understand the costs of running a program within the proof pipeline. These metrics can be retrieved from running formally and from executing a dry run

### DIFF
--- a/core/src/runtime/execution_report.rs
+++ b/core/src/runtime/execution_report.rs
@@ -1,0 +1,170 @@
+use super::Runtime;
+use std::{any::Any, collections::HashMap};
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+/// Errors which may occur when querying an execution report for
+/// instruction details.
+pub enum MetricRetrievalError {
+    #[error("category not found")]
+    CategoryNotFound,
+
+    #[error("opcode not found in category")]
+    OpcodeNotFound,
+
+    #[error("data type mismatch for the stored metric")]
+    DataTypeMismatch,
+}
+
+/// A report containing useful information about the execution of a program.
+/// It stores useful information about the resource footprint of a provable
+/// RISCV program, including number of cycles, instruction count, and syscall
+/// count.
+///
+/// ### Custom Metrics:
+/// An execution report contains a rich set of metrics and granular details
+/// that can be easily accessed. You can, for example, count the total
+/// number of ADD instructions that occurred during a program run:
+/// ```
+/// use sp1_core::runtime::{Instruction, Opcode, Program, Register, Runtime, SP1CoreOpts};
+///
+/// // Simple program to test the ADD instruction
+/// let instructions = vec![
+///     Instruction::new(Opcode::ADD, 29, 0, 5, false, true),
+///     Instruction::new(Opcode::ADD, 30, 0, 37, false, true),
+///     Instruction::new(Opcode::ADD, 31, 30, 29, false, false),
+/// ];
+///
+/// // Setup a program and a runtime
+/// let program = Program::new(instructions, 0, 0);
+/// let mut runtime = Runtime::new(program, SP1CoreOpts::default());
+/// // Runtime must execute to produce runtime statistics
+/// let mut res = runtime.run().unwrap();
+///
+/// // Retrieve the number of ADD operations
+/// let add_count = &res.get_metric("Arithmetic", "ADD").unwrap_or(0);
+/// assert_eq!(*add_count, 3);
+///
+/// // You can also get a count of all opcode occurences by category
+/// let memory_operations_count = res.get_total_for_category("Arithmetic").unwrap_or(0);
+/// assert_eq!(memory_operations_count, 3,);
+/// ```
+#[derive(Debug)]
+pub struct ExecutionReport {
+    /// Total number of cycles occurred during program execution
+    pub cycles: u32,
+    /// Total number of instructions occurred during program execution
+    pub total_instruction_count: usize,
+    /// Total number of syscalls occurred during program execution
+    pub syscall_count: usize,
+    /// Contains precise counts of each individual instruction,
+    /// as well as counts of instruction by category. See [ExecutionReport] Usage for
+    /// more details
+    pub custom_metrics: HashMap<String, Box<dyn Any>>,
+}
+
+impl ExecutionReport {
+    /// Accept a runtime from a program which has already executed
+    /// and extract some useful information from it.
+    pub fn new(runtime: &Runtime) -> Self {
+        let total_instruction_count = runtime
+            .instruction_log
+            .values()
+            .map(|counts| counts.values().sum::<u64>() as usize)
+            .sum();
+
+        let syscall_count = runtime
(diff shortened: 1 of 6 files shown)


core/src/runtime/mod.rs
```diff
@@ -529,6 +536,29 @@ impl Runtime {
 
         let mut next_pc = self.state.pc.wrapping_add(4);
 
+        // Determine the category based on opcode for logging purposes
+        let category = match instruction.opcode {
```
jtguibas, line 540: i don't really like this, feels not very extensible. instead, can you just keep track of Opcode => counts of how many time they appear in a HashMap?

all analysis like Arithmetic , Syscalls, etc can be done post hoc on the raw data in the execution report.

core/src/runtime/execution_report.rs
```diff
@@ -0,0 +1,171 @@
...
+/// let add_operations_count = report.get_total_for_category("Arithmetic").unwrap_or(0);
+/// assert_eq!(add_operations_count, 3,);
+/// ```
+#[derive(Debug)]
+pub struct ExecutionReport {
+    /// Total number of cycles occurred during program execution
+    pub cycles: u32,
+    /// Total number of instructions occurred during program execution
+    pub total_instruction_count: usize,
+    /// Total number of syscalls occurred during program execution
```
jtguibas, line 64: these two statistics are not that useful, would be better to have HashMap<Opcode, usize> and HashMap<SyscallCode, usize>

core/src/runtime/execution_report.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub syscall_count: usize,
+    /// Contains precise counts of each individual instruction,
+    /// as well as counts of instruction by category. See [ExecutionReport] Usage for
+    /// more details
+    pub custom_metrics: HashMap<String, Box<dyn Any>>,
```
jtguibas, line 69: don't need to support custom metrics for now

core/src/runtime/execution_report.rs
```diff
@@ -0,0 +1,171 @@
...
+        Self {
+            cycles: runtime.state.clk,
+            total_instruction_count,
+            syscall_count,
+            custom_metrics,
+        }
+    }
+
+    /// Retrieve specific metric based on category and opcode.
+    pub fn get_metric(&self, category: &str, opcode: &str) -> Result<u64, MetricRetrievalError> {
```
jtguibas, line 106: these methods are unecessary, can just let users directly access fields on struct for now

jtguibas: don't want to ensrhine categories inside of Runtime

jtguibas (commented):
overall looks clean!

prover/src/utils.rs
```diff
@@ -31,7 +31,7 @@ pub fn get_cycles(elf: &[u8], stdin: &SP1Stdin) -> u64 {
     let program = Program::from(elf);
     let mut runtime = Runtime::new(program, SP1CoreOpts::default());
     runtime.write_vecs(&stdin.buffer);
-    runtime.dry_run();
+    let _ = runtime.dry_run();
```
jtguibas, line 34: unwrap?

core/src/runtime/execution_report.rs
```diff
@@ -0,0 +1,171 @@
...
+    CategoryNotFound,
+
+    #[error("opcode not found in category")]
+    OpcodeNotFound,
+
+    #[error("data type mismatch for the stored metric")]
+    DataTypeMismatch,
+}
+
+/// A report containing useful information about the execution of a program.
```
jtguibas, line 20: great comment

core/src/runtime/execution_report.rs
```diff
@@ -0,0 +1,171 @@
+use super::Runtime;
+use std::{any::Any, collections::HashMap};
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
```
jtguibas, line 6: a bit odd to put the error above the actual struct we care about in this file

prover/src/lib.rs
```diff
@@ -213,16 +214,23 @@ impl SP1Prover {
 
     /// Generate a proof of an SP1 program with the specified inputs.
     #[instrument(name = "execute", level = "info", skip_all)]
-    pub fn execute(elf: &[u8], stdin: &SP1Stdin) -> Result<SP1PublicValues, ExecutionError> {
+    pub fn execute(
+        elf: &[u8],
+        stdin: &SP1Stdin,
+    ) -> Result<(SP1PublicValues, ExecutionReport), ExecutionError> {
```
jtguibas, line 220: instead of a tuple, maybe better to make a struct like ExecutionResult, so everything is more typed and extensible.

core/src/runtime/mod.rs
```diff
@@ -538,65 +549,78 @@ impl Runtime {
         match instruction.opcode {
             // Arithmetic instructions.
```
drcapybara, line 550: We can increment a counter a single time with the opcode here but Im not sure how we would account for the syscalls

prover/src/lib.rs
```diff
@@ -79,6 +80,14 @@ pub type ReduceAir<F> = RecursionAir<F, REDUCE_DEGREE>;
 pub type CompressAir<F> = RecursionAir<F, COMPRESS_DEGREE>;
 pub type WrapAir<F> = RecursionAir<F, WRAP_DEGREE>;
 
+/// Contains result of execution along with an [ExecutionReport].
```
jtguibas, line 83: nit: a bit weird to put this struct above `SP1Prover`

can you put it in `core`?

PR #839, opened 2024-05-30
feat: runtime instruction/syscall report

tqn:
Counts instructions and syscalls by their opcode/syscall code when `should_report` is set (currently always).


mattstam:
looks good, but please merge into [succinctlabs:dev](https://github.com/succinctlabs/sp1/tree/dev) branch instead of main

mattstam:
merging although FYI u64 is more appropriate for cycle counts as we already have some programs close to u32 cap, will fix in a follow up PR

PR #851, opened 2024-05-31
feat: docker process to use the go gnark library

tqn:
I set up the Dockerfile to wrap the crate in question (`recursion/gnark-ffi`). Use the commands `docker build -t sp1-recursion-gnark-ffi -f ./gnark.Dockerfile .` and `docker run --rm sp1-recursion-gnark-ffi asdf` to verify it works.

The `container.rs` file is intended to contain helper functions to pull (or perhaps build) the image and run the image with arguments, reading the stdout. The `main.rs` file will parse these arguments, perhaps with `clap`, and call the functions in `ffi.rs`.

The `Dockerfile` has many commented-out lines of code that should be deleted.


tqn:
Feature being implemented in ﻿#890.

PR #889, opened 2024-06-05
fix: set core opts from SHARD* env vars

chiefbiiko:
Reads runtime env vars `SHARD_SIZE` and `SHARD_BATCH_SIZE` into `SP1CoreOpts::default()`. It seems they just weren't applied before.


chiefbiiko:
For some reason though, increasing the `SHARD*` envs doesn't yield any performance improvements for my program..

jtguibas (approved):

PR #899, opened 2024-06-07
feat: sp1 circuit version

ctian1:
* Have global circuit version string rather than plonk commit
  * This is easier to track than commit as it directly maps to an sp1-sdk release

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -34,7 +34,9 @@ pub mod utils;
 use runtime::{Program, Runtime};
 use stark::StarkGenericConfig;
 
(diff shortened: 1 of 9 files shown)


core/src/lib.rs
```diff
@@ -33,3 +33,8 @@ pub mod utils;
 #[allow(unused_imports)]
 use runtime::{Program, Runtime};
 use stark::StarkGenericConfig;
+
+/// The global version for all components of SP1. This string should be updated whenever any step in
```
jtguibas, line 37: nit:

```
/// The global version for all components of SP1.
///
/// <rest of text>
```

jtguibas (approved):

PR #903, opened 2024-06-09
chore: logging cleanup + non-compressed fibonacci

ctian1:


examples/fibonacci/script/src/main.rs
```diff
@@ -16,7 +16,7 @@ fn main() {
     // Generate the proof for the given program and input.
     let client = ProverClient::new();
     let (pk, vk) = client.setup(ELF);
-    let mut proof = client.prove_compressed(&pk, stdin).unwrap();
+    let mut proof = client.prove(&pk, stdin).unwrap();
```
mattstam, line 19: btw theres a few other examples with default compressed to change back

mattstam (approved):

examples/fibonacci/script/src/main.rs
mattstam: maybe just do all of the ones in here https://github.com/succinctlabs/sp1/pull/891 or just `git cherry-pick 7085d50`

PR #905, opened 2024-06-10
refactor: refactor `compress`

TomTaehoonKim:
This PR refactors `compress` by using `compress_machine_proof` instead of `compress_machine.prove`.


ctian1:
rebased into #928

PR #912, opened 2024-06-11
feat: runtime hooks

tqn:
Implement runtime hooks, which allows offloading expensive computation to the runtime.

Currently, the system is not very extensible, and is generally somewhat basic.


core/src/runtime/hooks.rs
```diff
@@ -0,0 +1,60 @@
...
+                tracing::info!("hello world! {args:?}");
+                vec![]
+            }),
+        ),
+        ("ecrecover".to_owned(), Box::new(hook_ecrecover)),
+    ];
+    HashMap::from_iter(entries)
+}
+
+pub fn hook_ecrecover(buf: &[u8]) -> Vec<Vec<u8>> {
```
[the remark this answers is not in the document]
tqn: No, the outer `Vec` is of size 2 here. It is extended onto the `input_stream` to simulate two values being pushed onto it.

core/src/syscall/write.rs
```diff
@@ -73,6 +73,17 @@ impl Syscall for SyscallWrite {
                 rt.state.public_values_stream.extend_from_slice(slice);
             } else if fd == 4 {
                 rt.state.input_stream.push(slice.to_vec());
+            } else if fd == 5 {
```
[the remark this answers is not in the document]
tqn: I have now implemented this.

zkvm/precompiles/src/io.rs
```diff
@@ -80,3 +81,20 @@ pub fn hint_slice(buf: &[u8]) {
     let mut my_reader = SyscallWriter { fd: FD_HINT };
     my_reader.write_all(buf).unwrap();
 }
+
+pub fn invoke_hook<T: Serialize>(name: &str, value: &T) {
+    let mut my_writer = SyscallWriter {
+        fd: FD_RUNTIME_HOOK,
+    };
+    // Serialize twice, so we can deserialize the name, allowing any value for the rest of the data.
```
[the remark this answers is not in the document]
tqn: There is no more serialization.

core/src/runtime/hooks.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use k256::ecdsa::{RecoveryId, Signature, VerifyingKey};
+use k256::elliptic_curve::ops::Invert;
+
+use super::Runtime;
+
+type HookId = u32;
+type BoxedHook<'a> = Box<dyn Fn(HookEnv, &[u8]) -> Vec<Vec<u8>> + 'a>;
+
+// Ensure this value is synced with the values in `zkvm/precompiles/src/io.rs`
```
jtguibas, line 12: nit: comment is weird, inconsistent use of // and ///

also it should be more like

```
/// The file descriptor through which to access `hook_ecrecover`.
///
/// Note: Ensure this value is synced with the values in `zkvm/precompiles/src/io.rs`.
```

core/src/runtime/hooks.rs
```diff
@@ -0,0 +1,100 @@
+use core::convert::TryInto;
+use std::collections::HashMap;
+
+use k256::ecdsa::{RecoveryId, Signature, VerifyingKey};
+use k256::elliptic_curve::ops::Invert;
+
+use super::Runtime;
+
+type HookId = u32;
```
jtguibas, line 9: you should try to avoid alias types unless you have very complex types, just use u32

jtguibas (approved):

core/src/runtime/hooks.rs
ctian1: we might be able to use `static_assertions` crate to ensure they are the same

core/src/runtime/hooks.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+pub fn hook_ecrecover(_env: HookEnv, buf: &[u8]) -> Vec<Vec<u8>> {
+    let (sig, msg_hash) = buf.split_at(65);
+    let sig: &[u8; 65] = sig
+        .try_into()
+        .expect("buf should have correct size and offset of bytes for sig");
+    let msg_hash: &[u8; 32] = msg_hash
+        .try_into()
+        .expect("buf should have correct size and offset of bytes for msg_hash");
```
ctian1, line 67: nit: this could be a bit cleaner if you just check buf.len() and said something like "expected ecrecover input to have len 97"

core/src/syscall/write.rs
```diff
@@ -16,7 +16,12 @@ impl Syscall for SyscallWrite {
         let a2 = Register::X12;
         let rt = &mut ctx.rt;
         let fd = arg1;
-        if fd == 1 || fd == 2 || fd == 3 || fd == 4 {
+        let mut maybe_hook = None;
+        if fd == 1 || fd == 2 || fd == 3 || fd == 4 || {
```
ctian1, line 20: nit: it's a bit unusual to have a block in the if clause. Also do we even need this if statement?

core/src/runtime/hooks.rs
tqn: I instead wrote a test in `hooks.rs` to check for equality.

core/src/runtime/hooks.rs
```diff
@@ -58,13 +58,14 @@ pub struct HookEnv<'a, 'b: 'a> {
 }
 
 pub fn hook_ecrecover(_env: HookEnv, buf: &[u8]) -> Vec<Vec<u8>> {
+    assert_eq!(
+        buf.len(),
+        65 + 33,
```
ctian1, line 63: should be 65 + 32? this should be caught if we have ecrecover test?

core/src/syscall/write.rs
```diff
@@ -16,73 +16,64 @@ impl Syscall for SyscallWrite {
...
+                    .into_iter()
+                    .for_each(|line| println!("stderr: {}", line));
+            }
+        } else if fd == 3 {
+            rt.state.public_values_stream.extend_from_slice(slice);
+        } else if fd == 4 {
+            rt.state.input_stream.push(slice.to_vec());
+        } else if let Some(hook) = rt.hook_registry.table.get(&fd) {
+            rt.state.input_stream.extend(hook(rt.hook_env(), slice));
         }
```
ctian1, line 77: if it's an unknown fd that is pretty bad, we should error / tell the user

ctian1 (changes requested):

core/src/runtime/hooks.rs
tqn: fixed and added test

ctian1 (approved):

zkvm/precompiles/src/secp256k1.rs
```diff
@@ -190,22 +190,11 @@ fn double_and_add_base(
 /// use `ecrecover`.
 pub fn unconstrained_ecrecover(sig: &[u8; 65], msg_hash: &[u8; 32]) -> ([u8; 33], Scalar) {
     unconstrained! {
```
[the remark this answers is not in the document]
tqn: I put it there because I reasoned that the computations are not directly used by the output, so there is no reason to have the VM constrain these values. @ctian1 said it was okay for now, although there is some overhead with `unconstrained!`. We could benchmark it to determine whether or not we should remove it.

core/src/runtime/hooks.rs
```diff
@@ -0,0 +1,119 @@
...
+use k256::ecdsa::{RecoveryId, Signature, VerifyingKey};
+use k256::elliptic_curve::ops::Invert;
+
+use super::Runtime;
+
+type BoxedHook<'a> = Box<dyn Fn(HookEnv, &[u8]) -> Vec<Vec<u8>> + 'a>;
+
+/// The file descriptor through which to access `hook_ecrecover`.
+pub const FD_ECRECOVER_HOOK: u32 = 5;
+// Note: To ensure any `fd` value is synced with `zkvm/precompiles/src/io.rs`,
```
jtguibas, line 13: why is this a // comment, use "///" and put it above

tqn: This comment does not have to do with `FD_ECRECOVER_HOOK` specifically, and would confuse people who are merely using that constant. I put it where other people may eventually add more constants so that they read it first.

zkvm/precompiles/src/secp256k1.rs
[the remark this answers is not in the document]
tqn: I just checked: the cycle count slightly increases without the `unconstrained!` block, from 69957 to 70110.

core/src/runtime/hooks.rs
tqn: Moved it

PR #918, opened 2024-06-11
chore: failure on sp1 on unsupported target

jtguibas:


cli/src/lib.rs
```diff
@@ -37,6 +37,20 @@ pub async fn url_exists(client: &Client, url: &str) -> bool {
...
 
+#[allow(unreachable_code)]
+pub fn is_supported_target() -> bool {
+    #[cfg(all(target_arch = "x86_64", target_os = "linux"))]
+    return true;
+
+    #[cfg(all(target_arch = "aarch64", target_os = "linux"))]
+    return true;
+
+    #[cfg(all(target_arch = "aarch64", target_os = "macos"))]
```
ctian1, line 48: x86 mac also should work

ctian1 (changes requested):

ctian1 (approved):

PR #919, opened 2024-06-11
chore: fix sys rand

jtguibas:


zkvm/entrypoint/src/lib.rs
```diff
@@ -98,16 +98,14 @@ mod zkvm {
     static GETRANDOM_WARNING_ONCE: std::sync::Once = std::sync::Once::new();
 
     fn zkvm_getrandom(s: &mut [u8]) -> Result<(), Error> {
-        use rand::Rng;
-        use rand::SeedableRng;
-
         GETRANDOM_WARNING_ONCE.call_once(|| {
```
ctian1, line 101: maybe we should move this warning into sys_rand?

ctian1 (approved):

PR #920, opened 2024-06-11
chore(book): document `cargo prove build --docker`

ctian1:


jtguibas (approved):

book/getting-started/install.md
```diff
@@ -18,16 +18,18 @@ sp1up is the SP1 toolchain installer. Open your terminal and run the following c
...
 
 ```bash
 sp1up
 ```
 
-This will install support for the `riscv32im-succinct-zkvm-elf` compilation target within your Rust compiler
-and a `cargo prove` CLI tool that will let you compile provable programs and then prove their correctness.
+This will install two things:
+
+1. The `succinct` Rust toolchain which has support for the `riscv32im-succinct-zkvm-elf` compilation target
```
jtguibas, line 31: period at end of 1

PR #923, opened 2024-06-12
chore: add version to sdk

jtguibas:


sdk/src/utils.rs
```diff
@@ -0,0 +1,9 @@
+pub use sp1_core::utils::setup_logger;
+use sp1_core::SP1_CIRCUIT_VERSION;
+
+/// Gets the current version of SP1 zkVM.
+///
+/// Note: this is not the same as the version of the SP1 SDK.
+pub fn version() -> String {
```
ctian1, line 7: should we just re-export? if not, should we name this `circuit_version` and put it in ProverClient

ctian1 (approved):

PR #951, opened 2024-06-17
fix: set sponge state to be zero

erabinov:


recursion/program/src/challenger.rs
```diff
@@ -49,13 +49,31 @@ pub struct DuplexChallengerVariable<C: Config> {
...
-        }
+        };
+
+        // Assert that the lengths of the arrays are equal to the permutation width.
+        builder.assert_usize_eq(PERMUTATION_WIDTH, result.sponge_state.len());
+        builder.assert_usize_eq(PERMUTATION_WIDTH, result.input_buffer.len());
+        builder.assert_usize_eq(PERMUTATION_WIDTH, result.output_buffer.len());
+
+        // Constrain the state of the challenger to contain all zeroes.
+        builder.assert_var_eq(result.nb_inputs, C::N::zero());
```
tamirhemo, line 66: Actually here it might be more efficient and correct to assign the value zero instead of asserting. Basically, here we are allocating an array but we did not initialize values. So you could do e.g.

builder.set(&result.sponge_state, I, C::F::zero())

tamirhemo (approved):
LGTM

PR #956, opened 2024-06-18
feat: optimize byte trace generation

jtguibas:


core/src/bytes/mod.rs
```diff
@@ -34,19 +33,10 @@ pub const NUM_BYTE_LOOKUP_CHANNELS: u32 = 16;
...
-    /// the index of the result in the array of multiplicities.
-    pub fn trace_and_map(
-        shard: u32,
-    ) -> (RowMajorMatrix<F>, HashMap<ByteLookupEvent, (usize, usize)>) {
-        // A map from a byte lookup to its corresponding row in the table and index in the array of
-        // multiplicities.
-        let mut event_map = HashMap::new();
-
+    /// This function returns a `trace` which is a matrix containing all possible byte operations.
+    pub fn trace(shard: u32) -> RowMajorMatrix<F> {
```
ctian1, line 39: do we even need shard? if it's always 0? I'm kind of confused actually. Should this function even be a compile time constant?

edit: I see there's a comment below about making it constant

ctian1 (approved):

PR #957, opened 2024-06-18
fix: v1.0.8-testnet

ctian1:
* Change runtime to emit MemoryInit only at the end of execution; this way, when committing/proving in batches, it's only emitted in the last batch
* Move MemoryProgram and MemoryInit to be in last shard only
* Change rust / recursion verifiers to handle ^
* Add num_core_shards to recursion public values, verified 1) to be same across proofs in `compress` and 2) in `assert_complete` to `== next_shard - 1`


recursion/core/src/air/public_values.rs
```diff
@@ -80,7 +80,7 @@ pub struct RecursionPublicValues<T> {
     /// First shard being proven.
     pub start_shard: T,
 
-    /// Next shard that should be proven, or 0 if the program halted.
+    /// Next shard that should be proven, if there are more.
     pub next_shard: T,
```
tamirhemo, line 84: Do we ever use this?

tamirhemo (approved):
LGTM

tamirhemo (approved):
LGTM

PR #965, opened 2024-06-21
fix: use correct value for blowup

zlangley:
This change doesn’t seem to impact the logic as `blowup` is used solely to calculate matrix heights through the expression `mat.domain.size() * blowup`, and these computed heights are then only used in equality comparisons against each other. Any nonzero value for blowup should therefore "work". (But this does make me wonder why both `batch_heights_log2` and `batch_dims` are computed, since the former should also work for this kind of logic.)


jtguibas:
Ah nice catch! Will fix & merge after running CI.

jtguibas (approved):

PR #973, opened 2024-06-24
feat: put hook results in front of the input stream

tqn:
Hook results must be accessed immediately after calling.

One no longer has to rely on the `input_stream` being empty when calling a hook and accessing its results.


ctian1 (approved):
nit: could use Vec::insert

PR #988, opened 2024-06-28
ci: fix cli workflow

mattstam:


cli/src/commands/new.rs
```diff
@@ -23,6 +23,8 @@ impl NewCmd {
         // Clone the repository.
         let output = Command::new("git")
             .arg("clone")
+            .arg("-b")
```
matthiasgoergens, line 26: In general, in script you should pick long options, like `--branch` in this case.  That helps with readability, and if you make a typo, it's more likely that you get a proper error message, instead of silently doing something else entirely.

PR #997, opened 2024-07-01
fix: multi-builder first/last row issue

erabinov:


recursion/core/src/multi/mod.rs
```diff
@@ -218,6 +254,8 @@ where
         let mut sub_builder = MultiBuilder::new(
             builder,
             local_multi_cols.is_poseidon2.into(),
+            local_multi_cols.poseidon2_first_row.into(),
+            builder.is_last_row(),
```
kevjue, line 258: Should we also add the condition `local_multi_cols.is_poseidon` (to handle the case when there are no poseidon2 rows).

recursion/core/src/multi/mod.rs
```diff
@@ -189,6 +223,8 @@ where
         let mut sub_builder = MultiBuilder::new(
             builder,
             local_multi_cols.is_fri_fold.into(),
+            builder.is_first_row(),
```
kevjue, line 226: Should we also add the condition local_multi_cols.is_fri (to handle the case when there are no fri rows).

jtguibas (approved):

PR #998, opened 2024-07-01
feat: make `AddressValue.value` a `Block`

tqn:
The base field ALU (currently the only ALU) uses `AddressValueBase` to avoid extra columns and checks. Some ugly abstractions remain and need to be fixed.


erabinov:
LGTM.

erabinov (approved):

PR #1019, opened 2024-07-02
feat: suggest prover network if high cycles

mattstam:
Example output:
```
2024-07-02T21:15:10.518487Z  INFO prove_core: summary: cycles=11106, e2e=2.716364875, khz=4088.55, proofSize=1419784
2024-07-02T21:15:10.523268Z  WARN prove_core: high cycle count, consider using the prover network for proof generation: https://docs.succinct.xyz/prover-network/setup.html
2024-07-02T21:15:10.523282Z  INFO prove_core: close time.busy=2.72s time.idle=27.5µs
```

### DIFF
--- a/prover/src/lib.rs
+++ b/prover/src/lib.rs
@@ -255,8 +255,9 @@ impl SP1Prover {
             .get_or_insert_with(|| Arc::new(self));
         let config = CoreSC::default();
         let program = Program::from(&pk.elf);
-        let (proof, public_values_stream) =
+        let (proof, public_values_stream, cycles) =
             sp1_core::utils::prove_with_context(program, stdin, config, opts.core_opts, context)?;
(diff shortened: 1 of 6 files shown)


prover/src/lib.rs
```diff
@@ -678,6 +679,12 @@ impl SP1Prover {
         }
         digest
     }
+
+    fn check_for_high_cycles(cycles: u64) {
+        if cycles > 100_000_000 {
```
mattstam, line 684: open to suggestions on this number

mattstam, line 684: one consideration is that this could get annoying for repeated use, so we might want a env var like `DISABLE_CYCLE_COUNT_WARNING=true` to skip

jtguibas (approved):

PR #1031, opened 2024-07-03
feat(cli): use GH token during installation to avoid rate limiting

mattstam:
During `sp1up` installation, you can run into this:
```
Successfully cleaned up ~/.sp1 directory.
Successfully created ~/.sp1 directory.
thread 'main' panicked at cli/src/lib.rs:72:41:
Failed to download Succinct toolchain. Likely caused by GitHub rate limiting. Please try again.
```

Now you can use `sp1up --token $GITHUB_TOKEN` to avoid the rate-limiting.

### DIFF
--- a/book/getting-started/install.md
+++ b/book/getting-started/install.md
@@ -42,7 +42,9 @@ If this works, go to the [next section](./quickstart.md) to compile and prove a 
 
 ### Troubleshooting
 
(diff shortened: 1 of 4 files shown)


mattstam:
> We should add this to the docs, and then we should also add this to the CLI with a link to the docs: "Likely caused by GitHub rate limiting. Please try again, docs: ....".

added in https://github.com/succinctlabs/sp1/pull/1031/commits/0636148908baafaf2efbc715e4b4a8914dc4efc0

jtguibas (approved):

PR #1034, opened 2024-07-04
feat: add memory error

jtguibas:


sdk/src/provers/local.rs
```diff
@@ -65,6 +66,12 @@ impl Prover for LocalProver {
         let compress_proof = self.prover.shrink(reduce_proof, opts)?;
         let outer_proof = self.prover.wrap_bn254(compress_proof, opts)?;
 
+        let total_ram_gb = System::new_all().total_memory() / 1_000_000_000;
+        if total_ram_gb <= 120 {
```
jtguibas, line 70: did 120 not 128 intentionally here, don't want weird edge cases

sdk/src/provers/local.rs
```diff
@@ -65,6 +66,12 @@ impl Prover for LocalProver {
         let compress_proof = self.prover.shrink(reduce_proof, opts)?;
         let outer_proof = self.prover.wrap_bn254(compress_proof, opts)?;
 
+        let total_ram_gb = System::new_all().total_memory() / 1_000_000_000;
```
ctian1, line 69: could be nice to move the check earlier in the function so users can see faster

ctian1 (approved):

PR #1075, opened 2024-07-10
feat: Fp precompile

0xWOLAND:
Currently, bls12-381 implements CPU-optimized Fp multiplication. However, this is quite slow inside of SP1. Instead, we should use a precompile to accelerate Fp-type multiplication. The corresponding pr in `succinct/bls12-381` is https://github.com/succinctlabs/bls12_381/pull/4


core/src/syscall/precompiles/fp384/fp12.rs
```diff
@@ -0,0 +1,1167 @@
...
+
+use super::Fp12;
+
+/// The number of columns in the FpMulCols.
+const NUM_COLS: usize = size_of::<Fp12MulCols<u8>>();
+
+trait Fp12Builder<F> {
+    type DType: Clone;
+
+    fn _add(
```
tamirhemo, line 40: nit: this function is already private, no need to have _add

core/src/syscall/precompiles/fp384/fp12.rs
```diff
@@ -0,0 +1,1167 @@
...
+        &mut self,
+        dest: &mut FieldOpCols<F, U384Field>,
+        a: &Self::DType,
+        b: &Self::DType,
+    ) -> Self::DType;
+    fn _mul(
```
tamirhemo, line 46: nit: same

core/src/syscall/precompiles/fp384/fp12.rs
```diff
@@ -0,0 +1,1167 @@
...
+        &mut self,
+        dest: &mut FieldOpCols<F, U384Field>,
+        a: &Self::DType,
+        b: &Self::DType,
+    ) -> Self::DType;
+    fn _sub(
```
tamirhemo, line 52: nit: same

core/src/syscall/precompiles/fp384/fp12.rs
```diff
@@ -0,0 +1,1167 @@
...
+use std::marker::PhantomData;
+use std::mem::size_of;
+use typenum::Unsigned;
+
+use super::Fp12;
+
+/// The number of columns in the FpMulCols.
+const NUM_COLS: usize = size_of::<Fp12MulCols<u8>>();
+
+trait Fp12Builder<F> {
```
tamirhemo, line 37: The builder name is usually reserved for something that also satisfies `AirBuilder` in our crates.

Issue #1024, opened 2024-07-03
mul_assign will panic if scalar is zero


jtguibas:
https://github.com/succinctlabs/sp1/blob/14eb569d41d24721ffbd407d6060e202482d659c/zkvm/precompiles/src/utils.rs#L75

PR #1077, opened 2024-07-10
fix: handle mul_assign scalar zero

mattstam:
fixes https://github.com/succinctlabs/sp1/issues/1024

### DIFF
--- a/zkvm/lib/src/utils.rs
+++ b/zkvm/lib/src/utils.rs
@@ -11,6 +11,11 @@ pub struct AffinePoint<C: CurveOperations<NUM_WORDS>, const NUM_WORDS: usize> {
     _marker: std::marker::PhantomData<C>,
 }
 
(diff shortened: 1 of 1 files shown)


zkvm/lib/src/utils.rs
```diff
@@ -53,12 +53,17 @@ impl<C: CurveOperations<NUM_WORDS> + Copy, const NUM_WORDS: usize> AffinePoint<C
         C::double(&mut self.limbs);
     }
 
-    pub fn mul_assign(&mut self, scalar: &[u32]) {
+    pub fn mul_assign(&mut self, scalar: &[u32]) -> Option<()> {
```
jtguibas, line 56: Better to return a reuslt here.

mattstam: updated

jtguibas (approved):

PR #1091, opened 2024-07-11
chore: add more context to input stream panic

mattstam:


core/src/syscall/hint.rs
```diff
@@ -12,7 +12,11 @@ impl SyscallHintLen {
 impl Syscall for SyscallHintLen {
     fn execute(&self, ctx: &mut SyscallContext, _arg1: u32, _arg2: u32) -> Option<u32> {
         if ctx.rt.state.input_stream_ptr >= ctx.rt.state.input_stream.len() {
-            panic!("not enough vecs in hint input stream");
+            panic!(
+                "not enough vecs in hint input stream: input_stream_ptr={}, input_stream_len={}",
```
jtguibas, line 16: better error to a user might be "stdin is empty: ...", as they use SP1Stdin

jtguibas (approved):

core/src/syscall/hint.rs
mattstam: made it better

PR #1093, opened 2024-07-12
feat: exp reverse bits in ir

tqn:
Adds IR and circuit instructions for `ExpReverseBits` and `HintBits` (bit decomposition of a base field element).

Note: `HintBits` does not invoke a `receive` interaction on its inputs, but it invokes a `send` interaction on its output bits. One may believe that this means that values that are still used in the final computation may be written with zero multiplicity, but since every invocation of `HintBits` comes with assertions (see below), this is not true.

An alternative would be to `receive` all the inputs, but since the values are not being constrained for the `HintBits` operation itself, this seems kind of useless.

When an IR `HintBits` instruction is compiled, it yields many assert statements that constrain the result to be the bit decomposition. These constraints are untested, but I checked that they do in fact yield instructions. Additionally, I essentially just copy-pasted them, so they should work.


recursion/compiler/src/circuit/builder.rs
```diff
@@ -0,0 +1,63 @@
...
+        let x: SymbolicFelt<_> = output
+            .iter()
+            .enumerate()
+            .map(|(i, &bit)| {
+                self.assert_felt_eq(bit * (bit - C::F::one()), C::F::zero());
+                bit * C::F::from_canonical_u32(1 << i)
+            })
+            .sum();
+
+        // let sum: Felt<_> = self.eval(C::F::zero());
```
jtguibas, line 32: can u delete this

recursion/compiler/src/circuit/builder.rs
```diff
@@ -0,0 +1,63 @@
...
+        input: Felt<C::F>,
+        power_bits: Vec<Felt<C::F>>,
+    ) -> Felt<C::F> {
+        let output: Felt<_> = self.uninit();
+        self.operations
+            .push(DslIr::CircuitV2ExpReverseBits(output, input, power_bits));
+        output
+    }
+    fn poseidon2_permute_v2(&mut self, array: [Felt<C::F>; WIDTH]) -> [Felt<C::F>; WIDTH] {
+        // Make new felts and then pass it along
```
jtguibas, line 57: comment should always end in period

jtguibas (approved):

jtguibas (approved):

PR #1095, opened 2024-07-12
feat: parallel recursion tracegen

ctian1:
* parallel recursion tracegen
* also fix deferred bug in compress program


core/src/utils/mod.rs
```diff
@@ -186,3 +193,24 @@ pub fn log2_strict_usize(n: usize) -> usize {
     assert_eq!(n.wrapping_shr(res), 1, "Not a power of two: {n}");
     res as usize
 }
+
+pub fn par_for_each_row<P, F>(vec: &mut [F], num_cols: usize, processor: P)
```
jtguibas, line 197: OO nice.

jtguibas (approved):

PR #1102, opened 2024-07-15
ci: use global workspace version

mattstam:
[release-plz](https://release-plz.ieni.dev/) bumps each crate `version` individually based on changes, and I was concerned they would get out of sync with the "global" `version` that we have.

The [solution](https://github.com/MarcoIeni/release-plz/issues/1557#issuecomment-2226803542) is to define a `version` at the workspace `Cargo.toml` level, and have the crates reference that `version`. Also doing this for `edition` and `license` since those are also consistent throughout the workspace (leaving `description` and `readme` because long-term it will have these as different).

**This will allow new crate releases to always stay in-sync on the same `version` defined in the workspace Cargo.toml**.

Also adding `repository`, `keywords`, and `categories` as this helps with crate.io searchability.


Dockerfile.gnark-ffi
```diff
@@ -23,7 +23,7 @@ WORKDIR /sp1/recursion/gnark-cli
 
 RUN \
   --mount=type=cache,target=target \
-  cargo build --release && cp target/release/sp1-recursion-gnark-cli /gnark-cli
+  cargo build --release && cp ../../target/release/sp1-recursion-gnark-cli /gnark-cli
```
mattstam, line 26: note: this is required for putting `recursion/gnark-cli` in workspace

jtguibas (approved):

PR #1114, opened 2024-07-16
fix: lookup id collisions

jtguibas:
at around ~4 billion lookup ids, the chance of a collision is 50% which could cause some non-determinism. switched to using u128 which should require 10^19 samples...


tamirhemo (approved):
LGTM

PR #1118, opened 2024-07-17
perf: poseidon2 parallel tracegen

kevjue:


core/src/utils/mod.rs
```diff
@@ -194,23 +194,33 @@ pub fn log2_strict_usize(n: usize) -> usize {
     res as usize
 }
 
-pub fn par_for_each_row<P, F>(vec: &mut [F], num_cols: usize, processor: P)
-where
+pub fn par_for_each_row<P, F>(
+    vec: &mut [F],
+    num_rows_per_event: usize,
+    num_cols: usize,
```
ctian1, line 200: it's a bit confusing that these values aren't really used individually. should it just be one variable `num_items_per_row` or something? maybe the term row is misleading then though

recursion/core/src/poseidon2_wide/trace.rs
```diff
@@ -34,43 +39,71 @@ impl<F: PrimeField32, const DEGREE: usize> MachineAir<F> for Poseidon2WideChip<D
...
-                &mut rows,
-                || {
-                    let mut padded_row = vec![F::zero(); num_columns];
-                    self.populate_permutation([F::zero(); WIDTH], None, &mut padded_row);
-                    padded_row
-                },
-                self.fixed_log2_rows,
-            );
+            padded_rows.for_each(|padded_row| {
+                self.populate_permutation([F::zero(); WIDTH], None, padded_row);
```
ctian1, line 102: this could be slow if populate_permutation is slow? could be better to populate once and copy to all the rows

ctian1 (approved):

PR #1119, opened 2024-07-17
fix: `sp1-helper` workspace build

ratankaliani:
Fix the package lock issues when building the program with `sp1-helper`.

Note: In the previous PR as build scripts for a `program` and `script` in the same workspace were left untested. I've confirmed these work now.


build/src/lib.rs
```diff
@@ -256,12 +280,19 @@ pub fn build_program(args: &BuildArgs, program_dir: Option<PathBuf>) -> Result<U
 
     // Get the command corresponding to Docker or local build.
     let cmd = if args.docker {
-        docker::create_docker_command(args, &metadata.workspace_root, &program_dir)?
+        docker::create_docker_command(args, &program_dir, &metadata.workspace_root)?
```
ratankaliani, line 283: Note: This was broken, and was not caught when I last tested this.

build/src/lib.rs
```diff
@@ -128,22 +129,39 @@ fn get_rust_compiler_flags() -> String {
...
 }
 
-/// Execute the command and handle the output. Note: Strip the rustc configuration if this is called
-/// by sp1-helper so it uses the Succinct toolchain.
-fn execute_command(mut command: Command, build_with_helper: bool, docker: bool) -> Result<()> {
-    // Strip the rustc configuration if this is called by sp1-helper, otherwise it will attempt to
-    // compile the SP1 program with the toolchain of the normal build process, rather than the
-    // Succinct toolchain.
+/// Execute the command and handle the output depending on the context.
+fn execute_command(
```
ctian1, line 144: it'd be a bit better for decoupling/clarity if this command didn't have helper specific logic. Do we even need the if statement? Could just do the logic every time.

ctian1 (approved):

build/src/lib.rs
ratankaliani: I haven't tested the first command in non-helper when removing rustc, the second command is there to reduce build sizes, as you're re-building the entire directory inside of the subdirectory on the target runs.

ratankaliani: Simplified. Removed helper-specific logic from both functions.

PR #1122, opened 2024-07-17
docs: update outdated quickstart in SP1 book

grjte:
The SP1 book quickstart page seems to be outdated. This PR corrects the sample output and information related to modifying the program or script and running with different values of `n`.

Not addressed:
The new fibonacci project now contains `contracts` directory, which means the project structure in the quickstart is outdated. There's also an option to prove for the EVM and create a fixture, which isn't documented here. I didn't change this, since it would complicate the quickstart.


jtguibas:
Hi @grjte, thanks for the this PR! Unfortunately, our latest changes are all contained in `dev`, so this PR may not get merged for a while due to merge conflicts.

PR #1125, opened 2024-07-19
perf: generate dep optimizations

kevjue:


core/src/bytes/event.rs
```diff
@@ -9,28 +11,41 @@ use serde::{Deserialize, Serialize};
...
     /// The first input operand.
-    pub b: u32,
+    pub b: u8,
 
     /// The second input operand.
-    pub c: u32,
+    pub c: u8,
+}
+
+impl Hash for ByteLookupEvent {
```
jtguibas, line 38: i'm skeptcail this is faster

jtguibas: did you try it?

kevjue: Ya, I did.  It did speed it up significantly.  

The original implementation will hash each individual elements, e.g.

```
impl $crate::hash::Hash for ByteLookupEvent {
    fn hash<H: $crate::hash::Hasher>(&self, ra_expand_state: &mut H) {
        match self {
            ByteLookupEvent {
                shard: shard,
                channel: channel,
                opcode: opcode,
                a1: a1,
                a2: a2,
                b: b,
                c: c,
            } => {
                shard.hash(ra_expand_state);
                channel.hash(ra_expand_state);
                opcode.hash(ra_expand_state);
                a1.hash(ra_expand_state);
                a2.hash(ra_expand_state);
                b.hash(ra_expand_state);
                c.hash(ra_expand_state);
            }
        }
    }
}
```

So both the number of hash invocations and the amount of data it's hashing is improved.

kevjue: It improved the gen dep time by ~25% for the chips with lots of these events (e.g. CPU, AddSub).

jtguibas (approved):

PR #1141, opened 2024-07-19
test: rough bench of `compress` with dummy circuit

tqn:
Also widened the base ALU and memory tables, each via a configurable constant factor. We should consider doing this for some other tables.


recursion/compiler/src/circuit/mod.rs
```diff
@@ -3,3 +3,117 @@ mod compiler;
...
+
+    const DEGREE: usize = 3;
+
+    type SC = BabyBearPoseidon2Inner;
+    type F = <SC as StarkGenericConfig>::Val;
+    type EF = <SC as StarkGenericConfig>::Challenge;
+    type A = RecursionAir<F, DEGREE>;
+
+    /// Rough test to give an idea of how long the compress stage would take on the v2 circuit
+    /// relative to the recursion VM. The constants below were manually populated by running
```
jtguibas, line 32: nit:

you should have the first sentence be its own paragraph

recursion/compiler/src/circuit/mod.rs
```diff
@@ -3,3 +3,117 @@ mod compiler;
...
+        );
+        runtime.run();
+
+        // Construct the machine ourselves so we can pad the tables, avoiding `A::machine`.
+        let config = SC::default();
+        let chips: Vec<Chip<F, A>> = vec![
+            A::Memory(MemoryChip::default()),
+            A::BaseAlu(BaseAluChip::default()),
+            A::ExtAlu(ExtAluChip::default()),
+            // A::Poseidon2Skinny(
```
jtguibas, line 89: nit: cleanup

jtguibas (approved):

PR #1146, opened 2024-07-22
feat: streaming prover for core

jtguibas:


core/src/utils/prove.rs
```diff
@@ -133,241 +140,385 @@ where
...
-            );
-        }
-    };
-    create_checkpoints_span.exit();
-
-    // Commit to the shards.
-    #[cfg(debug_assertions)]
-    let mut debug_records: Vec<ExecutionRecord> = Vec::new();
+    // Record the start of the process.
+    let proving_start = Instant::now();
```
tamirhemo, line 151: nit: should this be a span?

jtguibas: it gets used in the e2e log

tamirhemo (approved):

PR #1149, opened 2024-07-22
perm: trace checkpoint optimizations

kevjue:


core/src/cpu/event.rs
```diff
@@ -1,7 +1,64 @@
+use std::array;
+
```
jtguibas, line 2: nit: CpuEvent should be the first thing in this file, not LookupIds given the name of the file is event.rs

core/src/cpu/event.rs
```diff
@@ -1,7 +1,64 @@
...
 use serde::{Deserialize, Serialize};
 
 use crate::runtime::Instruction;
+use crate::runtime::LookupIdSampler;
 use crate::runtime::MemoryRecordEnum;
+use crate::runtime::Opcode;
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
```
jtguibas, line 10: nit: keep style consistent with rest of repo and add a comment for this struct

core/src/cpu/event.rs
```diff
@@ -1,7 +1,64 @@
...
+pub enum LookupIds {
+    AluLookupId(u128),
+    SyscallLookupId(u128),
+    MemoryLookupIds([u128; 2]),
+    BranchLookupIds([u128; 3]),
+    JumpLookupIds([u128; 2]),
+    AuipcLookupId(u128),
+
```
jtguibas, line 18: nit: this space / comment here is kind of random, would just remove for style reasons

core/src/cpu/event.rs
jtguibas, line 10: a better name for this struct is "CpuLookupIds"

core/src/cpu/trace.rs
```diff
@@ -123,12 +124,18 @@ impl CpuChip {
...
-        cols.nonce = F::from_canonical_u32(
-            nonce_lookup
-                .get(&event.alu_lookup_id)
-                .copied()
-                .unwrap_or_default(),
-        );
+        if event.instruction.is_alu_instruction() {
+            let alu_lookup_id = match event.lookup_ids {
+                LookupIds::AluLookupId(id) => id,
+                _ => panic!("Expected AluLookupId"),
```
jtguibas, line 130: most panics in the repo have lowercase error messages

core/src/runtime/mod.rs
```diff
@@ -116,6 +118,11 @@ pub struct Runtime<'a> {
 
     /// The maximum number of cpu cycles to use for execution.
     pub max_cycles: Option<u64>,
+
+    pub preallocated_random_nums: Vec<u128>,
```
jtguibas, line 122: every other attribute in this struct has a comment, why no comments on these?

jtguibas, line 122: why is this called random_nums, not lookup_ids?

core/src/runtime/mod.rs
```diff
@@ -134,6 +141,8 @@ pub enum ExecutionError {
     Unimplemented(),
 }
 
+const NUM_PREALLOCATED_RANDOM_NUMS: usize = 1_000_000;
```
jtguibas, line 144: constants should almost always be at the top of the file. this is a weird place to put a const

core/src/runtime/mod.rs
```diff
@@ -1255,6 +1287,73 @@ impl<'a> Runtime<'a> {
     }
 }
 
+pub trait LookupIdSampler {
```
jtguibas, line 1290: why is this a trait? when would another struct implement this

core/src/runtime/state.rs
```diff
@@ -133,3 +133,16 @@ impl ExecutionState {
         Ok(())
     }
 }
+
+#[derive(Debug, Clone, Default, Copy, PartialEq, Eq)]
+pub struct EventCounts {
```
jtguibas, line 138: 1) why is this struct here if it's not in ExecutionState
2) just use HashMap<Opcode, usize> for most of the fields?

core/src/utils/options.rs
```diff
@@ -30,6 +30,7 @@ pub struct SP1CoreOpts {
     pub prove_stream_capacity: usize,
     pub split_opts: SplitOpts,
     pub reconstruct_commitments: bool,
+    pub event_counts: EventCounts,
```
jtguibas, line 33: why is event counts in sp1 core opts?

jtguibas: this feels wrong, you can just put it into ExecutionState

core/src/cpu/event.rs
```diff
@@ -1,7 +1,64 @@
...
+        } else if instr.is_jump_instruction() {
+            LookupIds::JumpLookupIds([lookup_ids[0], lookup_ids[1]])
+        } else if instr.opcode == Opcode::AUIPC {
+            LookupIds::AuipcLookupId(lookup_ids[0])
+        } else {
+            LookupIds::DefaultLookupIds
+        }
+    }
+
+    pub fn new_sublookups(rng_sampler: &mut impl LookupIdSampler) -> [u128; 6] {
```
jtguibas, line 57: what's the point of having this enum if we're just going to use [u128; 6]?

shouldn't we do the same for Alu lookups?

core/src/runtime/mod.rs
```diff
@@ -1255,6 +1287,73 @@ impl<'a> Runtime<'a> {
...
+                    let mut rng = rand::thread_rng();
+                    chunk.iter_mut().for_each(|num| {
+                        *num = rng.gen();
+                    });
+                });
+
+            self.preallocated_random_nums_idx = 0;
+        }
+
+        let rnd_nums = self.preallocated_random_nums
```
jtguibas, line 1315: instead of doing this you should just use VecDeque and then just push_back and pop_front.

core/src/utils/options.rs
jtguibas: oh hm maybe its cause you need opts in new() in Runtime?

core/src/runtime/mod.rs
kevjue: I was considering using a VecDeque, but I don't think it supports multi threaded inserts.  That's one of the main performance gains from the PR.

core/src/runtime/mod.rs
kevjue: Main reason is that seems weird to pass in a reference to a runtime object here: https://github.com/succinctlabs/sp1/blob/f2ece44c6f303e8b785c958ff62d90ddf185af8d/core/src/cpu/event.rs#L75

core/src/cpu/event.rs
kevjue: Ya, this shouldn't be part of the enum function.  I didn't want to add a variant with [u128; 6] internal type, since that would made the size of CpuLookupIds much bigger.

kevjue: The reason I didn't do it for ALU lookups is that it seems like all 6 of the sub_lookup ids are needed for the divrem chip.

I was considering making the sublookups optional, but there will be no space savings (an Option type is as big as the internal type).

However, it would save on the time to generate random ids, so its probably worth making it optional.

core/src/utils/options.rs
kevjue: There are a few reasons why it's not part of the state:

1) In the runtime recover function:  https://github.com/succinctlabs/sp1/blob/545ab109a75ccf942ba78b43014fe2c4928609d3/core/src/runtime/mod.rs#L242, the state is assigned after the creation of the runtime.
2) Semantically, it seems like the ExecutionState stores information about the runtime state as opposed to statistics of it.  So you could argue it should be a separate object.

core/src/runtime/state.rs
kevjue: Ya, I'll move it to another file and use a hashmap.

kevjue: Actually, I ended up creating a static array so that it can have the Copy trait.

PR #1174, opened 2024-07-25
feat: verifier circuit `verify_challenges` and `verify_two_adic_pcs`

tqn:


recursion/circuit-v2/src/fri.rs
```diff
@@ -0,0 +1,733 @@
...
+        &proof.commit_phase_openings,
+        betas,
+    ) {
+        folded_eval = builder.eval(folded_eval + reduced_openings[log_folded_height + 1]);
+
+        let one: Felt<_> = builder.constant(C::F::one());
+        let index_sibling: Felt<_> = builder.eval(one - index_bits[offset]);
+        let index_pair = &index_bits[(offset + 1)..];
+
+        let evals_ext = select_chain(
```
jtguibas, line 206: nice

recursion/circuit-v2/src/fri.rs
```diff
@@ -0,0 +1,733 @@
...
+        let (commit, rounds) = const_two_adic_pcs_rounds(&mut builder, commit.into(), os);
+        let mut challenger = DuplexChallengerVariable::new(&mut builder);
+        challenger.observe_commitment(&mut builder, commit);
+        let x2 = challenger.sample_ext(&mut builder);
+        let x1: Ext<_, _> = builder.constant(x1);
+        builder.assert_ext_eq(x1, x2);
+        verify_two_adic_pcs(&mut builder, &config, &proof, &mut challenger, rounds);
+
+        run_test_recursion(builder.operations);
+        // let mut backend = ConstraintCompiler::<InnerConfig>::default();
```
jtguibas, line 729: nit: do we want to comment this in?

jtguibas (approved):

recursion/circuit-v2/src/fri.rs
tqn: It doesn't work because it calls an unsupported instruction `CircuitV2Poseidon2PermuteBabyBearWide`, and more generally my runtime, compiler, etc. currently do not support Bn254.

PR #1175, opened 2024-07-25
feat: streaming recursion

jtguibas:


prover/src/lib.rs
```diff
@@ -191,8 +194,8 @@ impl<C: SP1ProverComponents> SP1Prover<C> {
 
         Self {
             recursion_program,
-            rec_pk,
-            rec_vk,
+            recursion_pk,
```
tamirhemo, line 197: nice, much better

prover/src/lib.rs
```diff
@@ -424,109 +431,329 @@ impl<C: SP1ProverComponents> SP1Prover<C> {
...
+            let input_rx = Arc::new(Mutex::new(input_rx));
+            for _ in 0..opts.recursion_opts.trace_gen_workers {
+                let record_and_trace_sync = Arc::clone(&record_and_trace_sync);
+                let record_and_trace_tx = Arc::clone(&record_and_trace_tx);
+                let input_rx = Arc::clone(&input_rx);
+                let span = tracing::debug_span!("generate records and traces");
+                s.spawn(move || {
+                    let _span = span.enter();
+                    loop {
+                        let received = { input_rx.lock().unwrap().recv() };
```
tamirhemo, line 496: This might be a very small different but it seems a bit inefficient that we have a Mutex on the Receiver because they would basically compete for the same resource. Another way to do it is to have one loop on the recv() and send tasks to workers, this might reduce context switching but it might also end up being the same.

prover/src/lib.rs
```diff
@@ -424,109 +431,329 @@ impl<C: SP1ProverComponents> SP1Prover<C> {
...
+                            // Generate the traces.
+                            let record = records.into_iter().next().unwrap();
+                            let traces = tracing::debug_span!("generate traces")
+                                .in_scope(|| self.compress_prover.generate_traces(&record));
+
+                            // Wait for our turn to update the state.
+                            record_and_trace_sync.wait_for_turn(index);
+
+                            // Send the record and traces to the worker.
+                            record_and_trace_tx
```
tamirhemo, line 566: This is for ordering purposes?

tamirhemo (approved):

PR #1188, opened 2024-07-26
feat(cli): tracer

puma314:
Added a RISC-V tracer command to the `cargo prove` CLI that will allow users to trace which functions are expensive.

Usage:
- Use the `TRACE_FILE` environment variable set to a file name, when executing the ELF (with inputs) to save a trace of the `pc` to the specified file
- Use `cargo prove trace -- --elf path_to_elf --trace trace_file` to output a trace of function calls


cli/src/commands/trace.rs
```diff
@@ -0,0 +1,456 @@
...
+use textwrap::wrap;
+
+#[derive(Parser, Debug)]
+#[command(
+    name = "trace",
+    about = "Trace a program execution and analyze cycle counts."
+)]
+pub struct TraceCmd {
+    #[arg(short, long, default_value_t = 30)]
+    /// Include the "top" number of functions
```
jtguibas, line 43: nit: a bit unusual to have the comment under the macro

cli/src/commands/trace.rs
```diff
@@ -0,0 +1,456 @@
...
+    if let Some(index) = function_stack.iter().position(|s| s == function_name) {
+        let truncated_stack = &function_stack[0..=index];
+        let count = filtered_stack_counts
+            .entry(truncated_stack.to_vec())
+            .or_insert(0);
+        *count += num_instructions;
+    }
+}
+
+fn _build_radare2_lookups(
```
jtguibas, line 152: can delete?

cli/src/commands/trace.rs
jtguibas: also usually attribute comments have periods at end throughput repo

jtguibas (approved):

PR #1189, opened 2024-07-26
refactor: `sp1-lib`, syscall docs

jtguibas:
## Overview

Refactor the `sp1-lib` crate in `zkvm`, and add testing for the `ecdsa-core` patch for `k256` to the `patch-testing` program.

## Details
- Simplify curve modules in `sp1-lib`. Create an `AffinePoint` trait which is re-used across curves.
   - Move `ecrecover` logic to `ecdsa-core` patch for maintainability https://github.com/sp1-patches/signatures/pull/2. Note: Decision to refactor these crates was driven by the circular dependency caused when trying to patch `k256`.
   - Remove `decompress_pubkey` logic from BLS in `sp1-lib`.
- Add better documentation and use safer types for syscalls.

## Note on SemVer
Includes backwards-incompatible changes in `sp1-lib`, but because it only causes issues for developers who are creating patches for crates themselves with `sp1-lib` (which is highly uncommon), we've decided to not update the major version of `sp1-lib` in this release, although that would follow convention.


examples/patch-testing/program/src/main.rs
```diff
@@ -1,28 +1,54 @@
...
+    let input = [1u8; 32];
+    let y = CompressedEdwardsY(input);
+    let _ = y.decompress();
+}
 
-    // Test SHA256.
+/// Emits KECCAK_PERMUTE syscalls.
+fn test_keccak() {
+    let input = [1u8; 32];
+    let _ = keccak256(input);
```
ctian1, line 46: would be good to validate this and the dalek_ng above return correct results

ctian1 (approved):

examples/patch-testing/program/src/main.rs
ratankaliani: Validated.

PR #1191, opened 2024-07-26
feat: early stopping

erabinov:


Cargo.toml
```diff
@@ -62,46 +62,46 @@ sp1-recursion-circuit-v2 = { path = "recursion/circuit-v2", version = "1.0.0-rc.
...
-# p3-blake3 = { path = "../Plonky3/blake3" }
-# p3-mds = { path = "../Plonky3/mds" }
-# p3-merkle-tree = { path = "../Plonky3/merkle-tree" }
-# p3-poseidon2 = { path = "../Plonky3/poseidon2" }
-# p3-symmetric = { path = "../Plonky3/symmetric" }
-# p3-uni-stark = { path = "../Plonky3/uni-stark" }
-# p3-maybe-rayon = { path = "../Plonky3/maybe-rayon" }
-# p3-bn254-fr = { path = "../Plonky3/bn254-fr" }
+# p3-air = { path = "../my_p3_fork/Plonky3/air" }
+# p3-field = { path = "../my_p3_fork/Plonky3/field" }
```
jtguibas, line 89: can we keep it the way it was?

core/src/utils/config.rs
```diff
@@ -63,6 +63,7 @@ pub fn sp1_fri_config() -> FriConfig<InnerChallengeMmcs> {
     };
     FriConfig {
         log_blowup: 1,
+        log_final_poly_len: 4,
```
jtguibas, line 66: do we feel pretty confident this is the best parameter choice?

PR #1232, opened 2024-08-02
chore: bump version and changelog

mattstam:


book/developers/common-issues.md
```diff
@@ -12,7 +12,7 @@ This is likely due to two different versions of `alloy_sol_types` being used. To
 
 ```toml
 [dependencies]
-sp1-sdk = { version = "1.0.1", default-features = false }
```
ctian1, line 15: can we maybe make these placeholder so we don't need to update every version?

tests/bls12381-add/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 [package]
 name = "bls12381-add-test"
-version = "1.0.1"
+version = "1.1.0"
```
ctian1, line 4: we should not need to version tests/examples. they could just always be 0.1.0? since they aren't published or used externally

ctian1 (approved):

jtguibas (approved):

book/developers/common-issues.md
mattstam: hmm I will look into this later

tests/bls12381-add/Cargo.toml
mattstam: yeah your right, Ill reset those next release

PR #1242, opened 2024-08-04
chore: update gpu image

jtguibas:


mattstam (approved):

mattstam:
we can still release this crate under v1.1.0, lmk if you want it before or after this change

PR #1244, opened 2024-08-05
perf: remove `read_mult` column, instead use negative `write_mult`

tqn:


recursion/core-v2/src/chips/exp_reverse_bits.rs
```diff
@@ -122,17 +122,14 @@ impl<F: PrimeField32, const DEGREE: usize> MachineAir<F> for ExpReverseBitsLenCh
                     row.is_real = F::one();
                     row.x_mem = MemoryAccessCols {
                         addr: addrs.base,
-                        read_mult: *mult * F::from_bool(i == 0),
-                        write_mult: F::zero(),
+                        write_mult: *mult * F::neg_one() * F::from_bool(i == 0),
```
kevjue, line 125: Maybe rename `write_mult` column to `mult`?

tqn: I left it as `write_mult` because I want to make it clear what a "positive" mult means. I don't think it would be a good idea to leave the name neutral. However, I also don't feel great about the name. Maybe I could rename `MemoryAccessCols` to `MemoryWriteCols` but that's not that great either.

erabinov: Maybe `write_minus_read` or `write_minus_read_mult`?

tqn: I changed it to `mult` and wrote a few doc comments.

kevjue (approved):
LGTM!

PR #1252, opened 2024-08-05
fix: docker curl sp1 cache break

mattstam:


cli/docker/Dockerfile
```diff
@@ -1,12 +1,16 @@
 FROM ubuntu:24.04@sha256:e3f92abc0967a6c19d0dfa2d55838833e947b9d74edbcb0113e48535ad4be12a
 
+ARG BUILDTIME
```
ctian1: should be further down since apt can be cached

ctian1 (approved):

PR #1274, opened 2024-08-07
feat(cli): vkey command to output program vkey hash

mattstam:
```sh
cargo prove vkey --elf examples/tendermint/program/elf/riscv32im-succinct-zkvm-elf

Verification Key Hash:
0x00c8cf1fbd86e856e8cb5b1dabfff9d42bf08b08ac18255db5170fe885cf3c37
```

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "sp1-cli"
-description = "SP1 is a performant, 100% open-source, contributor-friendly zkVM."
-readme = "../README.md"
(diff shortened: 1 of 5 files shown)


cli/src/commands/vkey.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Parser)]
+#[command(name = "vkey", about = "View the verification key hash for a program.")]
+pub struct VkeyCmd {
+    /// Path to the ELF.
+    #[arg(long, required = true)]
+    elf: String,
+}
+
+impl VkeyCmd {
```
mattstam, line 17: will update template later to use this command in instructions and remove our custom bin to get the vkey hash

jtguibas (approved):

PR #1288, opened 2024-08-08
chore: `verify_sp1_proof` naming

ratankaliani:
Correct the naming for the inputs to `verify_sp1_proof` and make them consistent.

### DIFF
--- a/zkvm/lib/src/verify.rs
+++ b/zkvm/lib/src/verify.rs
@@ -3,7 +3,7 @@ use crate::syscall_verify_sp1_proof;
 /// Verifies the next proof in the proof input stream given a verification key digest and public
 /// values digest. If the proof is invalid, the function will panic.
 ///
-/// Enable this function in `sp1-lib` with the `verify` feature.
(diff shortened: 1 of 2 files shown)


zkvm/lib/src/verify.rs
```diff
@@ -1,10 +1,11 @@
 use crate::syscall_verify_sp1_proof;
 
-/// Verifies the next proof in the proof input stream given a pkey digest and public values digest.
+/// Verifies the next proof in the proof input stream given a verification key digest and public
+/// values digest. If the proof is invalid, the function will panic.
 ///
-/// Note: sp1_zkvm must also have feature `verify` enabled for this function to work.
-pub fn verify_sp1_proof(pkey_digest: &[u32; 8], pv_digest: &[u8; 32]) {
+/// Enable this function in `sp1-lib` with the `verify` feature.
```
ctian1, line 6: this note is saying that the entrypoint `sp1_zkvm` crate also needs verify feature, not just sp1-lib

ratankaliani: Will add back good catch

PR #1311, opened 2024-08-13
feat: circuit types for the constant proof shape case

tamirhemo:


recursion/circuit-v2/src/witness.rs
```diff
@@ -114,47 +111,77 @@ impl<C: Config, T: Witnessable<C>> Witnessable<C> for Vec<T> {
...
+//         [
+//             self.proof.commitment.write(),
+//             self.proof.opened_values.write(),
+//             self.proof.opening_proof.write(),
+//             Witnessable::<C>::write(&self.proof.public_values),
+//         ]
+//         .concat()
+//     }
+// }
+
```
tqn, line 184: Should we remove this?

tqn (approved):

PR #1315, opened 2024-08-13
chore: fix `fp` tests, record

0xWOLAND:
- Fix `Fp` type tests for Bls and Bn precompiles
- Add Bn254 `Fp2` precompiles to record (caused nonce issue)


tests/bls12381-fp/Cargo.toml
```diff
@@ -7,7 +7,7 @@ publish = false
```
tamirhemo, line 7: should also change version to 1.1.1

0xWOLAND: fixed

tamirhemo (approved):
LGTM

PR #1317, opened 2024-08-14
feat: faster runtime

jtguibas:


core/src/runtime/utils.rs
```diff
@@ -74,6 +76,8 @@ impl<'a> Runtime<'a> {
                 self.state.global_clk,
                 self.state.pc
             );
+            println!("{}", self.report_single);
```
ratankaliani, line 79: u print the report here just comment it

ctian1 (approved):

PR #1318, opened 2024-08-14
feat: `decompress pubkey`

0xWOLAND:


core/src/runtime/mod.rs
```diff
@@ -853,6 +855,9 @@ impl<'a> Runtime<'a> {
                         .and_modify(|c| *c += 1)
                         .or_insert(1);
                 }
```
jtguibas, line 857: nit: add space here

jtguibas (approved):

PR #1335, opened 2024-08-17
fix: rt block on cuda

jtguibas:


cuda/src/lib.rs
```diff
@@ -313,6 +306,19 @@ fn cleanup_container(container_name: &str) {
     }
 }
 
+/// Utility method for blocking on an async function. If we're already in a tokio runtime, we'll
+/// block in place. Otherwise, we'll create a new runtime.
+fn block_on<T>(fut: impl Future<Output = T>) -> T {
```
ratankaliani, line 311: FYI: This exists in https://github.com/succinctlabs/sp1/blob/6a6c06481712eb5362dcb1dbe37a241e932dd690/crates/sdk/src/install.rs#L110-L121

jtguibas: oh i see, ill move it to utils.

PR #1339, opened 2024-08-19
perf: push traces iff `SP1_DEBUG` is set in v2 compiler

tqn:


recursion/compiler/src/ir/builder.rs
```diff
@@ -49,18 +50,11 @@ impl<T> TracedVec<T> {
...
-                self.traces.push(Some(Backtrace::new_unresolved()));
-            }
-            _ => {
-                self.traces.push(None);
-            }
-        };
+        if sp1_debug_mode() {
+            self.traces.push(Some(Backtrace::new_unresolved()));
+        } else {
+            self.traces.push(None);
```
tamirhemo, line 56: Is it necessary to have this vector of `None`?

tqn: I think the old runtime may panic because it will try to look up a trace by index in `traces`.

tamirhemo: Oh I see now that this is the code that was here before. I think you can leave it as it for now.

tamirhemo (approved):
LGTM

PR #1360, opened 2024-08-20
perf: use crate `vec_map`, box large `Instruction` variants

tqn:
Additionally:
- Correct documentation for `sp1_debug_mode`.
- Calculate total memory used at compile time and reserve the memory in the runtime.
- Create a trait for Runtime's memory and two implementations, backed by `VecMap` and `Vec`. There is a `HashMap` implementation in the PR history, too. Currently, only the `VecMap` implementation is used.

Potential future improvements:
- Avoid converting between `Address<C::F>` and `usize` all the time in the compiler. This is probably really cheap, though.
- Avoid striping `Var`, `Felt`, and `Ext`. Currently they have different counters and are thus each assigned a residue class mod 3 to use for "virtual addresses."

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5350,6 +5350,7 @@ dependencies = [
  "sp1-recursion-derive",
  "sp1-stark",
  "tracing",
+ "vec_map",
 ]
 
 [[package]]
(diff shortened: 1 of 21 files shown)


tamirhemo (approved):
LGTM

PR #1372, opened 2024-08-22
perf: unify variable counters

tqn:
Enables a contiguous `VecMap` in the compiler.

Depends on #1371 and will need a rebase.


tamirhemo (approved):
LGTM

PR #1374, opened 2024-08-22
feat: unify inner and outer witnesses in recursion circuit

tamirhemo:


crates/recursion/circuit-v2/src/build_wrap_v2.rs
```diff
@@ -143,7 +146,7 @@ where
 
 /// A utility function to convert a `ShardProof` into a `ShardProofVariable`. Should be replaced by
 /// more refined witness generation.
-fn const_shard_proof(
+pub fn const_shard_proof(
```
erabinov, line 149: Do you think it makes sense to "traitify" these functions? I just copy-pasted since the other one had all the configs coded in.

Since these are only used for testing, maybe it's not that important?

tamirhemo: I don't think so, we might not even use them anymore, but they are included for convenience and if someone wants to test a circuit.

erabinov:
LGTM!

erabinov (approved):

PR #1375, opened 2024-08-22
docs: precompile guide

0xWOLAND:


crates/core/machine/src/syscall/precompiles/README.md
```diff
@@ -0,0 +1,290 @@
...
+    pub x_ptr: T,
+    pub y_ptr: T,
+    pub x_memory: GenericArray<MemoryWriteCols<T>, WordsFieldElement>,
+    pub y_memory: GenericArray<MemoryReadCols<T>, WordsFieldElement>,
+    pub output: FieldOpCols<T, U256Field>,
+}
+```
+Adjust these fields according to your chip.
+
+> **Important Note:** For each field operation in your chip, there must be a corresponding column in your data structure. This ensures that the operations are properly represented in the AIR (Algebraic Intermediate Representation) framework and can be evaluated correctly.
```
tamirhemo, line 41: This comment seems a bit unclear, seems a bit too specific for the FieldOp case

tamirhemo (approved):
lgtm

PR #1415, opened 2024-08-27
feat: merkle tree-based vector commitments

erabinov:


crates/recursion/circuit-v2/src/merkle_tree.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub fn commit(leaves: Vec<HV::Digest>) -> (HV::Digest, Self) {
+        assert!(!leaves.is_empty());
+        let new_len = leaves.len().next_power_of_two();
+        let height = log2_strict_usize(new_len);
+
+        // Pre-allocate the vector.
+        let mut digest_layers = Vec::with_capacity(2 * new_len - 2);
+
+        // If `leaves.len()` is not a power of 2, we pad the leaves with default values.
+        let mut last_layer = leaves.clone();
```
tamirhemo, line 42: Why cloning here? You can just change the signature of the function to `leaves: mut Vec<HV::Digest>`. It's ok because you are taking ownership

crates/recursion/circuit-v2/src/merkle_tree.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+        if value == commitment {
+            Ok(())
+        } else {
+            Err(VcsError)
+        }
+    }
+}
+
+pub fn verify<C: CircuitConfig, HV: FieldHasherVariable<C>>(
```
tamirhemo, line 118: very nice!

crates/recursion/circuit-v2/src/merkle_tree.rs
```diff
@@ -0,0 +1,191 @@
...
+        utils::tests::run_test_recursion,
+        CircuitConfig,
+    };
+    type C = InnerConfig;
+    type F = BabyBear;
+    type HV = BabyBearPoseidon2;
+
+    #[test]
+    fn test_merkle_tree_inner() {
+        let mut rng = StdRng::seed_from_u64(0xCAFEBABE);
```
tamirhemo, line 161: can you make this random and also repeat the test a bunch of times?

crates/recursion/circuit-v2/src/merkle_tree.rs
erabinov: No good reason. Removed it.

tamirhemo (approved):
LGTM

PR #1419, opened 2024-08-27
refactor: "exp reverse bits" -> "exp bits" in v2 recursion

tqn:
The reversal is done inside the `CircuitConfig` implementations --- when the circuit is built, instead of during the runtime. I have not checked the change's effects on performance because I expect them to be negligible.


tqn:
It's a bit too complicated to refactor and a bit too unimportant right now.

PR #1427, opened 2024-08-29
Shaked/lookup

shakedregev:
I changed usages of lookup ids as u128 into structs of 4 u32s, this is done to be compatible with c data types.

### DIFF
--- a/crates/core/executor/src/events/alu.rs
+++ b/crates/core/executor/src/events/alu.rs
@@ -2,7 +2,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::Opcode;
 
-use super::create_alu_lookups;
+use super::{create_alu_lookups, LookupId};
 
 /// Arithmetic Logic Unit (ALU) Event.
 ///
@@ -11,7 +11,7 @@ use super::create_alu_lookups;
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct AluEvent {
     /// The lookup identifer.
-    pub lookup_id: u128,
+    pub lookup_id: LookupId,
     /// The shard number.
     pub shard: u32,
     /// The channel number.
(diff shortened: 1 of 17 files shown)


shakedregev:
@tamirhemo - fyi

crates/core/executor/src/events/utils.rs
```diff
@@ -1,22 +1,48 @@
+use serde::Deserialize;
+use serde::Serialize;
 use std::{
     fmt::Display,
     iter::{Map, Peekable},
 };
 
 use rand::{thread_rng, Rng};
+#[derive(Deserialize, Serialize, Debug, Clone, Copy, Default, Eq, Hash, PartialEq)]
```
jtguibas, line 9: nit: derive should be below the comment.

The comment has a lot of information (good) but it does not fit inline with the standards of the repo.

A better comment would just be

```
A unique identifier for lookups.

We use 4 u32s instead of a u128 to make it compatible with C.
```

shakedregev: Fixed

crates/core/executor/src/events/alu.rs
```diff
@@ -27,15 +27,16 @@ pub struct AluEvent {
     /// The third operand.
     pub c: u32,
     /// The result of the operation.
-    pub sub_lookups: [u128; 6],
+    /// make in the format of [``LookupId``; 6]
```
jtguibas, line 30: comments should be full sentence with proper punctuation, this is unprofessional

crates/core/executor/src/events/alu.rs
```diff
@@ -27,15 +27,16 @@ pub struct AluEvent {
...
+    pub sub_lookups: [LookupId; 6],
 }
-
+//use crate::events::create_alu_lookup_id;
```
jtguibas, line 33: please do not leave random commented lines around

jtguibas (approved):

PR #1428, opened 2024-08-29
feat: infrastructure for proof shape merkle proofs

erabinov:


crates/recursion/circuit-v2/src/machine/vkey_proof.rs
```diff
@@ -0,0 +1,163 @@
...
+    ///   implementation in this function assumes a fixed deferred verification program specified by
+    ///   `deferred_vk`.
+    /// - Compress proofs: these are proofs which refer to a prove of this program. The key for it
+    ///   is part of public values will be propagated accross all levels of recursion and will be
+    ///   checked against itself as in [sp1_prover::Prover] or as in [super::SP1RootVerifier].
+    pub fn verify(
+        builder: &mut Builder<C>,
+        digests: Vec<SC::DigestVariable>,
+        input: SP1MerkleProofWitnessVariable<C, SC>,
+        // TODO: add vk correctness check.
```
tamirhemo, line 72: nit: remove this TODO

crates/recursion/circuit-v2/src/machine/vkey_proof.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn shape(&self) -> SP1CompressShape {
+        self.compress_val.shape()
+    }
+}
+
+impl<C: CircuitConfig<F = BabyBear, EF = InnerChallenge>, SC: BabyBearFriConfigVariable<C>>
+    Witnessable<C> for SP1CompressWithVKeyWitnessValues<SC>
+where
+    Com<SC>: Witnessable<C, WitnessVariable = <SC as FieldHasherVariable<C>>::DigestVariable>,
+    // As for `SP1MerkleProofWitnessValues`, this is a redundant trait bound.
```
tamirhemo, line 145: If redundant them remove

crates/recursion/circuit-v2/src/merkle_tree.rs
```diff
@@ -67,7 +74,7 @@ impl<F: Field, HV: FieldHasher<F>> MerkleTree<F, HV> {
         (root, Self { height, digest_layers })
```
tamirhemo, line 74: Is there a reason why the root is not added to the merkle tree itself?

erabinov: I know you'll hate me for following Plonky3, but I think their distinction between Commitment and ProverData makes sense: the root is the commitment, and the ProverData should be what the prover needs to generate a proof. The root is not necessary for the prover to have after making the commitment.

PR #1436, opened 2024-08-29
fix: memory not added to checkpoint in unconstrained mode

ctian1:
We were running into a rare runtime issue where the runtime diverges from intended behavior when running from a checkpoint. Multiple checkpoints were halting, resulting in duplicate memory init/finalize records. Root cause is because mode is set to Simple in unconstrained mode, but checkpoint memory is only saved in Checkpoint mode.

Fix is to also save checkpoint memory if in unconstrained mode.


kevjue (approved):
LGTM

PR #1454, opened 2024-08-30
chore: added rsp example

yuwen01:


examples/rsp/eth-program/elf/riscv32im-succinct-zkvm-elf
jtguibas, line 1: can you not commit this file, too large. lets rely on build scripts.

examples/rsp/eth-program/Cargo.toml
jtguibas, line 1: can we just call this program, not eth-program

PR #1460, opened 2024-08-30
chore: allow optional proof shape in core record and program

erabinov:


crates/core/executor/src/program.rs
```diff
@@ -25,13 +26,15 @@ pub struct Program {
     pub pc_base: u32,
     /// The initial memory image, useful for global constants.
     pub memory_image: BTreeMap<u32, u32>,
+    /// The shape for the preprocessed tables.
+    pub shape: Option<CoreShape>,
```
tamirhemo, line 30: rename to preprocessed_shape?

crates/core/machine/src/alu/add_sub/mod.rs
```diff
@@ -105,13 +105,15 @@ impl<F: PrimeField> MachineAir<F> for AddSubChip {
             rows.extend(row_batch);
         }
 
+        pad_rows_fixed(
+            &mut rows,
+            || [F::zero(); NUM_ADD_SUB_COLS],
+            input.shape.as_ref().map(|s: &CoreShape| s.shape[&MachineAir::<F>::name(self)]),
```
tamirhemo, line 111: refactor to a method on `program` and a method on `record`

PR #1483, opened 2024-09-05
docs: `SP1VerifierGateway`

ratankaliani:


book/onchain-verification/contract-addresses.md
```diff
@@ -1,6 +1,6 @@
 # Contract Addresses
 
-When using SP1, we recommend using our deployed verifiers. Each contract is a [SP1VerifierGateway](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1VerifierGateway.sol) which can automatically routes your SP1 proof to the correct verifier based on the prover version.
+When using SP1, we recommend using our deployed verifiers. Each contract is a [SP1VerifierGateway](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1VerifierGateway.sol) which can automatically routes your SP1 proof to the correct verifier based on the prover version. Whenever a new prover version is deployed, the gateway contract will be updated to support it (`addRoute`). If an old prover version is deprecated, the gateway contract will stop supporting it (`freezeRoute`).
```
mattstam, line 3:
```suggestion
When using SP1, we recommend using our deployed verifiers. Each contract is a [SP1VerifierGateway](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1VerifierGateway.sol) which can automatically route your SP1 proof to the correct verifier based on the prover version. Whenever a new prover version is deployed, the gateway contract will be updated to support it (`addRoute`). If an old prover version is deprecated, the gateway contract will stop supporting it (`freezeRoute`).
```

mattstam, line 3: links to addRoute() and freezeRoute() code would be better

mattstam, line 3: > If an old prover version is deprecated, the gateway contract will stop supporting it (`freezeRoute`).

People probably care a lot about how much time they have to migrate. We should stick to some schedule like "assuming no vulnerabilities in the old verifier, the old verifier will be frozen 7 days after the new was added"

mattstam (changes requested):

book/onchain-verification/contract-addresses.md
ratankaliani: Updated. Didn't include 7D period, otherwise sgtm.

mattstam (approved):

PR #1498, opened 2024-09-12
feat: generate dependencies cleanup

kevjue:


crates/core/executor/src/record.rs
```diff
@@ -475,3 +394,64 @@ impl ByteRecord for ExecutionRecord {
         add_sharded_byte_lookup_events(&mut self.byte_lookups, new_events);
     }
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PrecompileEvents {
```
tamirhemo, line 399: Nice, deserves its own file

PR #1519, opened 2024-09-18
Feat: added new precompile called u256x2048Mul

umadayal:


Cargo.toml
```diff
@@ -30,7 +30,7 @@ members = [
   "crates/sdk",
   "crates/cuda",
   "crates/stark",
-  "crates/zkvm/*",
+  "crates/zkvm/*", "tests/uma-test",
```
jtguibas, line 33: why is this not on a new line like the rest of the the crates?

Also, do you see any other tests being included here? why did you add it. try to understand how the other tests work

crates/core/executor/src/syscalls/code.rs
```diff
@@ -128,6 +128,9 @@ pub enum SyscallCode {
 
     /// Executes the `BN254_FP2_MUL` precompile.
     BN254_FP2_MUL = 0x00_01_01_2B,
+
+    /// Executes the `UMA` precompile.
```
jtguibas, line 132: comment wrong

crates/core/executor/src/programs.rs
```diff
@@ -93,6 +93,9 @@ pub mod tests {
     pub const UINT256_MUL_ELF: &[u8] =
         include_bytes!("../../../../tests/uint256-mul/elf/riscv32im-succinct-zkvm-elf");
 
+    pub const U256X2048_MUL_ELF: &[u8] =
```
jtguibas, line 96: U256XU2048, not U256x2048

crates/zkvm/entrypoint/src/syscalls/bigint_uma.rs
```diff
@@ -0,0 +1,38 @@
+use super::syscall_u256x2048_mul;
+
+/// The number of limbs in a "uint256".
+const N: usize = 8;
+#[allow(unused_variables)]
```
jtguibas, line 5: perhaps this existed in where u copied from, but there should be a line between 4 and 5.

Cargo.toml
umadayal: Yeah I was a bit stuck on that. I think it added it automatically when I added the uma-test package to the tests directory. Not sure how to add it without it getting added to the root Cargo.toml in this way

.vscode/settings.json
```diff
@@ -57,6 +57,7 @@
         // "tests/sha-extend/Cargo.toml",
         // "tests/sha2/Cargo.toml",
         // "tests/verify-proof/Cargo.toml",
+        // "tests/uma-test/Cargo.toml",
```
jtguibas, line 60: let's not commit weird changes like this "uma-test"

Cargo.toml
```diff
@@ -30,7 +30,7 @@ members = [
   "crates/sdk",
   "crates/cuda",
   "crates/stark",
-  "crates/zkvm/*",
+  "crates/zkvm/*",
```
jtguibas, line 33: don't leave unecessary diffs

crates/core/executor/src/syscalls/code.rs
```diff
@@ -128,6 +128,9 @@ pub enum SyscallCode {
 
     /// Executes the `BN254_FP2_MUL` precompile.
     BN254_FP2_MUL = 0x00_01_01_2B,
+
+    /// Executes the `U256X2048_MUL` precompile.
+    U256X2048_MUL = 0x00_01_01_2C,
```
jtguibas, line 133: same here

PR #1536, opened 2024-09-23
fix: uninitialized_memory hashmap issue

ctian1:
With programs with huge amounts of input data, ExecutionState.uninitialized_memory could slow the execution down significantly due to hash collisions


crates/core/executor/src/memory.rs
```diff
@@ -185,6 +185,14 @@ impl<'a, V> Entry<'a, V> {
             Entry::Occupied(entry) => entry.into_mut(),
         }
     }
+
+    pub fn and_modify<F: FnOnce(&mut V)>(mut self, f: F) -> Self {
```
ratankaliani, line 189: nit: Add doc comment for this fn, similar to other 2

ratankaliani (approved):
Small doc comment, otherwise LGTM

PR #1542, opened 2024-09-25
fix: `multi_scalar_multiplication`

ratankaliani:
Handle the complete addition for `secp256k1_add` using a new `WeierstrassAffinePoint` trait that's implemented for secp256k1, bls12-381 and bn254. Easy to extend to `secpr1`.

- Modify `multi_scalar_multiplication` and `mul_assign` to use `complete_add_assign`, which handles all of the potential edge cases. 
- Add generic test for `secp256k1_add`, `bls12381_add` and `bn254_add` that covers the complete addition cases.


tests/common/src/weierstrass_add.rs
ratankaliani, line 1: Add a test verifying that P + (order of curve - 1) * P = point at infinity

crates/zkvm/lib/src/utils.rs
```diff
@@ -130,3 +139,57 @@ pub fn bytes_to_words_le(bytes: &[u8]) -> Vec<u32> {
         .map(|chunk| u32::from_le_bytes(chunk.try_into().unwrap()))
         .collect::<Vec<_>>()
 }
+
+/// A trait for affine points on Weierstrass curves.
+pub trait WeierstrassAffinePoint<const N: usize>: AffinePoint<N> {
+    /// The infinity point is set to (0, 0).
+    fn infinity() -> Self {
```
tamirhemo, line 146: This seems kind of bad? it would be good if an affine point is valid.

ratankaliani: Now represented as a separate type.

crates/zkvm/lib/src/secp256k1.rs
```diff
@@ -1,12 +1,25 @@
...
+};
 
 /// The number of limbs in [Secp256k1AffinePoint].
 pub const N: usize = 16;
 
 /// An affine point on the Secp256k1 curve.
 #[derive(Copy, Clone)]
 #[repr(align(4))]
-pub struct Secp256k1AffinePoint(pub [u32; N]);
+pub struct Secp256k1AffinePoint(pub WeierstrassPoint<N>);
```
tamirhemo, line 12: nit: rename to Secp256k1Point?

tamirhemo (approved):

PR #1549, opened 2024-09-26
chore: replace `raw()` with `bytes()` in fibonacci example

xJonathanLEI:
Using methods that return vectors or slices is more idiomatic.


mattstam (approved):

mattstam:
matches up with https://github.com/succinctlabs/sp1-project-template/blob/main/script/src/bin/evm.rs#L99 so lgtm

PR #1563, opened 2024-09-30
some changes

jtguibas:


.github/workflows/suite.yml
```diff
@@ -0,0 +1,69 @@
...
+  merge_group:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-x86:
+    name: ${{ matrix.s3_folder }} (${{ matrix.mode }})
+    runs-on:
```
yourbuddyconner, line 22: To handle the GPU runner, something like this: 
```
runs-on: ${{ matrix.mode == 'cuda' && 'nvidia-gpu-linux-x64' || fromJSON('["runs-on", "runner=64cpu-linux-x64", "spot=false", "run-id=${{ github.run_id }}"]') }}

```

PR #1608, opened 2024-10-05
chore(curves): memoize EC constants, fix curve tests

nhtyy:
Added memoization to some curve parameters to avoid string parsing more than once during runtime

The tests in `crates/curves` also werent working so I enabled the needed feature in dev-deps

### DIFF
--- a/crates/curves/Cargo.toml
+++ b/crates/curves/Cargo.toml
@@ -30,3 +30,4 @@ itertools = "0.13.0"
 
 [dev-dependencies]
 rand = "0.8.5"
+num = { version = "0.4.3", features = ["rand"] }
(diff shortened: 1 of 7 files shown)


jtguibas:
Did you add this to increase performance of the elliptic curve precompiles?

nhtyy:
Ya, it doenst look like some of these functions like `generator()` are used much, but modulus does at least seem to be used from `FieldParameters` in a few places

Some moduli are parsed as strings (overriding the default of behavior of `FieldParameters`) but I didnt want to change too much

ctian1:
@nhtyy Could you use `lazy_static` instead of defining a new macro? Otherwise seems like a good addition, though it's not clear how much effect it really has on performance

PR #1652, opened 2024-10-16
docs: update onchain usage

mattstam:
Adds instructions for deploying to new chains, as well as finding your program vkey.

### DIFF
--- a/book/onchain-verification/contract-addresses.md
+++ b/book/onchain-verification/contract-addresses.md
@@ -1,6 +1,8 @@
 # Contract Addresses
 
-To verify SP1 proofs on-chain, we recommend using our deployed verifier gateways. For the chains listed below, an [SP1VerifierGateway](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1VerifierGateway.sol) can automatically route your SP1 proof to the correct verifier based on the SP1 version.
+To verify SP1 proofs on-chain, we recommend using our deployed canonical verifier gateways. The [SP1VerifierGateway](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1VerifierGateway.sol) will automatically route your SP1 proof to the correct verifier based on the SP1 version.
+
+## Canonical Verifier Gateways
 
 | Chain ID | Chain            | Gateway                                                                                                                                 |
 | -------- | ---------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
@@ -24,7 +26,7 @@ support it with
 If a verifier for an SP1 version has an issue, the route will be frozen with
 [freezeRoute()](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1VerifierGateway.sol#L71).
 
-## Deploying to a new chain
+## Deploying to other Chains
 
 In the case that you need to use a chain that is not listed above, you can deploy your own
 verifier contract by following the instructions in the
@@ -38,6 +40,8 @@ interface](./contract-usage.md#isp1verifier-interface), you can choose to either
 * Deploy just the `SP1Verifier` contract that you want to use. Then point to the `SP1Verifier`
   address in
   your contracts.
+
+If you want support for a canonical verifier on your chain, contact us [here](https://t.me/+AzG4ws-kD24yMGYx). We often deploy canonical verifiers on new chains if there's enough demand.
 
 ## ISP1Verifier Interface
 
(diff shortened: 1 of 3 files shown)


book/onchain-verification/contract-addresses.md
```diff
@@ -16,9 +16,28 @@ To verify SP1 proofs on-chain, we recommend using our deployed verifier gateways
 | 534351   | Scroll Sepolia   | [0x3B6041173B80E77f038f3F2C0f9744f04837185e](https://sepolia.scrollscan.com/address/0x3B6041173B80E77f038f3F2C0f9744f04837185e)         |
 | 534352   | Scroll           | [0x3B6041173B80E77f038f3F2C0f9744f04837185e](https://scrollscan.com/address/0x3B6041173B80E77f038f3F2C0f9744f04837185e)                 |
 
-A complete reference for all of the SP1Verifier contract addresses can be also be found in the [SP1 Contracts Repo](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/deployments).
+A complete reference for all of the `SP1Verifier` contract addresses can be also be found in the [SP1 Contracts Repo](https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/deployments).
```
ratankaliani, line 19: nit: We don't need to re-link here. IMO, we should have one source of truth (the SP1 docs) that holds all of this. We should remove this link.

book/onchain-verification/contract-addresses.md
ratankaliani, line 1: Nit: Add new section: 

## Canonical Verifiers

## Deploying to other chains
---Instructions---

If you want support for a canonical verifier on your chain, contact us [here]. We often deploy canonical verifiers on new chains if there's enough demand.

book/onchain-verification/solidity-sdk.md
ratankaliani, line 1: On L40, we should link to docs.succinct.xyz/contract-addresses, rather than the `deployments` JSON files, as the contract addresses table is more clear than chain id numbers.

book/onchain-verification/contract-addresses.md
mattstam: If we want to have one source of truth, then we should just have the sp1-contracts `deployments` dir and remove this table from the docs. The docs will *always* lag behind the deployments json.

book/onchain-verification/solidity-sdk.md
```diff
@@ -64,7 +64,14 @@ contract Fibonacci {
...
 
-For more details on the contracts, refer to the [sp1-contracts](https://github.com/succinctlabs/sp1-contracts) repo.
+### Finding your program vkey
+
+The program vkey (`fibonacciProgramVKey` in the example above) is passed into the `ISP1Verifer` along with the public values and proof bytes. You
+can find your program vkey by going through the following steps:
+
+1. Find what version of SP1 crates you are using.
+2. Use the version from step to run this command: `sp1up --version <version>`
+3. Use the vkey command to get the program vkey: `cargo prove vkey -elf <path/to/elf>`
```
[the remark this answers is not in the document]
mattstam: That's why step 1 exists, so that they ensure it matches

book/onchain-verification/solidity-sdk.md
mattstam: updated

book/onchain-verification/contract-addresses.md
mattstam: added

book/onchain-verification/contract-addresses.md
mattstam: though it seems most reasonable to have both and just link to the deployments as a backup in the worst case that a user needs more information (e.g. a specific SP1Verifier address)

ratankaliani: Sure, then you can say: "The most up-to-date reference for the specific versions supported on each chain can be found in the [deployments] on the SP1 contracts repository. On mainnets, only official versioned releases are supported. 

We should always strive to keep the public-facing docs up to date. We can just make it a part of the deployment process as something that must be added. E.g. When you add a new chain, make a PR to add it to the SP1 docs.

book/onchain-verification/solidity-sdk.md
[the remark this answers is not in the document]
ratankaliani: Got it, still think that we should provide the two different ways that users can get the vkey.

Using `sp1up --version <version>` should return an error message when the version requested does not have published artifacts, as it currently doesn't work for `3.0.0-rc1` and `3.0.0-rc2`, because we never published artifacts for those.

Related: https://linear.app/succinct/issue/GRO-222/sp1up-verison-version-error-message

book/onchain-verification/contract-addresses.md
mattstam: ok sg

ratankaliani (approved):
I'd add two sub-sections in the getting started documnetation for getting the vkey

1. Generate using `cargo prove` CLI
2. Using `sp1-sdk`

Small nit on the docs in the contract addresses, otherwise LGTM.

PR #1659, opened 2024-10-17
fix: additional guards on entrypoint macro

sinasab:
These additional checks ensure attempting to run an sp1 program on a non-zkvm host will silently noop instead of error.


yuwen01 (approved):
I bumped the zkvm version after this change LGTM

PR #1661, opened 2024-10-17
fix(entrypoint): tests fail on program crates

nhtyy:
- `cargo test` usually creates a new "fn main" to be the entry point for running tests

- when we use `!#[no_main]` the test harness has unexpected behavior and just calls our main function

- you cant just not have a main symbol because rustc will complain

- compilation is currently broken in the template repo, and was fixed here https://github.com/succinctlabs/sp1/commit/1c1f84dd389b42b9743bff840e039d347c54c958

- if were not targeting the zkvm, we return an empty main function that has return values that the harness will expect

- todo: ux? We should add a log somewhere if a user does try to make tests in the program crate

### DIFF
--- a/crates/zkvm/entrypoint/src/lib.rs
+++ b/crates/zkvm/entrypoint/src/lib.rs
@@ -91,14 +91,12 @@ mod zkvm {
 #[macro_export]
 macro_rules! entrypoint {
     ($path:path) => {
-        const ZKVM_ENTRY: fn() = $path;
-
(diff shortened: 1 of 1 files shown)


nhtyy:
amended to reduce diff in lock

yuwen01:
Having some difficulty running ci tests from external contributor, for now I'll just finish up this PR here
https://github.com/succinctlabs/sp1/pull/1687

PR #1663, opened 2024-10-17
perf: reduce bit decompositions by using byte lookups

erabinov:


crates/core/machine/src/cpu/columns/mod.rs
```diff
@@ -101,6 +101,8 @@ pub struct CpuCols<T: Copy> {
     /// memory opcodes (i.e. LB, LH, LW, LBU, and LHU).
     pub unsigned_mem_val: Word<T>,
 
+    pub unsigned_mem_val_superposition: T,
```
tamirhemo, line 104: Where is this used?

tamirhemo (approved):
LGTM

PR #1681, opened 2024-10-19
Updating missing arguments in SP1 Quickstarter

Uttam-Singhh:
The command `cargo prove new fibonacci` was missing <--bare|--evm> arguments

image


yuwen01 (changes requested):
Could you briefly explain the difference between bare and evm? Besides that LGTM

Uttam-Singhh:
> Could you briefly explain the difference between bare and evm? Besides that LGTM

done

yuwen01 (approved):
LGTM

PR #1689, opened 2024-10-23
test: add network-v2 to perf test suite

mattstam:
Use `perf` crate using network-v2. Test results: https://github.com/succinctlabs/sp1/actions/runs/11543552296/job/32179604006

image

Note: because of the features, either only `network` or `network-v2` can be tested at once in the suite - but we should be okay to just do `network-v2` as there are other stress tests that work on the old network.

CLOUD-736


.github/workflows/suite.yml
```diff
@@ -71,44 +77,52 @@ jobs:
...
-        toolchain: 1.81.0
-        args: --release -p sp1-perf -- --program workdir/program.bin --stdin workdir/stdin.bin --mode cuda
-      env:
-        RUST_LOG: debug
-        RUSTFLAGS: -Copt-level=3 -Ctarget-cpu=native
-        RUST_BACKTRACE: 1
-        SP1_PROVER: cuda
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
```
mattstam, line 92: sorry for diff my auto formatter is aggressive

.github/workflows/suite.yml
```diff
@@ -117,43 +131,48 @@ jobs:
...
+        with:
+          command: clean
+          toolchain: 1.81.0
+
+      - name: Run sp1-perf
+        uses: actions-rs/cargo@v1
+        with:
+          command: run
+          toolchain: 1.81.0
+          args: --release -p sp1-perf --features "native-gnark,network-v2" -- --program workdir/program.bin --stdin workdir/stdin.bin --mode network
```
mattstam, line 172: the downside with this approach is that we no longer test network-v1 in ci, but maybe that is fine? @jtguibas

jtguibas (approved):

PR #1690, opened 2024-10-23
feat(docker-build): pass additional workspace directory

prajwolrg:
Currently, docker only mounts the guest program workspace. However, mounting a different workspace can be useful. This PR allows passing an additional `workspace_directory` as `BuildArgs`.

### DIFF
--- a/crates/build/src/command/docker.rs
+++ b/crates/build/src/command/docker.rs
@@ -28,7 +28,19 @@ pub(crate) fn create_docker_command(
         .expect("Failed to canonicalize program directory")
         .try_into()
         .unwrap();
-    let workspace_root = &program_metadata.workspace_root;
+
+    let workspace_root: &Utf8PathBuf = &args
+        .workspace_directory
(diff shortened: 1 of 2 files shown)


prajwolrg:
> Thank you for your patience on this PR!
> 
> Happy to take this over, but I think it be suitable to instead override the program dir? Therefore leading to a different workspace metadata?
> 
> See here:
> 
> https://github.com/succinctlabs/sp1/blob/f9640e8649b3d48b95b55fb2ae7ab7fefb424420/crates/build/src/build.rs#L30-L43
> 
> Currently, this is not exposed as an option via the CLI? Also likely this crate gets cleaned up more in the future so will keep this in mind going forward.

In our usecase, we had multiple guest programs in a single workspace. So, we used this approach. We didn't find it exposed via CLI. There seems to have been significant update on build from the last time I looked into this. Happy to help you if you take it over.

nhtyy:
Closed by #1975

PR #1692, opened 2024-10-23
feat: executor optimizations

ctian1:


crates/prover/src/shapes.rs
```diff
@@ -288,6 +312,53 @@ impl<C: SP1ProverComponents> SP1Prover<C> {
             }
         }
     }
+
```
jtguibas, line 315: do we need this code?

PR #1695, opened 2024-10-24
fix: patch bls and bn

yuwen01:
Before, the sp1 program imported the bn and bls patches directly, like `bn = { git = "sp1-patches/bn" }`.

Now, they import the unpatched version, and the patch is applied from examples/cargo.toml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ crates/prover/powersOfTau28_hez_final.ptau
 crates/prover/semaphore-gnark-11
 crates/prover/trusted-setup
 crates/prover/vk
+
+# Example legacy elf
+examples/elf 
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -33,4 +33,7 @@ crates/prover/pk
 crates/prover/powersOfTau28_hez_final.ptau
 crates/prover/semaphore-gnark-11
 crates/prover/trusted-setup
-crates/prover/vk
\ No newline at end of file
+crates/prover/vk
+
+# Example legacy elf
+examples/elf
```
yuwen01, line 39: When jonathan made !include_elf, he also manually copied the elf from /target to the workspace root, to maintain backward compatibility with the old !include_bytes method. Let's gitignore that elf in examples.

yuwen01: Should we gitignore the cargo.lock in examples too?

examples/Cargo.toml
```diff
@@ -71,3 +71,5 @@ sha2-v0-10-6 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", packa
 sha2-v0-9-9 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha2", branch = "patch-sha2-v0.9.9" }
 sha2-v0-9-8 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha2", branch = "patch-sha2-v0.9.8" }
 tiny-keccak = { git = "https://github.com/sp1-patches/tiny-keccak", branch = "patch-v2.0.2" }
+substrate-bn = { git = "https://github.com/sp1-patches/bn", rev = "43d854d45b5727b1ff2b9f346d728e785bb8395c"}
```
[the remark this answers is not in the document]
yuwen01: agreed i need some kind of admin access to work on that today

PR #1697, opened 2024-10-25
docs: Update example commands from cd script to cd ../script

collinlowenburg:
Under executing the program, you have to switch directory from program to script. The example command does not work as a direct copy/paste because the user is in the /program dir and needs to go up one.  Just changing from "cd script"  to "cd ../script" would make life easier for newbies. 

And for the step of generating a proof, the "cd script" is repeated which fails when copy/pasted. Again "cd ../script" will work nicely here.


yuwen01 (approved):
LGTM

PR #1698, opened 2024-10-25
docs: Add --evm template to the command.

collinlowenburg:
On the first step in Option 1, I got an error which is easily fixed, and an update to the docs would prevent others from the same issue.

 % cargo prove new fibonacci

error: the following required arguments were not provided:
  <--bare|--evm>

Usage: cargo prove new <--bare|--evm> <NAME>

I propose to add the argument --evm and update the description with this line: 

This command uses the `evm` template which includes Solidity smart contracts for onchain integration.


yuwen01:
Resolved here https://github.com/succinctlabs/sp1/pull/1681

PR #1699, opened 2024-10-26
feat: get rid of all assignment ops in airs

jtguibas:


adr1anh:
Out of curiosity, what's the reasoning behind this change? Are arithmetic operators less efficient in some situations?

jtguibas:
@adr1anh I noticed that the it helps both recursion / the gpu prover if the AIR definition assumes write-once memory.

erabinov (approved):
LGTM!

PR #1707, opened 2024-10-29
perf: implement select precompile

rkm0959:


crates/recursion/circuit/src/hash.rs
```diff
@@ -129,14 +129,20 @@ impl<C: CircuitConfig<F = BabyBear, Bit = Felt<BabyBear>>> FieldHasherVariable<C
...
-            core::array::from_fn(|_| selected.next().expect(err_msg)),
-            core::array::from_fn(|_| selected.next().expect(err_msg)),
-        ];
-        assert_eq!(selected.next(), None, "{}", err_msg);
-        ret
+        let result0: [Felt<BabyBear>; DIGEST_SIZE] = core::array::from_fn(|_| builder.uninit());
+        let result1: [Felt<BabyBear>; DIGEST_SIZE] = core::array::from_fn(|_| builder.uninit());
+
+        (0..DIGEST_SIZE).for_each(|i| {
+            builder.push_op(DslIr::SelectDigest(
```
erabinov, line 136: I think `Select` is probably a better name for this opcode, since you're doing one per element of the digest.

PR #1708, opened 2024-10-29
feat: add scalar_mult function to main.rs for Bn254 point multiplication

rebustron:
scalar_mult function: This function takes a reference to a point (Bn254Point) and a scalar (32-byte array) as inputs. It performs scalar multiplication, returning a new Bn254Point that represents the resulting point after multiplication.

Bn254Scalar Conversion: The scalar byte array is converted into a Bn254Scalar using from_bytes, enabling multiplication with the point.

Usage: This function is helpful when performing cryptographic operations that require scalar multiplication of points, such as in zero-knowledge proof computations.

### DIFF
--- a/tests/bn254-add/src/main.rs
+++ b/tests/bn254-add/src/main.rs
@@ -2,6 +2,7 @@
 
 use sp1_curves::params::FieldParameters;
 use sp1_lib::bn254::Bn254Point;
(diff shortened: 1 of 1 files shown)


nhtyy:
thank you but closing this because we dont need test helpers

PR #1716, opened 2024-10-30
feat: two shapes in recursion

erabinov:


crates/prover/src/lib.rs
```diff
@@ -355,40 +368,17 @@ impl<C: SP1ProverComponents> SP1Prover<C> {
...
-                let mut program = compiler.compile(operations);
-                if let Some(recursion_shape_config) = &self.recursion_shape_config {
-                    recursion_shape_config.fix_shape(&mut program);
-                }
-                let program = Arc::new(program);
-                compiler_span.exit();
-                program
-            })
-            .clone()
+        self.compress_programs.get(&input.shape()).map(Clone::clone).unwrap_or_else(|| {
```
tamirhemo, line 371: Don't you want to panic here? seems like this should not be allowed? 

Unless you want to take care of the non-existing shape config case, but in this case I would perhaps match on this option instead

erabinov: Good point

PR #1723, opened 2024-10-31
Return the intermediate STARK when requesting groth16

S1nus:


nhtyy:
Thank you its been a while and im going to close this as not planned for now, what was the motivation here?

PR #1725, opened 2024-10-31
chore: update network-v2 proto

mattstam:
and set a different default cycle limit

### DIFF
--- a/crates/sdk/src/network-v2/prover.rs
+++ b/crates/sdk/src/network-v2/prover.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use crate::{
-    network_v2::client::{NetworkClient, DEFAULT_PROVER_NETWORK_RPC},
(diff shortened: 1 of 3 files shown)


crates/sdk/src/network-v2/prover.rs
```diff
@@ -99,10 +99,6 @@ impl NetworkProver {
         let request_id_hex = "0x".to_string() + &hex::encode(request_id.clone());
         log::info!("Created request {} in transaction {}", request_id_hex, tx_hash_hex);
 
-        if NetworkClient::rpc_url() == DEFAULT_PROVER_NETWORK_RPC {
-            log::info!("View in explorer: https://explorer-v2.succinct.xyz/{}", request_id_hex);
```
mattstam, line 103: will add this back when explorer is done, better than giving out dead link

jtguibas (approved):

PR #1736, opened 2024-11-01
chore: add MSRV of 1.79

ratankaliani:
Users could not compile `sp1-core-executor` on Rust <1.78 because of this feature https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#extending-automatic-temporary-lifetime-extension.


yuwen01:
wait i thought the msrv was 1.81

ratankaliani:
@yuwen01 Our Rust toolchain has a **maximum** supported Rust version of 1.81

Because of our use of [this feature](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#extending-automatic-temporary-lifetime-extension) in `sp1-core-executor`, we need to set the **minimum** rust version of our crate to 1.79, otherwise users trying to run `sp1-sdk` with a lower version will encounter errors like this.

image

ctian1 (approved):
We should add rust-version in Cargo.toml too

PR #1739, opened 2024-11-01
perf: build time GRO-146

yuwen01:
1. Run cargo machete and cargo update on the lock.
Got rid of neon feature, since we're not using blake3 hash anymore.
Organize dependencies in Cargo.toml
2. Remove ethers from sp1-sdk, and replace it with alloy-signers

Tested [6fe72e8](https://github.com/succinctlabs/sp1/pull/1739/commits/6fe72e8e2292860c3803d5fb0844c79845e8e0fc) sanity check with network prover


rust_build_profiler.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
+import subprocess
```
yuwen01, line 2: remove this

.github/workflows/pr.yml
```diff
@@ -118,15 +118,6 @@ jobs:
         env:
           CARGO_INCREMENTAL: 1
 
-      - name: Add wasm target
```
yuwen01, line 121: this is no longer a priority since we have the verifier crate

crates/core/machine/Cargo.toml
```diff
@@ -19,7 +19,6 @@ p3-air = { workspace = true }
 p3-baby-bear = { workspace = true }
 p3-challenger = { workspace = true }
 p3-field = { workspace = true }
-p3-blake3 = { workspace = true }
```
yuwen01, line 22: this was unused

jtguibas (approved):

PR #1750, opened 2024-11-05
feat: multi-crate ELF building support

leruaa:
I renamed `--binary` to `--binaries` as it's more accurate now, but I'm not sure as it's a breaking change.

Also, maybe I can do this:

https://github.com/succinctlabs/sp1/blob/97bb83dbec5fbfef56ade3d08e61e561eff5d53d/crates/build/src/utils.rs#L21-L23


crates/core/machine/src/riscv/shape.rs
```diff
@@ -703,7 +703,7 @@ impl<F: PrimeField32> Default for CoreShapeConfig<F> {
     }
 }
 
-#[cfg(any(test, feature = "programs"))]
+#[cfg(test)]
```
yuwen01, line 706: Why did we turn this off for the programs feature?

crates/core/machine/src/syscall/precompiles/README.md
```diff
@@ -208,8 +208,7 @@ Then implement the tests and run `cargo prove build` to generate an ELF file.
 ### Include the ELF File in `program.rs`
 In your main SP1 project, include the generated ELF file by updating `program.rs`.
```
yuwen01, line 209: I think this step is no longer necessary

yuwen01, line 209: I think this step is replaced with updating the test artifacts cargo.toml

crates/test-artifacts/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+#![warn(clippy::pedantic)]
+
+use sp1_build::include_elf;
+/// Compiled test programs. TODO elaborate.
```
yuwen01, line 4: finish this todo

.github/workflows/pr.yml
```diff
@@ -110,6 +128,14 @@ jobs:
         env:
           CARGO_INCREMENTAL: 1
 
+      - name: Build test-artifacts
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
```
yuwen01, line 134: nit: check test-artifacts

crates/core/machine/src/riscv/shape.rs
yuwen01: Oh i see the programs feature is replaced with the test artifacts crate?

yuwen01 (commented):
Are test artifacts automatically built when running tests? Is it feasible to do this in CI now, or even with a build.rs in the `test-artifacts` crate?

crates/core/machine/src/riscv/shape.rs
leruaa: Yes exactly

crates/core/machine/src/syscall/precompiles/README.md
leruaa: Oh yeah it's `crates/test-artifacts/src/lib.rs` now

yuwen01 (approved):
LGTM

just for my own understanding, you use `--binaries` when you need to build multiple elfs for multiple sp1 programs within the same package, and you use `--packages` when you need to build multiple packages? 

Does that mean code like this is now deprecated (from RSP)?

```
use sp1_build::build_program;

fn main() {
    build_program("../client-eth");
    build_program("../client-op");
    build_program("../client-linea");
}
```

And you'd instead do something like `build_program_with_args("path_to_workspace_root", BuildArgs{packages=vec!["client-linea", "client-op", "client-eth"], ..Default}` , which is better than the previous approach, because it's parallelized.

leruaa:
@yuwen01 Yes exactly. And if the 3 crates in your example above are the only ones in theirs workspace, you don't even neet to specify them as `--packages`, it will build everything. `--packages` is if you want to filter.

PR #1753, opened 2024-11-05
feat: plonk verification with `.bytes()`

yuwen01:
Verifies plonk proofs using .bytes() instead of .raw_with_checksum() in the verifier crate.

Preparing for crates.io release with 3.0.0 sdk -- so this PR also updates the verifier readme to point to the package readme instead of the workspace readme, does some better error handling, and adds some comments


crates/verifier/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "sp1-verifier"
 description = "Verifier for SP1 Groth16 and Plonk proofs."
-readme = "../../README.md"
+readme = "README.md"
```
yuwen01, line 4: on one hand, this crate is pretty independent, so it deserves its own readme.

on the other hand, i'm gonna write some docs in the mdbook, so i can just have this point to the workspace readme maybe?

jtguibas: Doesn't matter

jtguibas (approved):

PR #1757, opened 2024-11-05
feat: cross compile amd64 Linux binaries with Musl

### LINKED ISSUE #1355: Segmentation Fault in sp1up, while attempting install.
I installed `sp1up` following the installation instructions, sourced the updated `.bashrc` and then run `sp1up' it says it's downloading `cargo-prove` and `rust toochain` which appear to work but then it fails with this error

```
/home/dominic/.sp1/bin/sp1up: line 337: 26591 Segmentation fault      (core dumped) "$@"
sp1up: command failed: /home/dominic/.sp1/bin/cargo-prove prove install-toolchain
``` 

this is on linux, if that helps

leruaa:
The seg fault in the related issue seems to be caused by [a bug in glibc](https://github.com/rust-lang/rust/issues/100711#issuecomment-1219293304). The solution would be to use musl instead, since anyways [static linking with glibc is discouraged](https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged).

In this PR, I modified the release CI to use a docker image called [muslrust](https://github.com/clux/muslrust) to cross compile the SP1 CLI amd64 Linux binaries with musl instead of glibc.

Closes #1355


yuwen01:
are we going to have to wait till the next release to ship new binaries, or can we just change some files now?

leruaa:
@yuwen01 same as before, I haven't changed the event when the workflow is ran:

https://github.com/succinctlabs/sp1/blob/2d73382d149001b39ea3d4e7b9e07bddea3fc012/.github/workflows/release.yml#L5-L10

Is it needed?

jtguibas (approved):

PR #1760, opened 2024-11-06
feat: Add `skip_deferred_proof_verification`

ratankaliani:
Systems such as `op-succinct` require an additional flag to skip deferred proof verification when generating mock proofs, because their execution relies on the verification of compressed proofs (deferred proofs) which are invalid in mock mode.

### DIFF
--- a/crates/sdk/src/action.rs
+++ b/crates/sdk/src/action.rs
@@ -88,7 +88,6 @@ pub struct Prove<'a> {
     core_opts: SP1CoreOpts,
     recursion_opts: SP1CoreOpts,
     timeout: Option<Duration>,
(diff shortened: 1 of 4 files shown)


crates/sdk/src/action.rs
```diff
@@ -217,4 +227,10 @@ impl<'a> Prove<'a> {
         self.timeout = Some(timeout);
         self
     }
+
+    /// Set the skip deferred proof verification flag.
```
yuwen01, line 231: can this go in the context builder instead?

ratankaliani: Moved it.

yuwen01 (approved):

PR #1763, opened 2024-11-06
chore: lockfile cleanup

umadayal:
Removes the Cargo.lock files in some of the tests that were accidentally added with the secp256r1 commit along with the merkle_tree.bin file. Updated examples since the ECDSA signatures patch was updated to take in a  specific commit instead of branch


ratankaliani (approved):
LGTM!

PR #1787, opened 2024-11-13
feat: stop printing unused syscalls in execution report

leruaa:
## Motivation

Printing unused syscalls just clutters the execution report

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/machine/src/utils/prove.rs
+++ b/crates/core/machine/src/utils/prove.rs
@@ -29,7 +29,10 @@ use crate::{
     riscv::cost::CostEstimator,
     utils::{chunk_vec, concurrency::TurnBasedSync},
 };
-use sp1_core_executor::{events::sorted_table_lines, ExecutionState};
+use sp1_core_executor::{
+    events::{format_table_line, sorted_table_lines},
+    ExecutionState,
+};
 use sp1_primitives::io::SP1PublicValues;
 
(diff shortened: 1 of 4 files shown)


crates/core/machine/src/utils/prove.rs
```diff
@@ -674,12 +677,19 @@ where
         // Print the opcode and syscall count tables like `du`: sorted by count (descending) and
         // with the count in the first column.
         tracing::info!("execution report (opcode counts):");
-        for line in sorted_table_lines(report_aggregate.opcode_counts.as_ref()) {
-            tracing::info!("  {line}");
+        let (width, lines) = sorted_table_lines(report_aggregate.opcode_counts.as_ref());
```
yuwen01, line 680: why is this logic duplicated from the executor? i notice it was before your changes as well -- just wondering

yuwen01 (approved):

crates/core/machine/src/utils/prove.rs
leruaa: Indeed I see no reason, I'll see to refactor

leruaa: Mmm, In report.rs we have writeln!, in prove.rs it's tracing::info!

PR #1790, opened 2024-11-14
feat: clean up NetworkProver environment variables

leruaa:
## Motivation

We need to remove environment variables as they make SP1 difficult to use in a prod environment.

## Solution

I added a `NetworkProverClientBuilder` builder that handle the equivalent of the following environment variables:

* `SKIP_SIMULATION`
* `SP1_PRIVATE_KEY`
* `PROVER_NETWORK_RPC`

The environment variables are still used in `ProverClient::new()` for backward compatibility.

I only added a builder for the network builder as there is nothing to build for other provers.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -13,6 +13,8 @@ pub mod network;
 #[cfg(feature = "network-v2")]
 #[path = "network-v2/mod.rs"]
 pub mod network_v2;
+use std::env;
+
 #[cfg(feature = "network")]
 pub use crate::network::prover::NetworkProver as NetworkProverV1;
 #[cfg(feature = "network-v2")]
@@ -30,8 +32,6 @@ use cfg_if::cfg_if;
 pub use proof::*;
 pub use provers::SP1VerificationError;
 use sp1_prover::components::DefaultProverComponents;
-
-use std::env;
 
 #[cfg(any(feature = "network", feature = "network-v2"))]
 use {std::future::Future, tokio::task::block_in_place};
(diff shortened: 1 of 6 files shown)


crates/sdk/src/lib.rs
```diff
@@ -85,14 +86,20 @@ impl ProverClient {
                 }
```
nhtyy, line 86: I think we should just get rid of the new function entirely. The only way its useable IMO is if you read the code, which is not ideal

nhtyy: cc @yuwen01

crates/sdk/src/network-v2/prover.rs
```diff
@@ -30,23 +28,22 @@ const DEFAULT_CYCLE_LIMIT: u64 = 100_000_000;
...
         log::info!("Client circuit version: {}", version);
         let local_prover = CpuProver::new();
-        let client = NetworkClient::new(private_key);
-        Self { client, local_prover }
+        let client = NetworkClient::new(private_key, rpc_url);
+        Self { client, local_prover, skip_simulation }
+    }
+
+    /// Creates a new network prover builder. See [`NetworkProverBuilder`] for more details.
+    pub fn builder() -> NetworkProverBuilder {
```
yuwen01, line 45: Is this used?

crates/sdk/src/lib.rs
```diff
@@ -283,6 +319,120 @@ impl Default for ProverClient {
...
+            }
+            ProverMode::Mock => ProverClient::mock(),
+        }
+    }
+}
+
+/// Builder type for network prover.
+#[cfg(any(feature = "network", feature = "network-v2"))]
+#[derive(Debug, Default)]
+pub struct NetworkProverBuilder {
```
yuwen01, line 394: When does someone use this struct?

yuwen01 (commented):
Looks good I think I'm just missing some context

crates/sdk/src/lib.rs
leruaa: I've discussed that with Chris, the thing is `ProverClient::new()` is used in the default SP1 template, so removing it break a lot of people.

PR #1794, opened 2024-11-15
feat: allow to view the verification key by providing the program name

leruaa:
## Motivation

Now that the ELF is not in the `elf` folder it harder to specify the path. 

## Solution

Adding a `--program` arg that just need the crate name.

Bonus: If there are multiple targets in the program, they will all be printed.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


yuwen01 (approved):
LGTM

PR #1796, opened 2024-11-18
feat: move a few logs to debug

leruaa:
## Motivation

1. Execute and prove_core end up printing out the same stuff twice, like cycle counts and syscall invocations. It's confusing for users to figure out if there should be a difference between these two outputs

2. A bunch of stuff about trace area for every table gets printed out, which users don't care about and is just confusing.

3. This refers more to the cuda prover I think: 
  We could change the default to be RUST_LOG=sp1_cuda=off,info or something similarbi

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/machine/src/utils/prove.rs
+++ b/crates/core/machine/src/utils/prove.rs
@@ -244,11 +244,11 @@ where
                                         shape_config,
                                     )
                                 });
-                            log::info!("generated {} records", records.len());
+                            tracing::debug!("generated {} records", records.len());
                             reset_seek(&mut checkpoint);
 
                             // Wait for our turn to update the state.
-                            log::info!("waiting for turn {}", index);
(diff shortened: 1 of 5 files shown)


crates/core/machine/src/utils/prove.rs
yuwen01, line 1: What is the difference between log::debug and tracing::debug? Are they both affected by the `RUST_LOG` environment variable? 

Based on my understanding we should just use tracing for everything. There's more performance overhead, but the output is a lot more detailed and useful, which is what you'd want from a debug message. Do you agree?

yuwen01 (commented):
Looks pretty good -- just not sure what the difference is between tracing::info and log::info.

yuwen01 (approved):

PR #1800, opened 2024-11-19
feat: explicit `skip_simulation` flag

ratankaliani:
## Motivation

User flags that are used to modify the behavior of the runtime should use a builder pattern rather than enshrined environment variables.

Note: This is not backwards compatible because you must set the `skip_simulation` flag on the `NetworkProver` directly. Now `ProverClient`'s will not have access to `skip_simulation`.

## Solution

Modify the `SKIP_SIMULATION` environment variable to instead by explicitly set on the `NetworkProver`.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [X] Breaking changes


ratankaliani:
Superseded by https://github.com/succinctlabs/sp1/pull/1790

PR #1805, opened 2024-11-20
N/sdk feature flags

nhtyy:
- introduces `conflicting` crate to throw compile time errors if you try to enable `network` and `network-v2`
- adds more feature gates 
- deprecates `ProverClient::new`
- fixes build error on `--no-default-features`


crates/sdk/src/network/prover.rs
```diff
@@ -39,11 +38,6 @@ impl NetworkProver {
         Self { client: NetworkClient::new(private_key, rpc_url), local_prover, skip_simulation }
     }
 
-    /// Creates a new network prover builder. See [`NetworkProverBuilder`] for more details.
```
yuwen01, line 42: I think either @ctian1 or @ratankaliani needs this I don't remember who

crates/prover/Cargo.toml
```diff
@@ -82,6 +82,8 @@ required-features = ["export-tests"]
```
yuwen01, line 82: I don't like how the prover crate now has a "cuda" and "network feature that don't do anything at all. On second thought, it's even kind of weird that the ProverMode lives in the prover crate in the first place. I feel like the prover crate should pretty much solely be concerned with generating different kinds of proofs on-device, and not really think about anything else. 

cc @ctian1 @ratankaliani for second opinions

@leruaa why did you need to move the provermode here in the first place?

crates/sdk/src/install.rs
```diff
@@ -42,18 +41,18 @@ pub fn try_install_circuit_artifacts(artifacts_type: &str) -> PathBuf {
...
-                println!(
-                    "[sp1] {} circuit artifacts for version {} do not exist at {}. downloading...",
-                    artifacts_type,
-                    SP1_CIRCUIT_VERSION,
-                    build_dir.display()
-                );
-                install_circuit_artifacts(build_dir.clone(), artifacts_type);
-            }
-        }
+        #[cfg(any(feature = "network", feature = "network-v2"))]
```
yuwen01, line 44: note: no action needed, but I think it's kind of weird how the `network` feature is, in the rest of the sdk, responsible for the network prover, but now it's responsible for *accessing* the network for getting artifacts.

crates/sdk/src/lib.rs
```diff
@@ -185,18 +189,18 @@ impl ProverClient {
     ///
     /// let client = ProverClient::network(private_key, rpc_url, skip_simulation);
     /// ```
+    #[cfg(any(feature = "network", feature = "network-v2"))]
```
yuwen01, line 192: We need to fix these CI failures before merging. Maybe we need to check sp1-sdk using both the network and network-v2 features ... 

Honestly, if we deprecate network v1 in the next couple months maybe we should just deal with these sins until then.

yuwen01 (commented):
Let's wait for Chris to come back to see if this is a good idea -- he's the one using a lot of the network stuff. Also curious about aurelien's opinion.

crates/sdk/src/install.rs
nhtyy: Ohh youre saying that the artifacts actually shouldn't be gated because all modes mauy need them? 

Not actually sure just kept the original functionality here

PR #1806, opened 2024-11-21
perf: Strip the RISC-V ELF binary

imikushin:
This reduces the size of produced binaries by more than 2x.

## Motivation

The generated ELF binaries are pretty large: the fibonacci example from Quickstart compiles to a 118KB binary. When looking at it with the `file` utility, one can see that it's not stripped.

```
❯ la $SP1_ELF_fibonacci_program
-rwxr-xr-x@ 1 ivan  staff   118K Nov 20 16:14 /Users/ivan/src/succinct/fibonacci/target/elf-compilation/riscv32im-succinct-zkvm-elf/release/fibonacci-program

❯ file $SP1_ELF_fibonacci_program
/Users/ivan/src/succinct/fibonacci/target/elf-compilation/riscv32im-succinct-zkvm-elf/release/fibonacci-program: ELF 32-bit LSB executable, UCB RISC-V, soft-float ABI, version 1 (SYSV), statically linked, not stripped
```

## Solution

Adding "-C", "strip=symbols", "-C", "opt-level=z" to Rust compiler flags reduces the produced ELF binary size by more than 2x.

```
❯ la $SP1_ELF_fibonacci_program
-rwxr-xr-x@ 1 ivan  staff    53K Nov 20 16:20 /Users/ivan/src/succinct/fibonacci/target/elf-compilation/riscv32im-succinct-zkvm-elf/release/fibonacci-program

❯ file $SP1_ELF_fibonacci_program
/Users/ivan/src/succinct/fibonacci/target/elf-compilation/riscv32im-succinct-zkvm-elf/release/fibonacci-program: ELF 32-bit LSB executable, UCB RISC-V, soft-float ABI, version 1 (SYSV), statically linked, stripped
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/build/src/command/utils.rs
+++ b/crates/build/src/command/utils.rs
@@ -50,9 +50,18 @@ pub(crate) fn get_program_build_args(args: &BuildArgs) -> Vec<String> {
 
 /// Rust flags for compilation of C libraries.
 pub(crate) fn get_rust_compiler_flags(args: &BuildArgs) -> String {
-    let rust_flags =
-        ["-C", "passes=loweratomic", "-C", "link-arg=-Ttext=0x00200800", "-C", "panic=abort",
-            "-C", "strip=symbols", "-C", "opt-level=z"];
+    let rust_flags = [
+        "-C",
+        "passes=loweratomic",
+        "-C",
+        "link-arg=-Ttext=0x00200800",
+        "-C",
+        "panic=abort",
+        "-C",
+        "strip=symbols",
+        "-C",
+        "opt-level=z",
+    ];
     let rust_flags: Vec<_> =
         rust_flags.into_iter().chain(args.rustflags.iter().map(String::as_str)).collect();
     rust_flags.join("\x1f")
--- a/crates/build/src/command/utils.rs
+++ b/crates/build/src/command/utils.rs
@@ -51,7 +51,8 @@ pub(crate) fn get_program_build_args(args: &BuildArgs) -> Vec<String> {
 /// Rust flags for compilation of C libraries.
 pub(crate) fn get_rust_compiler_flags(args: &BuildArgs) -> String {
     let rust_flags =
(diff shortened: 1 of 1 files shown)


matthiasgoergens:
What is this supposed to accomplish?

You should look at the amount of data the prover has to handle, and that's not directly related to ELF binary size: as far as I can tell, debug symbols are already ignored by the prover.

imikushin:
@matthiasgoergens This change is simply to make the produced binary size smaller: it seems that I'll need store these binaries, and making them 2x smaller will save 2x the cost — it's that simple. Stripping them of symbols doesn't seem to hurt neither prover nor verifier while cutting unnecessary waste.

matthiasgoergens:
Thanks for the reply.

There are two things happening in your PR:

1. You strip the debug symbols.  You are right that neither prover nor verifier care about those.  It's just a bit annoying for debuggability. (We just had an occasion at my work, where having the debug symbols available proved really useful.)
2. You change the optimisation level to 'z'.  That will have an impact on prover performance.  It can save you a few bytes that the prover doesn't have to load, but it might cost you more VM cycles that need proving.  I would suggest doing benchmarks of your individual guest programs, to determine whether that's a good trade-off.

There might be some circumstances where you want to save every last byte of the binary, even at the cost of debuggability or having extra VM CPU cycles that need proving.  But my guess is that those circumstances are rare enough, that you wouldn't want to enable the flags you suggested by default for everyone under all circumstances.

However, keep in mind that I'm not associated with Succinct or SP1.

Btw, if you want to improve the default build configuration for guest programs, I would suggest enabling link time optimisation.  That alone saved us 7% of cycles in our project.

matthiasgoergens:
@imikushin You bring up a great point!  I might be useful to stick your suggestion (together with an explanation of when you might want to do this) in a README somewhere in https://github.com/succinctlabs/sp1-project-template or as a comment inside the generated `Cargo.toml`.

imikushin:
Closing in favor of succinctlabs/sp1-project-template#42

PR #1818, opened 2024-11-25
chore: all feats, no feats, default feats top level workspace checks

nhtyy:


.github/workflows/pr.yml
```diff
@@ -186,7 +222,7 @@ jobs:
 
   cli:
     name: CLI
-    runs-on: [runs-on, runner=8cpu-linux-x64, "run-id=${{ github.run_id }}"]
+    runs-on: [runs-on, runner=8cpu-linux-x64, disk=medium, "run-id=${{ github.run_id }}"]
```
yuwen01, line 225: just curious what happened here

crates/core/executor/Cargo.toml
```diff
@@ -44,7 +44,7 @@ enum-map = { version = "2.7.3", features = ["serde"] }
 test-artifacts = { workspace = true, optional = true }
```
yuwen01, line 44: i ran into this issue too lol

yuwen01 (approved):
LGTM

I guess we could check the examples with this level of rigor too, but I think that's overkill.

yuwen01:
yea just don't merge until ci passes

.github/workflows/pr.yml
nhtyy: checkout the current failing CI, its out of space on CLI job, maybe we need large here too?

PR #1819, opened 2024-11-25
fix: non absolute build paths are relative to the build manifest

nhtyy:
sp1 build searches for the programs from the current dir

this can have unexpected behaviour if calling from an importing crate in a differnt location

opt in, but deprecating, way to have more reliable program resolution

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -4,6 +4,8 @@ mod utils;
 use build::build_program_internal;
 pub use build::{execute_build_program, generate_elf_paths};
 
+use std::path::Path;
+
 use clap::Parser;
 
 const BUILD_TARGET: &str = "riscv32im-succinct-zkvm-elf";
@@ -110,7 +112,7 @@ impl Default for BuildArgs {
 /// when changes are made to the source code or its dependencies.
 ///
 /// Set the `SP1_SKIP_PROGRAM_BUILD` environment variable to `true` to skip building the program.
-pub fn build_program(path: &str) {
+pub fn build_program(path: impl AsRef<Path>) {
     build_program_internal(path, None)
 }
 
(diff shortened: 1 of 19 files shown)


crates/build/src/lib.rs
```diff
@@ -119,14 +126,56 @@ pub fn build_program(path: &str) {
 ///
 /// # Arguments
 ///
-/// * `path` - A string slice that holds the path to the program directory.
+/// * `path` - A path to the guest program directory
```
yuwen01, line 129: nit punc

crates/build/src/lib.rs
```diff
@@ -119,14 +126,56 @@ pub fn build_program(path: &str) {
...
+/// Build a program at the given path.
+///
+/// # Arguments
+/// * `path` - A path to the guest program directory, if not absolute, assumed to be relative to
+///           the callers manifest directory.
+///
+///   `args` - A [`BuildArgs`] struct that contains various build configuration options.
+///            If not provided, the default options are used.
+#[macro_export]
+macro_rules! build_program_from_path {
```
yuwen01, line 159: wait why does this have to be a macro?

crates/build/src/lib.rs
```diff
@@ -119,14 +126,56 @@ pub fn build_program(path: &str) {
...
+pub fn build_program_with_args(path: impl AsRef<Path>, args: BuildArgs) {
     build_program_internal(path, Some(args))
 }
 
+/// Builds the program with the given arguments if the program at path, or one of its dependencies,
+///
+/// ### Note:
+/// This function is only exposed to support the `build_program_from_path!` macro.
+/// It is not recommended to use this function directly.
+pub fn build_program_with_maybe_args(path: impl AsRef<Path>, args: Option<BuildArgs>) {
```
yuwen01, line 146: i think we should avoid exposing this function if possible, though ngl i'm not really sure how we'd do it.

yuwen01 (commented):
Is there a way to avoid exposing `build_program_with_maybe_args` and the macro?

crates/build/src/lib.rs
nhtyy: Because we need to inline the manifest dir at compile time so the path is always relative from the crate dir

from slack: 
My contrived example of why this is bad is:
Crate A: lives in dir A exports a function build_A_program that uses a relative path to call sp1_build::build_program
Crate B:  lives in dir B imports this function
Crate B: Calls build_A_program with working dir B
this call will then try to look for the path relative to B which is def incorrect

PR #1821, opened 2024-11-26
fix: preprocessed trace bug

ctian1:
## Motivation

ProgramMemory chip has a non-deterministic bug

## Solution

Sort before generating preprocessed trace from HashMap

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


ratankaliani (approved):
LGTM

PR #1822, opened 2024-11-26
fix: Add missing boolean constraints for `is_real` flag

Nashtare:
## Motivation

There seem to be some missing boolean checks on the `is_real` flag for some precompiles.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


rkm0959:
The `is_real` checks are done implicitly outside of the AIR functions. For example, consider `fp.rs`. 

There, `eval_memory_access_slice` is called with `is_real` as `verify_memory_access`. If you go to that function you will see that the first thing done in `eval_memory_access` is that `do_check`, which is the `verify_memory_access` value, is checked to be boolean. The same idea goes for every syscall.

PR #1835, opened 2024-12-03
chore: use &'static str for chip names as much as possible

nhtyy:
Returning a string is an allocation, Ive at least gotten rid of it for basic comparsions but it could still use a clean up, honestly might better to make a ChipName enum.


nhtyy:
Still work to do around this if were supportive of continuing to use strings. 

If we changed to an enum I think we could even have `const name(&self) -> ChipName`

PR #1851, opened 2024-12-09
feat: write all logs to stderr (vs stdout)

imikushin:
## Motivation

Currently, all logs are being written to stdout. This breaks convention for CLI apps that might want to use stdout for, well, application output (not diagnostic stuff).

## Solution

Write logs to stderr.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/machine/src/utils/logger.rs
+++ b/crates/core/machine/src/utils/logger.rs
@@ -37,6 +37,7 @@ pub fn setup_logger() {
                     .with_thread_names(false)
                     .with_env_filter(env_filter)
                     .with_span_events(FmtSpan::CLOSE)
+                    .with_writer(std::io::stderr)
                     .finish()
                     .init();
             }


imikushin:
This simply changes one hardcoded default to another (saner) one, i.e. the simplest thing that could possibly work.

nhtyy:
Im not sure I support this, I also think this subscriber is more for internal use.. 

I think best practices here are to explicitly set your `RUST_LOG=<your_crate>=<level>`.

imikushin:
@nhtyy `<your_crate>=<level>` is orthogonal to what I'm proposing here. The point is whatever logging at whatever level is happening should only go to stderr, so that the familiar UNIX process contract is respected. 

If I'm building a CLI tool, and I want it to be used in shell scripts, it's users would expect my CLI tool to respect certain behavior for it to be easy to use in shell scripts. 

For example, if my tool can produce proofs and it can verify proofs, it'd be expected it'd work in scripts with commands like:
```sh
my_zkvm_tool prove -some -things | my_zkvm_tool verify -some -things
```

`my_zkvm_tool prove -some -things` should write the proof to stdout, and my_zkvm_tool verify -some -things` should read the proof from stdin. Should be as simple as that. My problem is that currently that's simply not possible! Precisely because stdout gets polluted with logs.

If, on the other hand, logs were being written to stderr, I'd be able to do as much logging as I want and still be able to run the above script successfully.

nhtyy:
>If I'm building a CLI tool, and I want it to be used in shell scripts, it's users would expect my CLI tool to respect certain behavior for it to be easy to use in shell scripts.

Absolutely, but I would disagree that we need to change `our` subscriber. As mentioned in the `tracing` crate docs here:  https://docs.rs/tracing/latest/tracing/#in-executables

> Warning: In general, libraries should not call set_global_default()! Doing so will cause conflicts when executables that depend on the library try to set the default later.

It is the executables responsibility to init a subscriber AND filter the relevant logs.

If you're importing our subscriber (the one you proposed changes on) you should not do this. If you are not importing our subscriber, your changes will have no effect!

By default, for instance `tracing_subscriber::fmt::init()` uses the EnvFilter, which reads from `RUST_LOG`.  So if youre using this, with something like `RUST_LOG=info` then you will see every crates tracing logs.

My recommendation would be to do something like
```rust
use tracing_subscriber::EnvFilter;

tracing_subscriber::fmt()
   .with_env_filter(
      EnvFilter::from_default_env()
         .add_directive("<your_crate>=<your_level>")
   ).init()
``` 

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

nhtyy:
This approaches allows users to expand logs if needed via the env var.

nhtyy:
I think it would be better to maybe even hide this from the docs, it is really not meant to be used by executables.

imikushin:
@nhtyy What would be a clean way to configure logging such that I get all the logs (that SP1 prints when RUST_LOG=info is set) printed to stderr? 

Should I still use `sp1_sdk::utils::setup_logger();`?

nhtyy:
```rust
let sp1_filter = EnvFilter::try_from_default_env()
            .unwrap_or_else(|_| EnvFilter::new("off"))
            .add_directive("hyper=off".parse().unwrap())
            .add_directive("p3_keccak_air=off".parse().unwrap())
            .add_directive("p3_fri=off".parse().unwrap())
            .add_directive("p3_dft=off".parse().unwrap())
            .add_directive("p3_challenger=off".parse().unwrap())
            .add_directive("sp1_cuda=off".parse().unwrap());
```

If you want to collect all the logs you would see with our subscriber, you could do something like this:
```rust
use tracing_subscriber::{fmt, Registry, Layer};

let subscriber = Registry::default()
    // writes with the same filtering logic as sp1 logger
    .with(fmt::layer().with_filter(env_filter).with_writer(std::io::stderr))
    // writes your crate only to std out
    .with(fmt::layer().with_filter(EnvFilter::new("<your_crate>=<your_level>"));

tracing::subscriber::set_global_default(subscriber).unwrap();
```

Note: Haven't tried to compile this, and of course theres multiple ways to do it.

nhtyy:
I do think we should probably have our logger hidden from docs, because its not supposed to be use in consumer executables.

PR #1855, opened 2024-12-10
feat: SDK refactor

mattstam:
API:

```rs
    // Local

    // Generate the proof.
    let client = ProverClient::builder()
        .local()
        .build();

    // Generate the proving key and verifying key for the given program.
    let pk = client.setup(&elf).await;

    // Generate the proof.
    let proof_result = client
        .prove(&pk, stdin)
        .compressed()
        .await;
        
    // Network

    // Generate the proof, using the specified network configuration.
    let client = ProverClient::builder()
        .network()
        .rpc_url(rpc_url)
        .private_key(private_key)
        .build();

    // Generate the proving key and verifying key for the given program.
    let pk = client.setup(Arc::from(&ELF[..])).await;

    // Generate the proof.
    let proof_result = client
        .prove(&pk, stdin)
        .compressed()
        .timeout(100)
        .cycle_limit(20_000)
        .await;

    // Old env var behavior
    
    let client = ProverClient::builder()
        .from_env();
```


crates/sdk/src/builder.rs
```diff
@@ -0,0 +1,55 @@
+pub struct None;
+pub struct Local;
```
mattstam, line 2: Should I also add a `Mock` struct at this level as well?

crates/sdk/src/local.rs
```diff
@@ -0,0 +1,83 @@
...
+        self.mode = mode;
+        self
+    }
+
+    pub async fn run(self) -> Result<SP1ProofWithPublicValues> {
+        self.prover.prove_with_options(self)
+    }
+}
+
+impl<'a> IntoFuture for LocalProofRequest<'a> {
```
mattstam, line 55: this is so you can just `.await` and it automatically runs

crates/sdk/src/network.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// `RequestTimedOut` error if the request does not received a response within the timeout.
+    pub async fn wait_proof<P: DeserializeOwned>(
+        &self,
+        request_id: &RequestId,
+        timeout_secs: u64,
+    ) -> Result<P, Error> {
+        todo!()
+    }
+
+    pub fn prove_with_options(
```
mattstam, line 76: I think `prove_with_options` could just be reduced to `prove` or `prove_with_request` is better

crates/sdk/src/client.rs
```diff
@@ -0,0 +1,62 @@
...
+        self
+    }
+
+    pub fn build(self) -> ProverClient {
+        ProverClient { inner: Box::new(self.inner_builder.build()) }
+    }
+}
+
+impl ProverClientBuilder<LocalProver> {
+    pub fn build(self) -> ProverClient {
```
nhtyy: I think there can be more options here

crates/sdk/src/request.rs
```diff
@@ -0,0 +1,8 @@
+use anyhow::Result;
+use sp1_prover::SP1ProofWithPublicValues;
+
+/// Common interface for proof requests
+pub trait ProofRequest {
+    /// Execute the proof request
```
nhtyy, line 6: This should not be async

crates/sdk/src/prover.rs
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    async fn prove(&self, pk: &SP1ProvingKey, stdin: SP1Stdin) -> Self::Request;
+
+    async fn verify(
+        &self,
+        proof: &SP1ProofWithPublicValues,
+        vk: &SP1VerifyingKey,
+    ) -> Result<(), SP1VerificationError> {
+        let local_prover = CpuProver::new();
```
nhtyy, line 29: Lets add feature gated *_blocking methods here

crates/sdk/src/prover.rs
```diff
@@ -0,0 +1,32 @@
...
+    type Request: ProofRequest;
+
+    async fn setup(&self, elf: &[u8]) -> Result<(SP1ProvingKey, SP1VerifyingKey)> {
+        let local_prover = CpuProver::new();
+        local_prover.setup(elf)
+    }
+
+    async fn execute(&self, elf: &[u8], stdin: SP1Stdin) -> Result<SP1Report> {
+        let local_prover = CpuProver::new();
+        local_prover.sp1_prover().execute(elf, stdin, Default::default())
```
nhtyy, line 19: this is blocking in an async fn

crates/sdk/src/network.rs
nhtyy: I agree I think that returning a `ProofRequest` with option methods replaces this

crates/sdk/src/local.rs
```diff
@@ -0,0 +1,83 @@
...
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'a>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(self.run())
+    }
+}
+
+#[async_trait]
+impl Prover for LocalProver {
+    type Request = LocalProofRequest<'_>;
```
nhtyy, line 66: Wait this doesnt compile right? GATs are not object (dyn) safe

crates/sdk/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub fn builder() -> ProverClientBuilder<None> {
+        ProverClientBuilder { inner_builder: None }
+    }
+
+    #[deprecated(note = "Use ProverClient::builder() instead")]
+    pub fn new() -> Self {
+        Self::create_from_env()
+    }
+
+    fn create_from_env() -> Self {
```
nhtyy, line 36: So I think that the DynBuilder should also expose default_{cycle_limt, timeout}?

I think that we should remove this method "create_from_env" and make the method like "build_with_env" that than overrides stuff at the end

crates/sdk/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub async fn prove(
+        &self,
+        pk: &SP1ProvingKey,
+        stdin: &SP1Stdin,
+    ) -> Result<SP1ProofWithPublicValues> {
+        let opts = ProofOpts::default();
+        self.inner.prove_with_options(pk, stdin, &opts).await
+    }
+
+    pub async fn prove_with_options(
```
nhtyy, line 74: rm this

crates/sdk/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    pub async fn execute(
+        &self,
+        elf: &[u8],
+        stdin: &SP1Stdin,
+    ) -> Result<(SP1PublicValues, ExecutionReport), ExecutionError> {
+        self.inner.execute(elf, stdin).await
+    }
+
+    pub async fn prove(
```
nhtyy, line 65: This should return a `DynProofRequest` which lets you set options like `prove(..).groth16()` as discussed yesterday

crates/sdk/src/local.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.mode = mode;
+        self
+    }
+
+    pub fn with_timeout(mut self, timeout: u64) -> Self {
+        self.timeout = timeout;
+        self
+    }
+
+    pub async fn run(self) -> Result<SP1ProofWithPublicValues> {
```
nhtyy, line 69: `run` methods should not be async

crates/sdk/src/local.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub timeout: u64,
+}
+
+impl<'a> LocalProofRequest<'a> {
+    pub fn new(prover: &'a LocalProver, pk: &'a SP1ProvingKey, stdin: SP1Stdin) -> Self {
+        Self { prover, pk, stdin, timeout: DEFAULT_TIMEOUT, mode: Mode::default() }
+    }
+
+    pub fn with_mode(mut self, mode: Mode) -> Self {
+        self.mode = mode;
```
nhtyy, line 60: I would rather expose `.plonk` and `.groth`

crates/sdk/src/local.rs
```diff
@@ -0,0 +1,144 @@
...
+    type Output = Result<SP1ProofWithPublicValues>;
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'a>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(self.run())
+    }
+}
+
+#[cfg(feature = "blocking")]
+impl<'a> ProofRequest for LocalProofRequest<'a> {
```
nhtyy, line 140: Lets just remove this trait

crates/sdk/src/network.rs
```diff
@@ -0,0 +1,292 @@
...
+        NetworkProverBuilder::new()
+    }
+}
+
+impl NetworkProverBuilder {
+    pub fn new() -> Self {
+        Self { rpc_url: None, private_key: None }
+    }
+
+    pub fn with_rpc_url(mut self, url: String) -> Self {
```
nhtyy, line 127: nit: `rpc_url`

crates/sdk/src/network.rs
```diff
@@ -0,0 +1,292 @@
...
+        self.rpc_url = Some(url);
+        self
+    }
+
+    pub fn with_private_key(mut self, key: String) -> Self {
```
nhtyy, line 132: nit: `private_key`

crates/sdk/src/network.rs
```diff
@@ -0,0 +1,292 @@
...
+        self.fulfillment_strategy = Some(strategy);
+        self
+    }
+
+    pub fn with_cycle_limit(mut self, cycle_limit: u64) -> Self {
+        self.cycle_limit = Some(cycle_limit);
+        self
+    }
+
+    pub async fn run(self) -> Result<SP1ProofWithPublicValues> {
```
nhtyy, line 193: `run` should not be async

crates/sdk/src/network.rs
```diff
@@ -0,0 +1,292 @@
...
+        &self,
+        proof: &SP1ProofWithPublicValues,
+        vk: &SP1VerifyingKey,
+    ) -> Result<(), SP1VerificationError> {
+        verify::verify(&self.prover, SP1_CIRCUIT_VERSION, proof, vk)
+    }
+}
+
+#[cfg(feature = "blocking")]
+impl ProofRequest for NetworkProofRequest<'_> {
```
nhtyy, line 288: rm this trait

crates/sdk/src/prover.rs
```diff
@@ -0,0 +1,40 @@
...
+use sp1_prover::{SP1ProvingKey, SP1VerifyingKey};
+
+use crate::{opts::ProofOpts, proof::SP1ProofWithPublicValues, provers::SP1VerificationError};
+
+#[async_trait]
+pub trait Prover: Send + Sync {
+    async fn setup(&self, elf: &[u8]) -> (SP1ProvingKey, SP1VerifyingKey);
+
+    async fn execute(
+        &self,
```
nhtyy, line 15: needs `setup_sync`, `execute_sync`

crates/sdk/src/prover.rs
```diff
@@ -0,0 +1,40 @@
...
+use async_trait::async_trait;
+use sp1_core_executor::{ExecutionError, ExecutionReport};
+use sp1_core_machine::io::SP1Stdin;
+use sp1_primitives::io::SP1PublicValues;
+use sp1_prover::{SP1ProvingKey, SP1VerifyingKey};
+
+use crate::{opts::ProofOpts, proof::SP1ProofWithPublicValues, provers::SP1VerificationError};
+
+#[async_trait]
+pub trait Prover: Send + Sync {
```
nhtyy, line 11: Id imagine this just needs to be `Sync`?

crates/sdk/src/request.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::future::{Future, IntoFuture};
+use std::pin::Pin;
+
+/// The default timeout seconds for a proof request to be generated (4 hours).
+pub const DEFAULT_TIMEOUT: u64 = 14400;
+
+/// The default cycle limit for a proof request.
+pub const DEFAULT_CYCLE_LIMIT: u64 = 100_000_000;
+
+pub trait ProofRequest {
```
nhtyy, line 17: rm this

crates/sdk/src/verify.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use crate::install::try_install_circuit_artifacts;
+use crate::opts::ProofOpts;
+use crate::provers::SP1VerificationError;
+use crate::{proof::SP1Proof, proof::SP1ProofKind, proof::SP1ProofWithPublicValues};
+
+/// Verify that an SP1 proof is valid given its vkey and metadata.
+/// For Plonk proofs, verifies that the public inputs of the PlonkBn254 proof match
+/// the hash of the VK and the committed public values of the SP1ProofWithPublicValues.
+pub fn verify(
```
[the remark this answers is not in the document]
nhtyy: Ya I think we need to expose verify somehow but lets revisit towards the end

crates/sdk/src/local.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn run(self) -> Result<SP1ProofWithPublicValues> {
+        executor::block_on(self.run())
```
nhtyy, line 142: Dont use `futures::executor` you should be using `tokio::spawn_blocking` only

nhtyy (commented):
okie looking good,

Some issues with async stuff though, lets remove `ProofRequest` traits, so users dont need to import it to use it, and make sure any method called `run` is not async.

So `LocalProofRequest` run just calls `prove` on the underlying `sp1_prover`, and its async version will just do a `tokio::task::spawn_blocking`

Now for Network its `run` either needs to use the blocking methods on the client if they exist or create a new runtime and block on that.

And finally for `DynProofRequest` its run method calls `prover.prove_with_options_sync` and its into future method will call `prove_with_options`

crates/sdk/src/client.rs
mattstam: do you mean as methods? Currently I have:
```
/// The default timeout seconds for a proof request to be generated (4 hours).
pub const DEFAULT_TIMEOUT: u64 = 14400;

/// The default cycle limit for a proof request.
pub const DEFAULT_CYCLE_LIMIT: u64 = 100_000_000;
```

crates/sdk/src/client.rs
mattstam: updated

crates/sdk/src/client.rs
mattstam: updated

crates/sdk/src/local.rs
mattstam: but they need to `await` in their logic

crates/sdk/src/local.rs
mattstam: updated

crates/sdk/src/local.rs
mattstam: updated

crates/sdk/src/network.rs
mattstam: Is that better? one idea is that we might have a `rpc_url` getting func that returns the value, so maybe it is good to keep `with_` for the setter

crates/sdk/src/prover.rs
mattstam: updated, also added verify_sync

crates/sdk/src/prover.rs
mattstam: updated

crates/sdk/src/request.rs
mattstam: updated

crates/sdk/src/local.rs
nhtyy: For the networkprover right? If the underlying client exposes blocking functions use that, otherwise youll need. to spawn a new exeucutor and call block on

LocalProver shouldnt need async to run, the point of run is to be the blocking (not async) entrypoint

crates/sdk/src/client.rs
nhtyy: Yaa I think since these are common defaults we can optionally let the user set them at builder level wdyt? Otherwise default to these

crates/sdk/src/network.rs
mattstam: updated

crates/sdk/src/network.rs
mattstam: updated

crates/sdk/src/local.rs
mattstam: updated

examples/fibonacci/script/bin/network.rs
```diff
@@ -0,0 +1,106 @@
+use sp1_sdk::network_v2::{Error, FulfillmentStrategy};
+use sp1_sdk::{include_elf, utils, ProverClient, proof::SP1ProofWithPublicValues, SP1Stdin};
+use std::time::Duration;
+use dotenv::dotenv;
+
+use std::sync::Arc;
+
+/// The ELF we want to execute inside the zkVM.
+const ELF: &[u8] = include_elf!("fibonacci-program");
```
mattstam, line 9: note: we will update this macro to return an Arc so its easy

crates/sdk/src/network/auth.rs
```diff
@@ -1,142 +0,0 @@
-use std::{borrow::Cow, str::FromStr};
```
mattstam, line 1: note: I am nuking v1 and I will convert `network-v2` to just `network` (after review because otherwise diff would be nightmarish)

PR #1891, opened 2024-12-23
chore(nix): initial nix support

dzmitry-lahoda:
Currently just setting up env for dev and building cli.

~Next step is making `build_toolchain` deterministic reproducible way (automated manual approach from source).~
i guess this is it for now . as i see tools are way integrated with home/s3/git releases/rustup, so nix will be hard to maintain until other people will as for secured and reproducible artifacts.

also there is magical random folder name genration in home. seems weird.


nhtyy:
Closing as stale.

PR #1896, opened 2024-12-24
fix: return error instead of panic when fail to upload artifact

keroro520:
## Motivation

The previous implementation used an assertion to handle upload failures, which isn't ideal for production code. This change provides better error handling and logging.

## Solution

- Replace assertion with proper error handling in `create_artifact_with_content`
- change debug log to error log for failed artifacts uploads


leruaa (approved):

leruaa:
Thanks, but I'm closing as the assert has been removed in the dev branch.

PR #1902, opened 2024-12-29
create bn254 scalar field chip

S1nus:
## Motivation

We want to compute [this function](https://github.com/scroll-tech/poseidon-circuit/blob/main/poseidon-base/src/hash.rs#L39) in SP1, poseidon hash over bn254 scalar, as it would be computed in a Halo2 circuit. This is for a DA integration with a zkrollup built with Halo2 KZG.

## Solution

It appears the bn254 precompiles in SP1 are for the base field, which makes sense to help verify proofs and compute pairings. However, iiuc this is not sufficient to perform computations as a halo2 circuit would, since the circuit inputs are elements of the **_scalar field_**.

Here I am developing a bn254 scalar field precompile, I expect it to be the same as Fp precompile but with a different modulus.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/machine/src/syscall/precompiles/fptower/fr.rs
+++ b/crates/core/machine/src/syscall/precompiles/fptower/fr.rs
@@ -1,0 +1,10 @@
+use std::marker::PhantomData;
+pub struct FrOpChip<P> {
+    _marker: PhantomData<P>,
+}
+
+impl<P> FrOpChip<P> {
+    pub const fn new() -> Self {
+        Self { _marker: PhantomData }
+    }
+}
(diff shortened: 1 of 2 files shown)


lispc:
bn254 fr is grumpkin fp.  this pr is useful for you  https://github.com/scroll-tech/sp1/pull/17/files

leruaa:
Unfortunately this PR has gone stale, sorry about @S1nus. Can you rebase your commits on latest dev branch?

PR #1940, opened 2025-01-16
docs(patches): add secp256r1 section

ratankaliani:
- Add `secp256r1` acceleration docs.


nhtyy (changes requested):
We should update this in the "docs" folder also, otherwise it wont be visible from "next"

nhtyy (approved):

PR #1941, opened 2025-01-16
feat: v4.0.1

nhtyy:
- Includes a warning message when using `cargo prove build --elf-name` without a corresponding `--output-directory`
- Includes a fix for programs failing to build on host that depend on `sp1-zkvm` & `sp1-lib`
- Bumps gnark docker test storage
- Adds docs that sp1up always installs and the newest version and you should use -C in ci
- Adds docs about potential toolchain issues with old versions of `sp1-build`
- Better panic messages on `read` / `read_vec`
- Fixes an eror when installing the cli from crates
- Updates the toolchain tag, the new version supports old libc versions on x86

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -5,9 +5,9 @@ name: book
 
 on:
   push:
-    branches: [main]
+    branches: [dev]
   pull_request:
-    branches: [main]
+    branches: [dev]
     paths:
       - "book/**"
   merge_group:
@@ -46,7 +46,7 @@ jobs:
 
   deploy:
     # Only deploy if a push to main
-    if: github.ref_name == 'main' && github.event_name == 'push'
+    if: github.ref_name == 'dev' && github.event_name == 'push'
     runs-on: ubuntu-latest
     needs: [build]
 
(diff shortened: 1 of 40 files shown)


nhtyy:
Missing version bump, do not merge until thats done.

crates/build/src/build.rs
```diff
@@ -72,6 +72,8 @@ pub fn execute_build_program(
 
             std::fs::copy(&elf_path, &output_path)?;
         }
+    } else if args.elf_name.is_some() {
```
ratankaliani, line 75: cargo:warning=ELF name is set but --output-directory is not used. ELF will be written to `target/xyz`

crates/zkvm/entrypoint/src/allocators/embedded.rs
ratankaliani, line 1: Can we add a new example/change one of the examples we run in CI to use the `embedded` allocator for an E2E test?

ratankaliani (dismissed):
LGTM, after the inclusion of `4.0.1` tag bump.

.github/workflows/book.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
   deploy:
     # Only deploy if a push to main
```
ratankaliani, line 48: nit: update comment

nhtyy, line 48:
```suggestion
    # Only deploy if a push to dev
```

PR #1950, opened 2025-01-17
feat: better failure messages when reading too much input

nhtyy:
Previously, reading to much input would panic the executor and lead to ambiguous error messages, 

this change means that u32::MAX len vecs will fail. It displays the callsite of `read` or `read_vec`

### DIFF
--- a/crates/core/executor/src/syscalls/hint.rs
+++ b/crates/core/executor/src/syscalls/hint.rs
@@ -10,7 +10,7 @@ impl Syscall for HintLenSyscall {
         _arg1: u32,
         _arg2: u32,
     ) -> Option<u32> {
-        // Note: This cast could be truncating on 64bit systems.
+        // Note: If the user supplies an input > than length 2^32, then the length returned will be truncated to 32-bits. Reading from the syscall will definitely fail in that case, as the BabyBear field is < 2^32.
         Some(ctx.rt.state.input_stream.front().map_or(u32::MAX, |data| data.len() as u32))
     }
 }
--- a/crates/core/executor/src/syscalls/hint.rs
+++ b/crates/core/executor/src/syscalls/hint.rs
@@ -10,10 +10,8 @@ impl Syscall for HintLenSyscall {
         _arg1: u32,
         _arg2: u32,
     ) -> Option<u32> {
-        panic_if_input_exhausted(ctx);
-
-        // Note: This cast could be truncating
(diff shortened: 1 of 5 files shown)


crates/core/executor/src/syscalls/hint.rs
```diff
@@ -10,10 +10,8 @@ impl Syscall for HintLenSyscall {
         _arg1: u32,
         _arg2: u32,
     ) -> Option<u32> {
-        panic_if_input_exhausted(ctx);
-
-        // Note: This cast could be truncating
-        ctx.rt.state.input_stream.front().map(|data| data.len() as u32)
+        // Note: This cast could be truncating on 64bit systems.
```
ratankaliani, line 13: I'm confused by this comment. Is the call below safe?

crates/zkvm/entrypoint/src/lib.rs
```diff
@@ -61,6 +62,9 @@ pub struct ReadVecResult {
 /// When the `embedded` feature is enabled, the buffer is read into the reserved input region.
 ///
 /// When there is no allocator selected, the program will fail to compile.
+///
+/// This function will return with junk data and set the `failed` flag to true if the input stream
```
ratankaliani, line 66: If the input stream is exhausted, the `failed` flag will be returned as `true`. In this case, the other outputs from the function are likely incorrect.

nhtyy: pls make suggestions so I can one click them in future

nhtyy, line 66:
```suggestion
/// If the input stream is exhausted, the failed flag will be returned as true. In this case, the other outputs from the function are likely incorrect.
```

crates/zkvm/entrypoint/src/lib.rs
```diff
@@ -61,6 +62,9 @@ pub struct ReadVecResult {
 /// When the `embedded` feature is enabled, the buffer is read into the reserved input region.
 ///
 /// When there is no allocator selected, the program will fail to compile.
+///
+/// If the input stream is exhausted, the failed flag will be returned as true. In this case, the other outputs from the function are likely incorrect.
+/// is exhausted.
```
nhtyy, line 67:
```suggestion
```

ratankaliani (approved):
LGTM

PR #1957, opened 2025-01-21
chore: remove redundant .run() call in prove method

mdqst:
## Motivation
 
I noticed that the `prove` method already handles the proof execution internally, so calling `.run()` after it is unnecessary. This PR removes the redundant call to clean up the code and avoid potential confusion.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


leruaa:
No, `prove()` creates a builder, and `run()` runs the prover.

PR #1972, opened 2025-01-23
perf: rm vec allocs from words conversions

mattsse:
## Motivation
replace intermediary to_vec allocs with [`array::from_fn`](https://doc.rust-lang.org/stable/std/array/fn.from_fn.html) leveraging the available const generics

## Solution
noticed these were duplicated, so I replaced the duplicated ones with re-exports

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


nhtyy:
LGTM thanks!

nhtyy (approved):

PR #1973, opened 2025-01-23
fix: deferred checkpoints optimization

ctian1:
* Currently, the entire ExecutionState.proof_stream is being cloned for every checkpoint, which is very wasteful for programs with many deferred proofs. Since it's only used as a runtime check, we can skip the clone and verification with checkpoints.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
           pull_token: ${{ secrets.PULL_TOKEN }}
 
       # If it's a nightly release, tag with the release time. If the tag is `main`, we want to use
-      # `latest` as the tag name. Else, use the tag name as is.
+      # `latest` as the tag name. else, use the tag name as is.
       - name: Compute release name and tag
         id: release_info
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           pull_token: ${{ secrets.PULL_TOKEN }}
 
       # If it's a nightly release, tag with the release time. If the tag is `main`, we want to use
-      # `latest` as the tag name. Else, use the tag name as is.
+      # `latest` as the tag name. else, use the tag name as is.
```
ratankaliani, line 36: nit: keep as is

.github/workflows/release.yml
ratankaliani, line 1: Are these changes relevant?

.github/workflows/toolchain-ec2.yml
ratankaliani, line 1: Are these changes relevant?

ratankaliani (approved):
Remove the release workflows, otherwise LGTM.

PR #1975, opened 2025-01-23
feat(docker-build): pass additional workspace directory

MdTeach:
## Motivation
In the current docker build it only mounts the guest-program workspace root and any relative import like [this](https://github.com/alpenlabs/strata/blob/main/provers/sp1/guest-btc-blockspace/Cargo.toml) will not work.

## Solution
Extend the BuildArgs, to make the workspace directory that docker mounts  configurable instead of guest-program workspace root 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/build/src/lib.rs
```diff
@@ -74,6 +74,15 @@ pub struct BuildArgs {
     pub elf_name: Option<String>,
     #[clap(alias = "out-dir", long, action, help = "Copy the compiled ELF to this directory")]
     pub output_directory: Option<String>,
+
+    #[clap(
+        alias = "workspace-dir",
+        long,
+        action,
+        help = "Optional workspace directory to be used",
```
nhtyy, line 82:
```suggestion
        help = "The top level directory to be used in the docker invocation.",
```

crates/build/src/lib.rs
```diff
@@ -74,6 +74,15 @@ pub struct BuildArgs {
...
+        default_value = None
```
nhtyy, line 83:
```suggestion
```

defaults to none

crates/build/src/command/docker.rs
```diff
@@ -28,7 +28,19 @@ pub(crate) fn create_docker_command(
...
+        .as_deref()
+        .map(|workspace_path| {
+            std::path::Path::new(workspace_path)
+                .to_path_buf()
+                .canonicalize()
+                .expect("Failed to canonicalize workspace directory")
+                .try_into()
+                .unwrap()
+        })
+        .unwrap_or_else(|| program_metadata.workspace_root.clone());
```
nhtyy, line 43: Can we add a sanity check here for ensuring the workspace directory is indeed a parent of the program?

nhtyy (changes requested):

nhtyy:
Thank you for this suggestion and sorry for the delay.

I left a few nits, also you can optionally allow maintainers to push to your branch, and ill get this over the line.

crates/build/src/command/docker.rs
MdTeach: We could but there could be the case like [ours](https://github.com/alpenlabs/strata/blob/main/provers/sp1/guest-btc-blockspace/Cargo.toml) where the guest program is separate cargo workspace project and parent directory which we want to be top level directory to be used in the docker mount would be cargo workspace. There could be need to make guest program separate workspace root since cargo patches can be applied to workspace root only.

crates/build/src/lib.rs
nhtyy: Option already implements default as `None` value

crates/build/src/command/docker.rs
nhtyy: Im not sure I follow, how could you compile a program without mounting its parent?

I don't mean direct child, just use `Path::starts_with`

MdTeach: Thanks, updated on  [b24a62a](https://github.com/alpenlabs/sp1/commit/b24a62a241f07cd199f3bfccee4831b8dc25ff01)

PR #1978, opened 2025-01-24
fix: Write diagnostics to stderr (vs stdout)

imikushin:
## Motivation

Before this change (even when logging was not enabled) SP1 was writing diagnostic messages to stdout, breaking the implicit contract that only the application logic should decide what gets written to stdout. This contract is important, e.g. for passing the output for further processing in shell-driven use cases:

```sh
cat ./input-data.json | my-app | another-app
```

This change aims to fix this issue, and to safeguard against it (to a reasonable extent) in the future.

## Solution

- Change `println!` usages to `eprintln!` in production modules.
- Add clippy warnings to runtime modules guarding against using printing to stdout. `eprintln!` is okay 👍.
- Add exceptions to the above clippy warnings to tests and build scripts.
- ~~Direct logging output to stderr by default.~~

I've left alone tests, examples and build scripts, and tried to make the change as narrow as possible so it affects only production code paths.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,3 +157,6 @@ default.extend-ignore-re = [
   "CommitCommitedValuesDigest",
 ]
 default.extend-ignore-words-re = ["(?i)groth", "TRE"]
+
+[workspace.lints.clippy]
+print_stdout = "deny"
--- a/crates/core/executor/Cargo.toml
+++ b/crates/core/executor/Cargo.toml
@@ -65,3 +65,6 @@ profiling = [
   "dep:gecko_profile",
   "dep:indicatif",
 ]
+
+[lints]
+workspace = true
--- a/crates/core/executor/src/executor.rs
+++ b/crates/core/executor/src/executor.rs
@@ -271,12 +271,12 @@ impl<'a> Executor<'a> {
             });
 
             if let Some(trace_buf) = trace_buf {
-                println!("Profiling enabled");
+                eprintln!("Profiling enabled");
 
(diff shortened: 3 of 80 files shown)


ratankaliani (commented):
Thanks for this @imikushin!

Can you move the `warn!` and `allow!` attributes to the TOML files, rather than in `.rs`?

This will avoid replication like what we see in `crates/core/machine/src/**`.

By the way, have you confirmed this branch works as you expect?

imikushin:
@ratankaliani I believe the CI only complained about formatting: fixed that.

imikushin:
From a chat with @nhtyy elsewhere: 

> Sorry @imikushin last thing would you mind setting up the clippy lint to be applied globally?
> 
> You should be able to achieve this by adding the following lines to 
> 
> Top level Cargo.toml
> [workspace.lints.clippy]
> print_stdout = "deny"
> and in each package you also need to add
> [lints]
> workspace = true
> and then you can get rid of the various lint declrations in the source code

@nhtyy Please confirm this is the only thing left that you guys want from this PR. IMO this grows the scope of the PR and is inconsistent with all other existing lints that SP1 already has in `lib.rs` files. 

A separate PR could migrate all of them to Cargo.toml if there's a will.

nhtyy:
@imikushin Yes we would appreciate it if you could add it to the `Cargo.toml`. Thank you

imikushin:
@nhtyy Done!

PR #1982, opened 2025-01-24
fix(cli): elf name requires output directory

### LINKED ISSUE #1979: `--elf-name` is ignored when `--output-directory` is missing
### Component

cargo prove CLI/sp1up

- [x] SP1 SDK
- [x] cargo prove CLI/sp1up

_No response_

4a1dcea 2025-01-15T23:21:07.590940000Z

### Operating System

None

### Describe the bug

cargo prove build accepts `--elf-name` and `--output-directory` arguments. If `--elf-name` is present while the `--output-directory` is not, then the `--elf-name` is fully ignored and the binary is generated in the `target/` directory with the default name. The user also receives [this warning](https://github.com/succinctlabs/sp1/blob/5972269ccae78024b27bfc4dd19a5ba5a56534b3/crates/build/src/build.rs#L76).

The warning doesn't make it clear that `--elf-name` is ignored. Moreover, in a different but similar scenario, [cargo-prove simply returns an error](https://github.com/succinctlabs/sp1/blob/5972269ccae78024b27bfc4dd19a5ba5a56534b3/crates/build/src/build.rs#L53) when `--elf-name` would be ignored. I believe that in this scenario, cargo-prove should simply return an error, or use the `--elf-name`.

nhtyy:
closes #1979

The issue brings up a good point that error message currently provided when trying to use elf name without an output directory is not clear enough.

As we want to discourage copying of ELF files, and the use of `include_elf`/ build script flow, we require output directory is set if elf name is set.

We still do not allow elf name if multiple guest targets are found.

### DIFF
--- a/crates/build/src/build.rs
+++ b/crates/build/src/build.rs
@@ -72,8 +72,6 @@ pub fn execute_build_program(
 
             std::fs::copy(&elf_path, &output_path)?;
         }
-    } else if args.elf_name.is_some() {
-        println!("cargo:warning=ELF name is set but --output-directory is not used, the ELF will be place in the `target/` directory");
     }
 
     print_elf_paths_cargo_directives(&target_elf_paths);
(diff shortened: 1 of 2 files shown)


ratankaliani (approved):
LGTM, once CI fixes are made.

PR #2002, opened 2025-01-30
docs: Add memory argument to Proof System

flyingnobita:
## Description

- Add mention of the memory argument in the Proof System section

## Motivation

- Let readers be aware of memory argument that makes use of multiset hashing

## PR Checklist

- [x] Added Documentation


nhtyy:
@flyingnobita I think youre gonna need to move the PDF to `docs/static` ?

Ill run the workflow to see if it builds

leruaa:
The book has been moved to another repo

PR #2011, opened 2025-01-31
chore: bump alloy deps

ratankaliani:
## Overview

Bump `alloy-signer` dependencies to latest. Due to `alloy`'s `0.X` status, minor versions with different traits cause conflicts with one another.

https://github.com/op-rs/kona recently bumped to the latest `alloy` version, so in order to have `sp1-sdk` and `kona-host` in the same workspace, we need to bump `alloy-signer` and `alloy-signer-local` to `0.11`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,9 +111,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-consensus"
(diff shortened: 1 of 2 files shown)


nhtyy (approved):

SyedAhkam:
Could we please have a new release (v4.0.2) with this applied? I can't upgrade my project to alloy `0.11` and have it work alongside sp1 `v4.0.1`.

PR #2023, opened 2025-02-03
docs: add note about slow initialization times for ProverClient

umadayal:
## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/book/docs/generating-proofs/recommended-workflow.mdx
+++ b/book/docs/generating-proofs/recommended-workflow.mdx
@@ -14,7 +14,7 @@ Note that printing out the total number of executed cycles and the full executio
 
 **Crate Setup:** We recommend that your program crate that defines the `main` function (around which you wrap the `sp1_zkvm::entrypoint!` macro) should be kept minimal. Most of your business logic should be in a separate crate (in the same repo/workspace) that can be tested independently and that is not tied to the SP1 zkVM. This will allow you to unit test your program logic without having to worry about the `zkvm` compilation target. This will also allow you to efficient reuse types between your program crate and your crate that generates proofs.
 
-**Note:** The `ProverClient` should be initialized once and reused for subsequent proving operations. This is because the initial initialization may be slow as it loads necessary proving parameters and sets up the environment. For this reason, you should initialize the `ProverClient` once and reuse it rather than creating new instances.
+**Note:** The `ProverClient` should be initialized once and reused for subsequent proving operations rather than creating new instances. This is because the initial initialization may be slow as it loads necessary proving parameters and sets up the environment. You can wrap the `ProverClient` in an `Arc` to share it across tasks.
 
 ## Step 2: Generate proofs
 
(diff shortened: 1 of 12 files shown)


nhtyy:
I wonder if this should be in common issues actually? We should add something that says how to share it, I think just saying put it in an Arc is fine.

nuke-web3:
Hit this exactly building a service around the SDK here: https://github.com/celestiaorg/eq-service/issues/29

I would not expect the client to be so very heavy to startup, without clear caching mechanisms for the heavy parts. Had to make my own (that might not be widely applicable) to store details rather than recompute them each time.

I am wanting to preprocess as much as possible for requesting proofs against a common program, to have the lightest job at startup for the ProverClient. Would love to see this kind of workflow more supported :pray: and any tips/examples on how to achieve that. 

For example, should I be using https://docs.rs/sp1-sdk/latest/sp1_sdk/network/client/struct.NetworkClient.html#method.register_program at all to speedup startup and/or fulfillment?

umadayal:
> Hit this exactly building a service around the SDK here: [celestiaorg/eq-service#29](https://github.com/celestiaorg/eq-service/issues/29) I think this will be common as all examples show how to set this up as a one-off, nothings about reusing the service. Most RPC/REST clients I have used are very light weight, so sharing isn't a hard requirement, where this one surely is.
> 
> I would not expect the client to be so very heavy to startup, without clear caching mechanisms for the heavy parts. Had to make my own (that might not be widely applicable) to store details rather than recompute them each time.
> 
> I am wanting to preprocess as much as possible for requesting proofs against a common program, to have the lightest job at startup for the ProverClient. Would love to see this kind of workflow more supported 🙏 and any tips/examples on how to achieve that.
> 
> For example, should I be using https://docs.rs/sp1-sdk/latest/sp1_sdk/network/client/struct.NetworkClient.html#method.register_program at all to speedup startup and/or fulfillment?

Thanks for your comment! Looking at your code, I noticed that you initialize the prover and setup for every job, but you can actually initialize/setup once and reuse the same prover, pk, and vk across multiple jobs. This should significantly reduce startup overhead. You may wrap the prover in an Arc to share it across tasks. Let us know if you have any questions!

book/docs/generating-proofs/recommended-workflow.mdx
```diff
@@ -14,6 +14,8 @@ Note that printing out the total number of executed cycles and the full executio
 
 **Crate Setup:** We recommend that your program crate that defines the `main` function (around which you wrap the `sp1_zkvm::entrypoint!` macro) should be kept minimal. Most of your business logic should be in a separate crate (in the same repo/workspace) that can be tested independently and that is not tied to the SP1 zkVM. This will allow you to unit test your program logic without having to worry about the `zkvm` compilation target. This will also allow you to efficient reuse types between your program crate and your crate that generates proofs.
 
+**Note:** The `ProverClient` should be initialized once and reused for subsequent proving operations. This is because the initial initialization may be slow as it loads necessary proving parameters and sets up the environment. For this reason, you should initialize the `ProverClient` once and reuse it rather than creating new instances.
```
kevjue, line 17: I think  the first and third sentence can be merged.  Also, I think @nhtyy 's idea of suggesting wrapping in an `Arc` is a good one that should be added.

kevjue (approved):
LGTM

PR #2040, opened 2025-02-07
fix: const test iterations

nhtyy:
The test benchmarking requires a fixed number of iterations to be comparable. 

The benchmarker assumes old tests are equivalent, and uses old code to get the comparison cycle counts, therefore, either the old tests must be conducive to passing the number off to the benchmarker, or they can simply use a fixed number of iterations. 

For now, this PR does the latter.


rkm0959 (approved):
LGTM!

PR #2047, opened 2025-02-07
feat: v4.1.0-dev

nhtyy:
This PR bumps the CUDA prover docker image to the newest verison, it also includes a QOL improvement in the SDK if an empty private key is submitted.

The patch-testing cycle tracking workflow is updated to display reductions in cycles with a negative value.

### DIFF
--- a/crates/sdk/src/network/builder.rs
+++ b/crates/sdk/src/network/builder.rs
@@ -76,7 +76,7 @@ impl NetworkProverBuilder {
     pub fn build(self) -> NetworkProver {
         let private_key = match self.private_key {
             Some(private_key) => private_key,
-            None => std::env::var("NETWORK_PRIVATE_KEY").expect(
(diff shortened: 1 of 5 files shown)


crates/sdk/src/network/builder.rs
```diff
@@ -76,7 +76,7 @@ impl NetworkProverBuilder {
     pub fn build(self) -> NetworkProver {
         let private_key = match self.private_key {
             Some(private_key) => private_key,
-            None => std::env::var("NETWORK_PRIVATE_KEY").expect(
+            None => std::env::var("NETWORK_PRIVATE_KEY").ok().filter(|k| !k.is_empty()).expect(
```
ratankaliani, line 79: Isn't this functionally the same as before?

ratankaliani (approved):
LGTM

PR #2048, opened 2025-02-08
docs: version 4.1.0

nhtyy:
This PR versions the docs for 4.1.0. 

Previously we had only maintained "versioned docs", but some commits ended up in the old docs, so this commit reenables "next" docs and unifies the diff between the unmaintained docs and the versioned, thankfully small.

I did this by deleting the unmaintained "next docs", and replacing it with the 4.0.0 docs, then I went through the `git diff`, and manually unified them.


crates/zkvm/lib/src/ecdsa/affine.rs
ratankaliani, line 1: Is this necessary?

ratankaliani (approved):
LGTM! Gave a pass over the `next` docs, the `4.1.0` docs should be the exact same.

book/versioned_docs/version-4.1.0/writing-programs/patched-crates.md
```diff
@@ -0,0 +1,226 @@
...
+
+Apply the following patches based on what crates are in your dependencies.
+
+- `k256`
+
+  ```toml
+  k256 = { git = "https://github.com/sp1-patches/elliptic-curves", tag = "patch-k256-13.4-sp1-4.1.0" }
+  ```
+
+  Note: The curve operations for `k256` are inside of the `ecdsa-core` crate, so you don't need to patch `k256` itself, and just patching `ecdsa-core` is enough.
```
distractedm1nd, line 108: this is confusing, is it just an artefact from the current version?

nhtyy:
Closing, as docs are temporally hosted elsewhere.

PR #2064, opened 2025-02-13
fix: SDK client latency

gg2001:
## Motivation

Cuts ~1.5s of network proving latency from the SDK

## Solution

Before:
Screenshot 2025-02-12 at 5 23 45 PM

After:
Screenshot 2025-02-12 at 5 21 28 PM

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


nhtyy:
My problem with this is that it requires you to drop the network prover if for some reason this channel gets closed right? Like id imagine if theres some intermittent outage, than this network prover is effectively unusable forever?

Issue #1839, opened 2024-12-04
--bare option seems to create EVM template


S1nus:
### Component

cargo prove CLI/sp1up

### Have you ensured that all of these are up to date?

- [X] SP1 SDK
- [X] cargo prove CLI/sp1up

### What version of the cargo prove CLI are you on?

cargo-prove sp1 (3917e90 2024-12-04T00:40:40.640074686Z)

### Operating System

Linux (Ubuntu)

### Describe the bug

Creating a new project with `cargo prove new --bare myproject` generates template files that seem to have EVM stuff. the `script/src/bin` directory contains an `evm.rs`, and `Cargo.toml` has `alloy-sol-types` as a dependency

varun-doshi:
This happens on MacOS as well

leruaa:
Work has been done to remove contracts folder and git submodules.

PR #2070, opened 2025-02-13
fix: Remove evm specific code when using new with --bare template

alexanderattar:
## Motivation
When creating a new bare project using `cargo prove new --bare`, EVM-related files and dependencies were still being included, which are unnecessary for projects not requiring EVM integration. 

This PR resolves issue: https://github.com/succinctlabs/sp1/issues/1839

## Solution
- Added functionality to remove EVM-specific files and dependencies when creating a bare project:
  - Removes the `contracts` directory and `.gitmodules` file
  - Removes the EVM-specific script file (`evm.rs`)
  - Removes any `alloy-sol` references from all `Cargo.toml` files in the project
- Added tests to verify the new behavior:
  - Tests for bare template creation
  - Tests for EVM template creation
  - Tests to ensure `alloy-sol` references are properly removed

image

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -44,3 +44,6 @@ prettytable-rs = "0.10"
 textwrap = "0.16.0"
 ctrlc = "3.4.2"
 cargo_metadata = "0.18.1"
+
+[dev-dependencies]
+tempfile = "3.16.0"
--- a/crates/cli/src/commands/new.rs
+++ b/crates/cli/src/commands/new.rs
@@ -9,29 +9,29 @@ use yansi::Paint;
 
 #[derive(Args)]
 #[group(required = true, multiple = false)]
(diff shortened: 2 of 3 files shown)


alexanderattar:
There wasn't precedent for tests in the cli crate so I just went ahead and added a test directory (I can move tests elsewhere if preferred). 

I also didn't see any docs in the cli crate. Please lmk if I should add something for these changes

nhtyy:
Hey thank you taking the time to make this PR! 

On our end we decided that the best way to handle is to actually just maintain two branches in the template, one for bare, one for EVM. This way, we can have faster clone times.

alexanderattar:
no problem. alright that makes sense. thanks for the heads up @nhtyy!

Issue #2059, opened 2025-02-11
Trying to create Groth16 proofs fails with "failed to prove with docker: docker command failed" in Apple Silicon


MauroToscano:
### Component

sp1-zkvm

### Have you ensured that all of these are up to date?

- [x] SP1 SDK
- [x] cargo prove CLI/sp1up

### What version of SP1 SDK are you on?

Using the 4.0.1 tag

### What version of the cargo prove CLI are you on?

Using the 4.0.1 tag

### Operating System

macOS (Apple Silicon)

### Describe the bug

Trying to create Groth16 proofs fails with "failed to prove with docker: docker command failed" in Apple Silicon

ratankaliani:
Was Docker running on the machine when you ran this command?

leruaa:
We added a way to have more meaningful errors from Docker

PR #2085, opened 2025-02-24
fix(recursion): include stderr from Docker in errors

nhtyy:
Improve observability if Docker call fails.

Helps with #2059

### DIFF
--- a/crates/recursion/gnark-ffi/src/ffi/docker.rs
+++ b/crates/recursion/gnark-ffi/src/ffi/docker.rs
@@ -51,13 +51,12 @@ fn call_docker(args: &[&str], mounts: &[(&str, &str)]) -> Result<()> {
     cmd.args(args);
     let result = cmd.output()?;
     if !result.status.success() {
+        let stderr = String::from_utf8_lossy(&result.stderr);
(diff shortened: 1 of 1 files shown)


crates/recursion/gnark-ffi/src/ffi/docker.rs
```diff
@@ -49,11 +49,15 @@ fn call_docker(args: &[&str], mounts: &[(&str, &str)]) -> Result<()> {
     }
     cmd.arg(get_docker_image());
     cmd.args(args);
-    let result = cmd.status()?;
-    if !result.success() {
+    let result = cmd.output()?;
+    if !result.status.success() {
         log::error!("Failed to run `docker run`: {:?}", cmd);
         log::error!("Execution result: {:?}", result);
```
ratankaliani, line 55: Why shouldn't the error be logged here? Will it be logged in a parent fn?

nhtyy: Added to logging as well. We unwrap on this err in the parent fn which would have printed the log.

ratankaliani (approved):
LGTM

PR #2089, opened 2025-02-25
feat: give access to used memory in the zkvm

leruaa:
## Motivation

For benchmarks it would be useful to have a way to access the memory usage in the zkvm, when using the bump global allocator

## Solution

Adding a `used_memory()` function that returns the used memory.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/zkvm/entrypoint/src/syscalls/memory.rs
+++ b/crates/zkvm/entrypoint/src/syscalls/memory.rs
@@ -15,6 +15,10 @@
 // Memory addresses must be lower than BabyBear prime.
 pub const MAX_MEMORY: usize = 0x78000000;
 
+// Pointer to next heap address to use, or 0 if the heap has not yet been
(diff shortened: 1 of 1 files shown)


crates/zkvm/entrypoint/src/syscalls/memory.rs
```diff
@@ -15,6 +15,10 @@
 // Memory addresses must be lower than BabyBear prime.
 pub const MAX_MEMORY: usize = 0x78000000;
 
+// Pointer to next heap address to use, or 0 if the heap has not yet been
+// initialized.
+static mut HEAP_POS: usize = 0;
```
ratankaliani, line 20: Gate behind `bump` feature flag?

ratankaliani (approved):
LGTM

PR #2109, opened 2025-02-28
feat: Rust 1.85 toolchain support

nhtyy:


crates/cli/src/commands/build_toolchain.rs
```diff
@@ -74,19 +74,11 @@ impl BuildToolchainCmd {
...
 
-        // Build the toolchain (stage 1).
-        Command::new("python3")
-            .env("RUST_TARGET_PATH", &temp_dir)
-            .env("CARGO_TARGET_RISCV32IM_SUCCINCT_ZKVM_ELF_RUSTFLAGS", "-Cpasses=lower-atomic")
-            .args(["x.py", "build"])
-            .current_dir(&rust_dir)
-            .run()?;
-
         // Build the toolchain (stage 2).
```
ratankaliani, line 77: Nit: Is this now stage 1?

ratankaliani (approved):
LGTM

crates/cli/src/commands/build_toolchain.rs
nhtyy: This builds all the stages

nhtyy: No need for explicit separation

ratankaliani: Should we update the comment then?

ctian1 (approved):

Issue #1625, opened 2024-10-10
Nicer error message when toolchain version and `sp1-zkvm` versions are incompatible


MCJOHN974:
Hi all! I worked on some project which used SP1. It was depending on a fixed commit, rather than latest version. I cloned the project, downloaded and setted up sp1 following sp1 book, runned `cargo prove build` and faced this monstrous error:
```
[sp1]  error: linking with `rust-lld` failed: exit status: 1
[sp1]    |
[...]

PR #2115, opened 2025-02-28
feat: locked version checking vs version in cargo prove

leruaa:
## Motivation

Version checking the sp1-zkvm crate version with the current toolchain to ensure compatibility.

Assumes that major/minor of sp1-zkvm and sp1-build matching implies compat, which would be inline if sp1-zkvm inherits workspace version in the future

Closes https://github.com/succinctlabs/sp1/issues/1625


.github/workflows/pr.yml
```diff
@@ -354,6 +354,13 @@ jobs:
         run: |
           find -name Cargo.lock -type f -exec rm {} \;
 
+
+      # We need `cargo prove` to verify the locked version of `sp1-zkvm`
+      - name: Install SP1
+        run: |
+          cd crates/cli
+          cargo install --locked --force --path .
+
```
ratankaliani, line 363: What are you using this for?

To test the issue in the PR you basically need to run `cargo prove build` on an older version than the `sp1-zkvm` version in a crate.

Or, you run `cargo prove build`, and there have been changes to the syscalls/linking that is non-backwards compatible across versions (e.g. changing the definition of `sys-rand`).

The user seems to have gotten this backwards (unless there was a modification to the arguments for `sys_rand`).

crates/build/src/build.rs
```diff
@@ -206,3 +211,132 @@ fn print_elf_paths_cargo_directives(target_elf_paths: &[(String, Utf8PathBuf)])
...
+
+    if let Some(sdk_version) = maybe_sdk {
+        if sdk_version.major != toolchain_version.major
+            || sdk_version.minor != toolchain_version.minor
+        {
+            return Err(anyhow::anyhow!(
+                "Locked version of sp1-sdk is incompatible with the current toolchain version"
+            ));
+        }
+    }
```
ratankaliani, line 327: Why do we need to check this?

crates/build/src/build.rs
```diff
@@ -206,3 +211,132 @@ fn print_elf_paths_cargo_directives(target_elf_paths: &[(String, Utf8PathBuf)])
...
+        });
+        println!("cargo:warning=Locked version of sp1-zkvm is {}", vm_version);
+        println!("cargo:warning=Current toolchain version = {}", toolchain_version);
+    }
+
+    if vm_version.major != toolchain_version.major || vm_version.minor != toolchain_version.minor {
+        return Err(anyhow::anyhow!(
+            "Locked version of sp1-zkvm is incompatible with the current toolchain version"
+        ));
+    }
```
ratankaliani, line 317: We should perform the following check here:

- `cargo prove` should always be backwards-compatible within the same major version. Ex. `cargo prove` `1.2` is compat with `sp1-zkvm` `1.1`. However, `sp1-zkvm` `1.2` is not guaranteed to be compat with `cargo prove` `1.1`.

crates/build/src/build.rs
```diff
@@ -206,3 +211,132 @@ fn print_elf_paths_cargo_directives(target_elf_paths: &[(String, Utf8PathBuf)])
...
+
+    // You need an entrypoint, therefore you need sp1-zkvm.
+    let vm_version = &lock_file
+        .packages
+        .iter()
+        .find(|p| p.name.as_str() == "sp1-zkvm")
+        .expect("The guest program should have a sp1-zkvm dependency")
+        .version;
+
+    let maybe_sdk =
```
ratankaliani, line 281: You shouldn't need the `sp1-sdk` version

ratankaliani (commented):
I wonder if there's a more minimal way to do this. IMO there are 3 levels.

1. Hold the invariant that for the latest `sp1up`, `cargo prove` and `sp1-zkvm` that compilation always works. This is what we have today.
2. Add better informational logging on `cargo prove --version`. Users should be able to check the supported `sp1-zkvm` range for their `cargo prove` version. `cargo prove --version` should return a supported range of `sp1-zkvm`.
3. In `cargo prove build`, add a simple check that validates `cargo prove` version > `sp1-zkvm` version and on the same major version. **That's this PR.**

You shouldn't need to check the `sp1-sdk` version.

crates/build/src/build.rs
```diff
@@ -206,3 +211,132 @@ fn print_elf_paths_cargo_directives(target_elf_paths: &[(String, Utf8PathBuf)])
...
+                break;
+            } else {
+                curr_path = parent;
+            }
+        }
+
+        if !lock_path.is_file() {
+            panic!("Failed to find Cargo.lock in the program directory or its parent directories");
+        }
+    }
```
ratankaliani, line 267: `cargo-metadata` is a crate which should expose a method to get the `workspace_root`.

crates/build/src/build.rs
```diff
@@ -206,3 +211,132 @@ fn print_elf_paths_cargo_directives(target_elf_paths: &[(String, Utf8PathBuf)])
...
+            version
+                .split_once('\n')
+                .expect("The version should have whitespace, this is a bug")
+                .1
+                .trim(),
+        )
+        .expect("Failed to parse version string, this is a bug")
+    };
+
+    if verbose {
```
ratankaliani, line 305: Add comment here for where verbose flag is set.

crates/build/src/lib.rs
```diff
@@ -74,14 +74,15 @@ pub struct BuildArgs {
...
     pub output_directory: Option<String>,
-
     #[clap(
         alias = "workspace-dir",
         long,
         action,
         help = "The top level directory to be used in the docker invocation."
     )]
     pub workspace_directory: Option<String>,
+    #[clap(long, action, help = "Any debug logs from the build script will be printed to stdout")]
```
ratankaliani, line 84:
```suggestion
    #[clap(long, action, help = "Prints debug logs to stdout.")]
```

crates/cli/src/lib.rs
```diff
@@ -12,8 +12,16 @@ pub const RUSTUP_TOOLCHAIN_NAME: &str = "succinct";
...
-    concat!("sp1", " (", env!("VERGEN_GIT_SHA"), " ", env!("VERGEN_BUILD_TIMESTAMP"), ")");
+pub const SP1_VERSION_MESSAGE: &str = concat!(
+    "sp1",
+    " (",
+    env!("VERGEN_GIT_SHA"),
+    " ",
+    env!("VERGEN_BUILD_TIMESTAMP"),
+    ")",
+    "\n",
+    env!("CARGO_PKG_VERSION")
```
ratankaliani, line 23: Why do you need to add this? And why is there a newline char?

crates/build/fixtures/fibonacci-old-zkvm/src/main.rs
ratankaliani, line 1: You should put this on version `3.0.0` so that you confirm the `cargo prove` mismatch check is caught. Version `4.0.0` shouldn't be incompatible with `cargo prove` on version `4.1.0`.

ratankaliani, line 1: Note: It's confusing this is fibonacci, just make this an empty program.

ratankaliani (commented):
The logic for validating the compatibility between `cargo prove` and `sp1-zkvm` should be corrected. Additionally, left some cleanup comments that are useful.

nhtyy:
Basically I don't think we should actually do this.

To actually check if a program will work on some given version it would require checking libs and patches. 

I think it makes more sense to inject a custom cfg flag that we check *in all future versions* of patches and guest crates.

We can instead do something like:

```
```

nhtyy:
TBH I think theres just a lot of nuance depending on what exactly we want to accomplish: 

We should define exactly what "incompatibles" we want to detect.

PR #2120, opened 2025-03-03
feat: track invocation count

leruaa:
## Motivation

Sometimes, when using report annotations, we need to know the invocation count. My use case was to be able to track the average precompiles and opcodes cycles count.

## Solution

Adding a `invocation_tracker` map that increment by 1 for the given label each times the `ReportEnd` command is called.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/executor/src/report.rs
+++ b/crates/core/executor/src/report.rs
@@ -17,7 +17,7 @@ pub struct ExecutionReport {
     pub syscall_counts: Box<EnumMap<SyscallCode, u64>>,
     /// The cycle tracker counts.
     pub cycle_tracker: HashMap<String, u64>,
-    /// The invocation tracker counts.
+    /// Tracker for the number of `cycle-tracker-report-*` invocations for a specific label.
     pub invocation_tracker: HashMap<String, u64>,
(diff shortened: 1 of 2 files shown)


crates/core/executor/src/report.rs
```diff
@@ -17,6 +17,8 @@ pub struct ExecutionReport {
     pub syscall_counts: Box<EnumMap<SyscallCode, u64>>,
     /// The cycle tracker counts.
     pub cycle_tracker: HashMap<String, u64>,
+    /// The invocation tracker counts.
```
ratankaliani, line 20:
```suggestion
    /// Tracker for the number of `cycle-tracker-report-start` invocations for a specific label.
```

crates/core/executor/src/syscalls/write.rs
ratankaliani, line 1: Add a comment above the `end_cycle_tracker` invocation about the tracking of the number of invocations.

ratankaliani (approved):
LGTM after CI passes and the recommended suggestions + comments are added

PR #2123, opened 2025-03-03
feat(ci): Create release and tag on pushes to main 

nhtyy:
fix(ci): Create a versioned release by parsing the version from `sp1-build`. This version is also used to set the default docker tag. Note: Any subsequent pushes to main, without changing the cargo version, will rebuild and replace the binaries on the tag.

Introduces a breaking change to `sp1up` to read from the actual "latest" release rather than the `latest` tag.


crates/build/Cargo.toml
```diff
@@ -2,6 +2,13 @@
 name = "sp1-build"
 description = "Build an SP1 program."
 readme = "README.md"
+
+# NOTE: The release workflow relies on this version being set by the toplevel workspace.
+# DO NOT CHANGE THIS, unless you change the release workflow accordingly.
+#
+# This version is used as
+# - The release tag name
+# - The default Docker image tag for this version.
```
ratankaliani, line 11:
```suggestion

# Note: The release workflow depends on `sp1-build` version being set by the workspace, and that the `sp1-build` version refers to the canonical version.
# **Do not change this**, unless the release workflow is changed accordingly.
#
# The version of `sp1-build` is used for:
# - The name of the release tag on merge to `main`.
# - The default Docker image tag for this SP1 version..
```

.github/workflows/release.yml
```diff
@@ -66,7 +63,6 @@ jobs:
         id: build_changelog
         uses: mikepenz/release-changelog-builder-action@v4
         with:
-          configuration: "./.github/changelog.json"
```
ratankaliani, line 69: Why'd you remove this?

.github/workflows/release.yml
```diff
@@ -41,19 +39,18 @@ jobs:
             echo "tag_name=nightly-${GITHUB_SHA}" >> $GITHUB_OUTPUT
             echo "release_name=Nightly ($(date '+%Y-%m-%d'))" >> $GITHUB_OUTPUT
           elif [[ "${GITHUB_REF_NAME}" == "main" ]]; then
-            echo "tag_name=latest" >> $GITHUB_OUTPUT
-            echo "release_name=Latest" >> $GITHUB_OUTPUT
+            # Parse the version from the `Cargo.toml` file.
+            VERSION=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "sp1-build") | .version')
+
+            echo "tag_name=v${VERSION}" >> $GITHUB_OUTPUT 
+            echo "release_name=v${VERSION}" >> $GITHUB_OUTPUT
```
ratankaliani, line 46: Is this the best way to do it? Can we not read from `CARGO_PKG_VERSION` here?

nhtyy: I dont think so

nhtyy: Open to suggestions though

PR #2129, opened 2025-03-05
fix: don't retry if a transport error is caused by "no native certs f…

leruaa:
…ound"

## Motivation

When proving on the prover network on an environment where the `ca-sertificates` package is missing, the error returned contains "transport error" and [is considered as a transient error](https://github.com/succinctlabs/sp1/blob/b6eb0482541fe8816230229574a4fb199c5c01a1/crates/sdk/src/network/retry.rs#L89), so we uselessly retry:

```
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...    
WARN Transient transport error when getting program: transport error, retrying...

thread 'main' panicked at src/main.rs:18:49:
called `Result::unwrap()` on an `Err` value: transport error

Caused by:
    no native certs found
```

## Solution

When the error is caused by "no native certs found",  don't retry and return a permanent error:

```
ERROR Permanent error when getting program: no native certs found    

thread 'main' panicked at src/main.rs:18:49:
called `Result::unwrap()` on an `Err` value: transport error

Caused by:
    no native certs found
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sdk/src/network/retry.rs
+++ b/crates/sdk/src/network/retry.rs
@@ -82,7 +82,14 @@ where
                 } else {
                     // Check for common transport errors.
                     let error_msg = e.to_string().to_lowercase();
-                    let is_transient = error_msg.contains("tls handshake") ||
+                    let error_debug_msg = format!("{e:?}");
+
+                    if error_debug_msg.contains("no native certs found") {
+                        log::error!("Permanent error when {}: no native certs found", operation_name);
+                        Err(BackoffError::permanent(e))
+                    }
+                    else {
+                        let is_transient =  error_msg.contains("tls handshake") ||
                         error_msg.contains("dns error") ||
                         error_msg.contains("connection reset") ||
                         error_msg.contains("broken pipe") ||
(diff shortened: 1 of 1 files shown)


leruaa:
I don't think we can, [TlsError is not public](https://github.com/hyperium/tonic/blob/72b0fd59442d71804d4104e313ef6f140ab8f6d1/tonic/src/transport/service/tls.rs#L11).

ratankaliani:
Made a PR to `tonic`: https://github.com/hyperium/tonic/pull/2209.

Can you update your PR description to be more clear? Can you add a TODO that we can parse this with the `TlsError` type once a new version with that Tonic PR is merged?

ratankaliani (approved):
LGTM once you've tested on an Ubuntu machine without `ca-certificates`.

PR #2139, opened 2025-03-08
fix(ci): create ref doesn't trigger CI

nhtyy:
In our recent upgrade, we changed Docker deployments to only trigger on tags, however using `createRef` from github scripts didn't trigger a "push" event.


ratankaliani (approved):
LGTM! This is what you ran for `v4.1.3` right?

nhtyy:
@ratankaliani No did it manually, not confident in this actually, looking into another solution

PR #2141, opened 2025-03-08
fix: correct max_cycles limit check

imikushin:
Fix `max_cycles` limit check when running and proving programs with a context.

## Motivation

When a user runs a program, gets the report, takes `report.total_instruction_count()`, uses it as `max_cycles` and tries to run the same program with it, it does not work. 

## Solution

The check if `max_cycles` was exceeded, is currently off by 1:
```rust
    if self.state.global_clk >= max_cycles {
        return Err(ExecutionError::ExceededCycleLimit(max_cycles));
    }
```

This breaks at the very last legitimate cycle if `max_cycles` is set to the exact number of the program cycles.

This PR fixes the check to be `if self.state.global_clk > max_cycles {`, because 
`self.state.global_clk == max_cycles` is still ok.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/executor/src/context.rs
+++ b/crates/core/executor/src/context.rs
@@ -172,6 +172,7 @@ impl<'a> SP1ContextBuilder<'a> {
     }
 
     /// Set the maximum number of cpu cycles to use for execution.
+    /// `report.total_instruction_count()` will be less than or equal to `max_cycles`.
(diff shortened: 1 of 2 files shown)


nhtyy:
Thanks!

leruaa (approved):

PR #2143, opened 2025-03-10
feat(executor): Add overrides to std{io, err} for guest programs

### LINKED ISSUE #2121: Allow stdout/stderr to redirection during execution
### Component

sp1-zkvm

### Describe the feature you would like

Currently stdout and stderr are [always printed to the console](https://github.com/succinctlabs/sp1/blob/dev/crates/core/executor/src/executor.rs#L2007).

```rust
let prover = SP1Prover::<CpuProverComponents>::new();
let (_, _, _, _) = prover.setup(&elf);
let _ = prover.execute(&elf, &stdin, SP1Context::default()).unwrap();
```

It would be helpful if the stdout/stderr of the executed ELF could be captured, e.g. by passing a `SP1Stdout` (similar to `SP1Stdin`).

### Additional context

_No response_

nhtyy:
Closes #2121

While IO options implements `Clone` for backwards compatibility, we should decide if this really is needed, if it is, we should document that `IOOptions` clone impl resets to `None`, as writers require exclusive access. 

The other option would be to wrap it in a mutex, which is not ideal.

### DIFF
--- a/crates/core/executor/src/syscalls/write.rs
+++ b/crates/core/executor/src/syscalls/write.rs
@@ -57,9 +57,11 @@ impl Syscall for WriteSyscall {
                                 flush_s.into_iter().for_each(|mut line| {
                                     line.push('\n');
 
-                                    writer
-                                        .write_all(line.as_bytes())
-                                        .expect("failed to write to stdout io override");
+                                    if let Err(e) = writer.write_all(line.as_bytes()) {
+                                        tracing::error!(
+                                            "failed to write to stdout io override: {e}"
+                                        );
+                                    }
                                 });
                             }
                             None => {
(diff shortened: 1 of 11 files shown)


crates/core/executor/src/syscalls/write.rs
```diff
@@ -49,17 +49,44 @@ impl Syscall for WriteSyscall {
...
                     if !flush_s.is_empty() {
-                        flush_s.into_iter().for_each(|line| eprintln!("stdout: {}", line));
+                        match rt.io_options.stdout {
+                            Some(ref mut writer) => {
+                                flush_s.into_iter().for_each(|mut line| {
+                                    line.push('\n');
+
+                                    writer
+                                        .write_all(line.as_bytes())
+                                        .expect("failed to write to stdout io override");
```
tqn, line 62: I don't feel great about adding more panic paths to the executor. Since we don't consider stdout/stderr a critically important part of the executor, it may be better to log this with `tracing::error!`, perhaps also with the contents of the buffer.

Similar comments apply to the flushing logic (and of course the stderr part too).

tqn (changes requested):
We don't want to panic when we fail to write to an overridden stdout or stderr.

Otherwise, looks good!

crates/core/executor/src/syscalls/write.rs
nhtyy: fixed with `tracing::error!`

PR #2145, opened 2025-03-12
feat: use blake3 for hashing

leruaa:
## Solution

Adding a `blake3` feature to use the corresponding hashing function instead of sha256.
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/verifier/src/tests.rs
```diff
@@ -103,6 +111,9 @@ fn test_verify_invalid_groth16() {
 #[serial]
 #[test]
 fn test_verify_plonk() {
```
ratankaliani, line 113: Will this test verification with `blake3` E2E in CI? Do we run tests with all features?

leruaa: I added a test with `-F blake3`

crates/verifier/fibonacci-blake3/fibonacci-blake3
ratankaliani, line 1: Why does `fibonacci-blake3` need to be inside of the verifier crate? Can't you just add a feature flag to the `test-artifacts/` crate (where the original Fibonacci is)?

This will pollute the verifier crate unnecessarily.

crates/test-artifacts/programs/fibonacci-blake3/Cargo.toml
ratankaliani, line 1: Add note here for why this needs to be in a separate workspace (specifically that you can't enable the `blake3` flag on `sp1-zkvm` in one of the programs, as it will enable it on all programs.

ratankaliani (approved):
LGTM after adding comment for `test-artifacts`.

crates/test-artifacts/programs/fibonacci-blake3/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+# This crate is in its own workspace because of the `blake3` feature that we don't want enabled
```
ratankaliani, line 1:
```suggestion
# Because of the `blake3` feature enabled on `sp1-zkvm`, this program should be in its own workspace. Otherwise, it changes PV hashing for all programs to `blake3`.
```

PR #2146, opened 2025-03-13
feat: allow to use an existing Moongate server

leruaa:
## Motivation

In order to be able to prove with Cuda from inside a container, we need to be able to specify a Moongate server endpoint instead of starting a container it when creating a Cuda prover.

A possible use case would be with Docker compose that includes the Moongate container.

## Solution

When the `SP1_MOONGATE_ENDPOINT` env variable is provided, use it to connect to the Moongate server instead of creating a container.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cuda/src/lib.rs
+++ b/crates/cuda/src/lib.rs
@@ -106,10 +106,9 @@ pub struct WrapRequestPayload {
 }
 
 impl SP1CudaProver {
-    /// Creates a new [SP1Prover] that runs inside a Docker container and returns a
-    /// [SP1ProverClient] that can be used to communicate with the container.
-    pub fn new() -> Result<Self, Box<dyn StdError>> {
-        let moongate_endpoint = std::env::var("SP1_MOONGATE_ENDPOINT").ok();
+    /// Creates a new [SP1CudaProver] that can be used to communicate with the Moongate server at
+    /// `moongate_endpoint`, or if not provided, create one that runs inside a Docker container.
+    pub fn new(moongate_endpoint: Option<String>) -> Result<Self, Box<dyn StdError>> {
         let reqwest_middlewares = vec![Box::new(LoggingMiddleware) as Box<dyn Middleware>];
 
         let prover = match moongate_endpoint {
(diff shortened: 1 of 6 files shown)


crates/cuda/src/lib.rs
```diff
@@ -104,65 +109,93 @@ impl SP1CudaProver {
...
+                )
+                .expect("failed to create client");
+
+                SP1CudaProver {
+                    client,
+                    managed_container: Some(CudaProverContainer {
+                        name: container_name.to_string(),
+                        cleaned_up: cleaned_up.clone(),
+                    }),
+                }
```
ratankaliani, line 196: You should pull this out into a separate function.

crates/cuda/src/lib.rs
```diff
@@ -104,65 +109,93 @@ impl SP1CudaProver {
...
-            .spawn()
-            .map_err(|e| format!("Failed to start Docker container: {}. Please check your Docker installation and permissions.", e))?;
-
-        // Kill the container on control-c
-        ctrlc::set_handler(move || {
-            tracing::debug!("received Ctrl+C, cleaning up...");
-            if !cleanup_flag.load(Ordering::SeqCst) {
-                cleanup_container(cleanup_name);
-                cleanup_flag.store(true, Ordering::SeqCst);
+        let moongate_endpoint = std::env::var("SP1_MOONGATE_ENDPOINT").ok();
```
ratankaliani, line 112: Instead of using an environment variable, you should have a separate `with_moongate_endpoint` new function that initializes with this environment variable.

I'm generally opposed to adding environment flags in the library unless they're absolutely necessary, as they're more brittle. (e.g. with `SKIP_SIMULATION` and others, we switched to declarative approaches).

ratankaliani (commented):
We shouldn't enshrine the `SP1_MOONGATE_ENDPOINT` environment variable and instead should expose a function for initializing the CUDA prover with an endpoint.

crates/sdk/src/cuda/builder.rs
```diff
@@ -9,9 +9,32 @@ use super::CudaProver;
...
+pub struct CudaProverBuilder {
+    moongate_endpoint: Option<String>,
+}
 
 impl CudaProverBuilder {
+    /// Sets the Moongate server endpoint.
+    ///
+    /// # Details
+    /// This method will set the Moongate server to use, instead using the built-in Moongate server
+    /// that is inside a Docker container,
```
ratankaliani, line 22:
```suggestion
    /// Run the CUDA prover with the provided endpoint for the Moongate (GPU prover) server. Enables more 
    /// customization and avoids DinD configurations.
```

crates/cuda/src/lib.rs
```diff
@@ -125,24 +188,24 @@ impl SP1CudaProver {
...
+                        "all",
+                        "--name",
+                        container_name,
+                        &image_name,
+                    ])
+                    // Redirect stdout and stderr to the parent process
+                    .stdout(Stdio::inherit())
+                    .stderr(Stdio::inherit())
+                    .spawn()
+                    .map_err(|e| format!("Failed to start Docker container: {}. Please check your Docker installation and permissions.", e))?;
```
ratankaliani, line 208: Remove diff here

ratankaliani (approved):
LGTM! Apply the two fixes and feel free to merge.

PR #2149, opened 2025-03-13
fix(build): override rustc only, not the entire toolchain

nhtyy:
Currently, our build commands overrides the entire toolchain, this is an issue since our latest release didnt include `host-tools`.

This PR changes the build command to search for the sysroot and override the rustc bin.


ratankaliani (approved):
LGTM

PR #2157, opened 2025-03-18
chore: fix logging

ratankaliani:
## Motivation

In several places, we emit `info` logs that should instead be `debug`. Users do not need these logs.


tqn (approved):
LGTM!

PR #2165, opened 2025-03-21
fix memory boundary check in decode function.

aso20455:
Replace `addr == MAXIMUM_MEMORY_SIZE` with `addr >= MAXIMUM_MEMORY_SIZE`
to correctly handle addresses that exceed the maximum limit.


leruaa:

Issue #727, opened 2024-05-13
SP1PublicValues handling for the aggregation proof


shriphani:
The aggregation proof serializes the `buffer` of an `SP1PublicValues` as a `Vec<u8>` and sends it over to the guest. If the guest wants to read the actual program output, it needs to use `bincode::read` on this value. Ideally the user shouldn't have to explicitly reference both values. Can we expose `SP1PublicValues` or some other object where we can do the following:

# Host:
```
let proof1 = ...; // an existing proof

// write proof1 public values to guest
aggregation_guest_stdin.write(proof1.public_values);
// also write the vk value
```

# Guest:

```
let proof: SP1PublicValues = sp1_zkvm::io::read();
sp1_zkvm::precompiles::verify::verify_sp1_proof(program_identifier, proof.digest());
let program_output: OutputStruct = proof.output();
```

basically, `pub use SP1PublicValues` so it isn't hidden in sp1_core, and implement a `digest` and  `output` method for it so guests don't have to clone bincode.

irfand29:
I will take the issue

PR #2187, opened 2025-03-30
feat: add SP1PublicValues API for aggregation

DeVikingMark:
## Motivation

Fixing issue #727

## Solution

I've implemented a more ergonomic API for SP1PublicValues:

1. Added two helper methods:
   - `digest()` - Gets the SHA-256 digest ready for verification
   - `output<T>()` - Directly extracts typed program output from public values

2. Re-exported SP1PublicValues in sp1_zkvm so it's directly accessible in guest code

3. Built a working demo example that shows how this improves the workflow for recursive proofs

These changes make recursive proof verification code cleaner while maintaining backward compatibility.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/io.rs
+++ b/crates/primitives/src/io.rs
@@ -77,6 +77,22 @@ impl SP1PublicValues {
         // Return the masked hash as a BigUint.
         BigUint::from_bytes_be(hash.as_slice())
     }
+
+    /// Returns the SHA-256 digest of the public values to be used in verification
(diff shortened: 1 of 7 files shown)


leruaa:
Thanks for you contribution, but we won't merge this for the following reasons:

- `digest()` is a duplicate of the existing `hash()` fn.
- `output()` is kind of a duplicate of the existing `read()` fn, and it won't work if multiples objects are set to the host.

PR #2228, opened 2025-04-08
feat: disallow SP1 to be ran in debug

leruaa:
## Motivation

Teams that run benchmarks might forget the release flag and see degraded performance. To avoid that, we want to disallow provers to be ran in debug in CPU and Cuda mode.

## Solution

Panics when the CPU and Cuda provers are instantiated if `sp1-sdk` is build in debug. It's not an issue with the network prover as progams are executed on a remote cluster.

### DIFF
--- a/crates/sdk/src/utils.rs
+++ b/crates/sdk/src/utils.rs
@@ -40,5 +40,5 @@ pub(crate) fn block_on<T>(fut: impl std::future::Future<Output = T>) -> T {
 /// will be performant, which is important for benchmarking.
 pub(crate) fn check_release_build() {
     #[cfg(debug_assertions)]
-    panic!("sp1-sdk must be built in release mode. please compile with the --release flag.");
+    panic!("sp1-sdk must be built in release mode, please compile with the --release flag.");
 }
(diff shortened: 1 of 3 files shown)


crates/sdk/src/utils.rs
```diff
@@ -35,3 +35,10 @@ pub(crate) fn block_on<T>(fut: impl std::future::Future<Output = T>) -> T {
         rt.block_on(fut)
     }
 }
+
+/// Forbid `sp1-sdk` to be compiled in debug to avoid teams doing benchmarks and forget to build in
+/// release seeing degraded performances.
```
ratankaliani, line 40:
```suggestion
/// Check that SP1 SDK was built in release mode. Ensures that the prover and executor
/// will be performant, which is important for benchmarking.
```

crates/sdk/src/utils.rs
```diff
@@ -35,3 +35,10 @@ pub(crate) fn block_on<T>(fut: impl std::future::Future<Output = T>) -> T {
         rt.block_on(fut)
     }
 }
+
+/// Check that SP1 SDK was built in release mode. Ensures that the prover and executor
+/// will be performant, which is important for benchmarking.
+pub(crate) fn check_release_build() {
+    #[cfg(debug_assertions)]
+    panic!("sp1-sdk must be built in release mode. please compile with the --release flag.");
```
ratankaliani, line 43:
```suggestion
    panic!("sp1-sdk must be built in release mode, please compile with the --release flag.");
```

ratankaliani (approved):
LGTM

PR #2232, opened 2025-04-09
Fix potential panic in `debug_cumulative_sums` when permutation matrix is empty

Pronoss:
## Motivation

This PR addresses a bug in `debug_cumulative_sums` that can cause a runtime panic when the permutation matrix has no columns (i.e. `width == 0`), such as in chips with no interactions.

Previously, the function used `.last().unwrap()` without checking if the row was empty. This would panic if the row contained no elements, which is possible in practice.

## Solution

- Replaced `.last().unwrap()` with `.last().unwrap_or(&EF::zero())`, making the function robust to empty rows.
- Added an explicit check for both `width == 0` and `height == 0` to safely return `EF::zero()` when the permutation matrix is effectively empty.
- Improved the assertion by adding a helpful error message to `assert_eq!`.

This ensures that the debug logic does not panic on valid edge cases and improves the safety and clarity of the code.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes — **No**
- [x] Safe fallback behavior for empty matrices


leruaa:

PR #2234, opened 2025-04-11
test: add secp256k1 v0.30.0 patch testing

leruaa:
## Motivation

We patched `secp256k1` v0.30.0, and we need to check the patch work properly.

### DIFF
--- a/patch-testing/secp256k1/src/lib.rs
+++ b/patch-testing/secp256k1/src/lib.rs
@@ -1,12 +1,16 @@
 use secp256k1::{Message, PublicKey, Secp256k1};
 
 #[sp1_test::sp1_test("secp256k1_recover", syscalls = [SECP256K1_DOUBLE, SECP256K1_ADD])]
-fn test_recover(stdin: &mut sp1_sdk::SP1Stdin) -> impl FnOnce(sp1_sdk::SP1PublicValues) {
+fn test_recover_rand_lte_100(
+    stdin: &mut sp1_sdk::SP1Stdin,
+) -> impl FnOnce(sp1_sdk::SP1PublicValues) {
     recover_rand_lte_100(stdin)
 }
 
 #[sp1_test::sp1_test("secp256k1_recover_v0_30_0", syscalls = [SECP256K1_DOUBLE, SECP256K1_ADD])]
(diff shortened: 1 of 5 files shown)


ratankaliani (commented):
Should we have a separate directory inside of `program/bin` for every version of the `secp256k1` crate that we patch?

Given that we want to confirm each version actually patches here (via the `sp1-test` macro), we should have separate programs and check the invocations with macro's separately too.

patch-testing/secp256k1/src/lib.rs
```diff
@@ -47,6 +58,18 @@ fn test_recover_rand_lte_100(
...
+}
+
+#[sp1_test::sp1_test("secp256k1_verify_v0_30_0", syscalls = [SECP256K1_DOUBLE, SECP256K1_ADD])]
+fn test_verify_v0_30_0_rand_lte_100(
+    stdin: &mut sp1_sdk::SP1Stdin,
+) -> impl FnOnce(sp1_sdk::SP1PublicValues) {
+    verify_rand_lte_100(stdin)
+}
+
+#[allow(dead_code)]
```
ratankaliani, line 71: Why is this `dead_code` necessary?

patch-testing/secp256k1/src/lib.rs
```diff
@@ -1,10 +1,21 @@
...
+}
+
+#[sp1_test::sp1_test("secp256k1_recover_v0_30_0", syscalls = [SECP256K1_DOUBLE, SECP256K1_ADD])]
+fn test_recover_v0_30_0_rand_lte_100(
+    stdin: &mut sp1_sdk::SP1Stdin,
+) -> impl FnOnce(sp1_sdk::SP1PublicValues) {
+    recover_rand_lte_100(stdin)
+}
+
+#[allow(dead_code)]
```
ratankaliani, line 17: Why is this `dead_code` necessary? Because the functions are tests?

ratankaliani (approved):
LGTM! Can you add a comment for why `dead_code` is necessary?

patch-testing/secp256k1/src/lib.rs
leruaa: Exactly

PR #2237, opened 2025-04-13
Remove debug-assertions flag from Rust compilation flags

leopardracer:
This PR removes the `-Cdebug-assertions` flag from RUSTFLAGS environment variable across all workflow files. This change helps to:

- Improve performance by removing runtime assertion checks
- Maintain consistent behavior between debug and release builds
- Reduce potential false positives in tests

The flag was removed from the following files:
- .github/workflows/docker-gnark.yml
- .github/workflows/main.yml
- .github/workflows/pr.yml


nhtyy:
Debug assertions are actually useful in test workflows

PR #2239, opened 2025-04-14
fix: concurrent mutation of opts by cloning config per thread

mdqst:
## Motivation

the previous implementation mutated the shared `opts` configuration inside a loop that spawns parallel Rayon threads.
this introduced a risk of concurrent mutation, which could lead to undefined behavior or subtle bugs during shape collection.

## Solution

cloned the `opts` variable per iteration and passed a thread-local `local_opts` into each spawned task. this ensures thread safety by avoiding shared mutable state across threads.

the logic remains unchanged, and original comments were preserved for clarity.

## PR Checklist

- [ ] Added Tests  
- [ ] Added Documentation  
- [ ] Breaking changes


leruaa:
`opts` is not shared because of `move` in `rayon::spawn()`closure.

PR #2240, opened 2025-04-14
refactor(patch-testing): `sha2`, `rsa` versions

leruaa:
## Solution

* Update `rsa` tag to latest
* Test all `sha2` patched versions separately

### DIFF
--- a/patch-testing/sha/src/lib.rs
+++ b/patch-testing/sha/src/lib.rs
@@ -1,14 +1,32 @@
-#[cfg(test)]
+use sha2::Digest;
 use sp1_sdk::SP1PublicValues;
 use sp1_test::sp1_test;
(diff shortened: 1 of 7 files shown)


patch-testing/sha/src/lib.rs
```diff
@@ -37,13 +52,16 @@ fn test_sha2_expected_digest_lte_100_times(
...
-            let committed = public.read::<([u8; 32], [u8; 32])>();
+            let committed = public.read::<[u8; 32]>();
 
             assert_eq!(digest, committed);
         }
     }
 }
 
+#[allow(dead_code)]
+fn build_digests() {}
```
ratankaliani, line 63: Can we remove this?

PR #2241, opened 2025-04-14
refactor: simplify benchmarking logic with `run_and_report` helper

famouswizard:
## Motivation  
There was noticeable duplication in the `run_*` blocks, making the code harder to maintain and read. Also, frequency calculation logic was inline and could benefit from being more explicit.

## Solution  
Refactored the repeated logic into a shared `run_and_report` function.  
Added `calculate_mhz` for cleaner and more understandable frequency calculations.  
As a result, the `match` block is now more concise and easier to follow.

## PR Checklist

- [ ] Added Tests  
- [ ] Added Documentation  
- [ ] Breaking changes


leruaa:

Issue #2247, opened 2025-04-17
Cuda proving fails without `setup()` call


koxu1996:
This applies to the GPU prover. According to the documentation, `client.setup()` returns the proving key (PK) for a given ELF, which can then be used to call `client.prove(pk, ...)`. Since it doesn't make sense to generate a new PK for every proving call (the client binary remains the same), I serialized the PK to disk and loaded it at program start.

Surprisingly, calling `client.prove(pk_deserialized, ...)` directly - without a prior setup call - fails with the following error:

```
thread 'main' panicked at /home/ubuntu/.cargo/git/checkouts/sp1-3cc27ff66a7c030c/45f36d1/crates/cuda/src/lib.rs:273:82:
called `Result::unwrap()` on an `Err` value: TwirpError(TwirpErrorResponse { code: Internal, msg: "proving key not provided", meta: {} })
```

---

As a workaround, I call `setup()` and discard the returned PK, then call `prove()` with my deserialized PK. This works, but it unnecessarily wastes a few seconds.

koxu1996:
I confirm that setup-less proving works with Moongate v4.2.1.

PR #2251, opened 2025-04-18
fix(cuda): Send PK in prove_core request (#2247)

gap-editor:
## Motivation

This PR addresses issue #2247, where CUDA proving fails when using a deserialized proving key (PK) without a preceding `client.setup()` call in the same session. The error `TwirpError(TwirpErrorResponse { code: Internal, msg: "proving key not provided", meta: {} })` occurs because the remote CUDA prover service (`moongate`) likely requires initialization associated with the PK, which normally happens during the `setup` call. Calling `prove_core` directly with a deserialized PK skips this remote initialization.

The current workaround involves calling `setup()`, discarding the result, and then calling `prove()` with the deserialized PK, which introduces unnecessary overhead.

## Solution

This PR modifies the client-side CUDA prover (`SP1CudaProver` in `crates/cuda/src/lib.rs`) to send the proving key directly within the `prove_core` request payload.

1.  **`ProveCoreRequestPayload` struct:** Added a `pk: SP1ProvingKey` field to include the proving key in the request sent to the remote service.
2.  **`prove_core` method:**
    *   Updated the method signature to accept `pk: &SP1ProvingKey` as an argument.
    *   Modified the method body to populate the `pk` field in the `ProveCoreRequestPayload`.

This change allows the remote `moongate` service (once updated accordingly) to receive and use the specific proving key required for the `prove_core` operation directly from the request, eliminating the dependency on prior `setup()` calls within the same session for PK initialization.

**Note:** This PR only contains the client-side changes. Corresponding changes are required in the `moongate` server implementation to accept and utilize the `pk` from the `ProveCoreRequestPayload`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes:
    - The public function signature of `SP1CudaProver::prove_core` has changed. It now requires an additional `pk: &SP1ProvingKey` argument:
        ```diff
        - pub fn prove_core(&self, stdin: &SP1Stdin) -> Result<SP1CoreProof, SP1CoreProverError>
        + pub fn prove_core(&self, pk: &SP1ProvingKey, stdin: &SP1Stdin) -> Result<SP1CoreProof, SP1CoreProverError>
        ```
    - The structure of `ProveCoreRequestPayload` has changed, adding a `pk` field. This affects the serialization format used for communication with the remote prover service.
    - This change requires a corresponding update in the remote `moongate` prover service to handle the modified request payload and utilize the provided `pk`. Without the server-side update, calls to `prove_core` will likely fail.

### DIFF
--- a/crates/cuda/src/lib.rs
+++ b/crates/cuda/src/lib.rs
@@ -72,6 +72,8 @@ pub struct SetupResponsePayload {
 /// We use this object to serialize and deserialize the payload from the client to the server.
 #[derive(Serialize, Deserialize)]
 pub struct ProveCoreRequestPayload {
+    /// The proving key.
+    pub pk: SP1ProvingKey,
     /// The input stream.
     pub stdin: SP1Stdin,
 }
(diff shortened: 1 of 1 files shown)


gap-editor:
@nhtyy whats wrong with my PR?

koxu1996:
@gap-editor SP1 project is just a joke, and IMO cheap rip-off of RISC0.

PR #2260, opened 2025-04-25
feat: allow to profile inline functions

leruaa:
## Motivation

It's sometimes useful to be able to have inlines function shown in the call stack when profiling a program

## Solution

Build the ELF with debug info, extract them and include them to the profiling json file.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/executor/src/profiler.rs
+++ b/crates/core/executor/src/profiler.rs
@@ -1,8 +1,14 @@
 use gecko_profile::{Frame, ProfileBuilder, StringIndex, ThreadBuilder};
+use gimli::{
+    AttributeValue, DW_AT_MIPS_linkage_name, DW_AT_high_pc, DW_AT_linkage_name, DW_AT_name,
+    DW_AT_ranges, DW_TAG_inlined_subroutine, DebugInfoOffset, DebuggingInformationEntry, DwAt,
+    Dwarf, EndianRcSlice, Range, Reader, RunTimeEndian, SectionId, Unit, UnitOffset,
+};
 use goblin::elf::{sym::STT_FUNC, Elf};
 use indicatif::{ProgressBar, ProgressStyle};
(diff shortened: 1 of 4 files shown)


crates/core/executor/src/executor.rs
```diff
@@ -2319,16 +2350,48 @@ impl<'a> Executor<'a> {
...
+    #[inline]
+    fn profile(&mut self, i: &Instruction, previous_pc: u64) {
         if let Some((ref mut profiler, _)) = self.profiler {
             if !self.unconstrained {
-                profiler.record(self.state.global_clk, self.state.pc as u64);
-            }
-        }
+                let stack_event = match i.opcode {
+                    // A JAL instruction should push the return address onto a return-address
+                    // stack (RAS) only when rd=x1/x5.
```
ratankaliani, line 2366:
```suggestion
                    // stack (RAS) only when rd=x1 or rd=x5.
```

crates/core/executor/src/profiler.rs
```diff
@@ -71,76 +132,116 @@ impl Profiler {
...
+                if let Some(f) = self.start_lookup.get(&pc) {
+                    // Jump to a new function.
+                    let (_, _, name) = self.function_ranges.get(*f).unwrap();
+                    self.function_stack.push(Function::regular(name.clone()));
+                    self.pop_stack.push(previous_pc);
                 }
             }
+        }
+
+        // Pop inline functions when the current PC is not in theirs range
```
ratankaliani, line 216:
```suggestion
        // Pop inline functions when the current PC is not in the range of the function stack.
```

quietBlockchain:
Thanks for this improvement!
Profiling inline functions seems very useful for understanding performance bottlenecks more precisely.
Just curious do you see any impact on compilation time or binary size with this change?

PR #2264, opened 2025-04-29
feat: stateless proving with Moongate

### LINKED ISSUE #2247: Cuda proving fails without `setup()` call

leruaa:
## Motivation

When proving with Cuda, calling `prove()` only work with the last program that was registered with `setup()`.

## Solution

Sends the ELF with each `prove()` request, and lazily call `setup()`on the Moongate server if the program is not already registered. 

Closes #2247, #2180

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cuda/src/lib.rs
+++ b/crates/cuda/src/lib.rs
@@ -79,8 +79,7 @@ pub struct ProveCoreRequestPayload {
 /// The payload for the [sp1_prover::SP1Prover::stateless_prove_core] method.
 ///
 /// We use this object to serialize and deserialize the payload from the client to the server.
-/// The ELF is included in the payload to allow to build the program and the pk on the Moongate
-/// server
+/// The proving key is sent in the payload with the request to allow the Moongate server to generate proofs without re-generating the proving key. 
 #[derive(Serialize, Deserialize)]
 pub struct StatelessProveCoreRequestPayload {
     /// The input stream.
--- a/crates/cuda/src/lib.rs
+++ b/crates/cuda/src/lib.rs
@@ -76,6 +76,19 @@ pub struct ProveCoreRequestPayload {
     pub stdin: SP1Stdin,
 }
 
+/// The payload for the [sp1_prover::SP1Prover::stateless_prove_core] method.
+///
+/// We use this object to serialize and deserialize the payload from the client to the server.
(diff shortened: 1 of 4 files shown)


crates/cuda/src/lib.rs
```diff
@@ -76,6 +76,19 @@ pub struct ProveCoreRequestPayload {
     pub stdin: SP1Stdin,
 }
 
+/// The payload for the [sp1_prover::SP1Prover::stateless_prove_core] method.
+///
+/// We use this object to serialize and deserialize the payload from the client to the server.
+/// The ELF is included in the payload to allow to build the program and the pk on the Moongate
+/// server
```
ratankaliani, line 83:
```suggestion
/// The proving key is sent in the payload with the request to allow the Moongate server to generate proofs without re-generating the proving key. 
```

crates/cuda/src/lib.rs
```diff
@@ -76,6 +76,18 @@ pub struct ProveCoreRequestPayload {
     pub stdin: SP1Stdin,
 }
 
+/// The payload for the [sp1_prover::SP1Prover::stateless_prove_core] method.
+///
+/// We use this object to serialize and deserialize the payload from the client to the server.
```
ratankaliani, line 81:
```suggestion
/// This object is used to serialize and deserialize the payloads for the Moongate server.
```

ratankaliani (approved):
LGTM!

PR #2267, opened 2025-04-30
feat: give access to cycle count when using CudaProver

leruaa:
## Motivation

For applications like ETHProofs, we don't want to execute twice to get the cycle count and it'd be nice to return it to reduce our latency.

## Solution

Add a `prove_with_cycles()` fn on `CudaProver` that returns a tuple with the proof and the cycle count.

### DIFF
--- a/crates/sdk/src/cuda/mod.rs
+++ b/crates/sdk/src/cuda/mod.rs
@@ -81,7 +81,7 @@ impl CudaProver {
     /// Proves the given program on the given input in the given proof mode.
     ///
     /// Returns the cycle count in addition to the proof.
-    pub fn prove_with_report(
+    pub fn prove_with_cycles(
         &self,
         pk: &SP1ProvingKey,
         stdin: &SP1Stdin,
@@ -89,6 +89,7 @@ impl CudaProver {
     ) -> Result<(SP1ProofWithPublicValues, u64)> {
         // Generate the core proof.
         let proof = self.cuda_prover.prove_core(stdin)?;
+        // TODO: Return the prover gas
         let cycles = proof.cycles;
         if kind == SP1ProofMode::Core {
             let proof_with_pv = SP1ProofWithPublicValues::new(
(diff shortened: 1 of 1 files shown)


crates/sdk/src/cuda/mod.rs
```diff
@@ -77,32 +77,26 @@ impl CudaProver {
...
-
-    fn inner(&self) -> &SP1Prover<CpuProverComponents> {
-        &self.cpu_prover
-    }
-
-    fn prove(
+    /// Proves the given program on the given input in the given proof mode.
+    ///
+    /// Returns the cycle count in addition to the proof.
+    pub fn prove_with_report(
```
ratankaliani, line 84:
```suggestion
    pub fn prove_with_cycles(
```

crates/sdk/src/cuda/mod.rs
```diff
@@ -150,17 +146,38 @@ impl Prover<CpuProverComponents> for CudaProver {
...
+        &self.cpu_prover
+    }
+
+    fn prove(
+        &self,
+        pk: &SP1ProvingKey,
+        stdin: &SP1Stdin,
+        kind: SP1ProofMode,
+    ) -> Result<SP1ProofWithPublicValues> {
+        self.prove_with_report(pk, stdin, kind).map(|(p, _)| p)
```
ratankaliani, line 177:
```suggestion
        self.prove_with_cycles(pk, stdin, kind).map(|(p, _)| p)
```

crates/sdk/src/cuda/mod.rs
```diff
@@ -77,32 +77,26 @@ impl CudaProver {
...
+    /// Returns the cycle count in addition to the proof.
+    pub fn prove_with_report(
         &self,
         pk: &SP1ProvingKey,
         stdin: &SP1Stdin,
         kind: SP1ProofMode,
-    ) -> Result<SP1ProofWithPublicValues> {
+    ) -> Result<(SP1ProofWithPublicValues, u64)> {
         // Generate the core proof.
         let proof = self.cuda_prover.prove_core(stdin)?;
```
ratankaliani, line 91: Can you add a note here that in a future version we could return the gas as well from `prove_core`?

ratankaliani (approved):
LGTM

Issue #2289, opened 2025-05-12
Potentially misleading warning: "Not all proofs were read. Proving will fail during recursion."


liuchengxu:
### Component

sp1-sdk

### Have you ensured that all of these are up to date?

- [ ] SP1 SDK
- [ ] cargo prove CLI/sp1up

### Operating System

macOS (Apple Silicon)

### Describe the bug

I encountered the following recurring warning `Not all proofs were read.` when generating an aggregation proof. However, despite the warning, the proof was successfully generated and verified later.

I patched the sp1 to also print the info `self.state.proof_stream_ptr=1 proof_stream_len=0`, it seems that the proof stream read is somehow incomplete.

```
2025-05-12T12:17:02.325522Z  WARN prove_core: sp1_core_executor::executor: Not all proofs were read. Proving will fail during recursion. Did you pass too
        many proofs in or forget to call verify_sp1_proof? self.state.proof_stream_ptr=1 proof_stream_len=0
2025-05-12T12:20:48.885794Z  WARN prove_core: sp1_core_executor::executor: Not all proofs were read. Proving will fail during recursion. Did you pass too
        many proofs in or forget to call verify_sp1_proof? self.state.proof_stream_ptr=1 proof_stream_len=0
```

Let me know if logs or minimal repro code would help. Happy to provide more info.

leruaa:
Thanks for opening this issue! Yes, I would appreciate a minimal repro.

Kalpazanova53:
Hi I can try to reproduce this issue and investigate it further if that would be helpful. Let me know if you'd be open to a contribution on this.

liuchengxu:
Here is the [repo](https://github.com/liuchengxu/repro-sp1-2289) to reproduce this issue, the command is in the README.md

PR #2299, opened 2025-05-18
fix: correct misleading proof verification warning on macOS

avorylli:
## Motivation

Issue #2289 reports a potentially misleading warning "Not all proofs were read" on macOS (Apple Silicon) that doesn't appear on Ubuntu. 

## Solution

Fixed the issue by only incrementing the proof_stream_ptr counter when verification actually occurs. 

Previously, the counter was always incremented regardless of whether verification happened, causing a mismatch between the number of processed proofs and the counter value.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


liuchengxu:
Very interesting that this issue occurred only on macOS, but Ubuntu is fine if the glitch is OS-unrelated.

PR #2327, opened 2025-06-17
Allow option to enable setting the GPU id and port while starting moongate server

pen-pal:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


leruaa:
This is already implemented in SP1 with `local()` fn on `CudaProverBuilder`, see here: https://docs.rs/sp1-sdk/5.0.6/sp1_sdk/cuda/builder/struct.CudaProverBuilder.html#method.local

It allow to configure the port and visible GPU devices

PR #2333, opened 2025-06-19
feat: whitelist feature on sepolia

jtguibas:


quietBlockchain:
Great addition! Using Nix with VSCode in a Dockerfile is a thoughtful setup makes onboarding much smoother. Thanks for this!

Issue #2338, opened 2025-06-24
Log some progress indicator for downloading files and docker images


nuke-web3:
### Component

sp1-sdk

### Describe the feature you would like

When using `cargo prove build --docker` I don't see any CPU activity or logging happening, so assumed something was amiss. Turns out that I noticed my network monitor spiking and figured it was the sdk internally pulling the (quite large) sp1 builder image.

https://github.com/succinctlabs/sp1/blob/6544380d8d71640929c862a65174613c6274baf8/crates/cuda/src/lib.rs#L218-L221

I was on a slower network, so this took _minutes_ to complete.

Also related to #2255 in a way: it was hard to discover why things were hanging after resolving the panic... that file download also took order of minutes for me.

I would like a way to get some indication that a docker image and/or file is being downloaded, if possible, piping the log stream of `docker pull` so it's visible.

Seems this _should_ be possible as IIUC `stderr` is how things are generally communicated in the build like here:
https://github.com/succinctlabs/sp1/blob/6544380d8d71640929c862a65174613c6274baf8/crates/sdk/src/install.rs#L108

### Additional context

Very-nice-to-have: A TUI progress indicator on downloads, perhaps using something like https://docs.rs/indicatif/latest/indicatif/ with streamed download via `reqwest::Response::bytes_stream()` to `stderr` if possible.

nuke-web3:
Example: when missing groth16 artifact, it downloads and prints this nice progress bar
![Image](https://github.com/user-attachments/assets/e5471cdc-3825-4190-b600-7b3328b01110)

quietBlockchain:
Thanks for raising this! A visible progress indicator would definitely improve the UX when pulling large Docker images or downloading files, especially on slower connections.

If it's helpful, the indicatif crate could be a great addition here. It supports customizable progress bars and integrates well with stderr or stdout. Piping the output of docker pull through it (or showing a minimal spinner) could give users confidence that the process is active and not stalled.

Let me know if this is something you'd like help with  happy to explore a small PR for it.

PR #2339, opened 2025-06-24
`docker pull` inherits `stderr` and `stdout`, so docker logs are printed

nuke-web3:
## Motivation

Partial fix for #2338 .

## Solution

https://doc.rust-lang.org/std/process/struct.Command.html#method.status
> stdout and stderr are inherited from the parent.

## PR Checklist

- [ ] ~Added Tests~ N/A
- [ ] ~Added Documentation~ N/A
- [ ] ~Breaking changes~ N/A

(lmk if anything else is needed :pray:)

### DIFF
--- a/crates/cuda/src/lib.rs
+++ b/crates/cuda/src/lib.rs
@@ -216,9 +216,15 @@ impl SP1CudaProver {
         }
 
         // Pull the docker image if it's not present
-        if let Err(e) = Command::new("docker").args(["pull", &image_name]).output() {
-            return Err(format!("Failed to pull Docker image: {}. Please check your internet connection and Docker permissions.", e).into());
+        let status = Command::new("docker")
+            .args(&["pull", &image_name])
+            .status()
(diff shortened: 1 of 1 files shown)


nuke-web3:
Well it seems this is not helpful for the original issue, although might expose the cuda docker pull logs that would be nice (yet to test though)

Digging deeper into the actual issue, it seems like we already _should_ print the `docker run` command (that internally pulls the required image) with piping:

https://github.com/succinctlabs/sp1/blob/6544380d8d71640929c862a65174613c6274baf8/crates/build/src/command/utils.rs#L78-L112
but here I think progress bars etc. are not captured via `piped`, you need https://doc.rust-lang.org/std/process/struct.Stdio.html#method.inherit to get that I think, but then wouldn't be able to prefix I think... :thinking: 

indicates we might need separate logic to check the if image to use if not pulled, and pull it if so (that we can `inherit` to show progress I think)

wangwa791:
Good

PR #2348, opened 2025-07-10
feat: enforce gas limit

0xernesto:
## Motivation
The prover network still charges self-serve/hosted customers using cycle-based pricing. The network has been updated to use gas-based pricing, but during testing, it was clear that gas limit is not enforced in the executor.

## Solution

Update the executor and related functions so gas can be checked every shard, and stop execution if gas limit is exceeded.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Mozzy59:
seems OK

PR #2354, opened 2025-07-11
refactor: improve error handling and optimize memory allocation

reject-i:
## Motivation

This PR improves error handling and memory allocation in two key areas:

1. The `parse_rustc_version` function currently uses `expect()` which can cause panics with unclear error messages when encountering unexpected rustc version formats. This makes debugging difficult and reduces reliability.

2. The `chunk_vec` function creates unnecessary memory allocations when processing large vectors, which can impact performance in memory-intensive operations.

## Solution

1. Enhanced `parse_rustc_version` to return a proper `Result` type instead of using `expect()`, providing more informative error messages and allowing callers to handle errors gracefully.

2. Optimized `chunk_vec` with pre-allocation strategies:
   - Pre-allocating the result vector with the correct capacity
   - Pre-allocating each chunk with its exact size
   - Using `extend` instead of `collect` to avoid intermediate allocations

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/build/src/build.rs
+++ b/crates/build/src/build.rs
@@ -41,7 +41,7 @@ pub fn execute_build_program(
     let cmd = if args.docker {
         create_docker_command(args, &program_dir, &program_metadata)?
     } else {
-        create_local_command(args, &program_dir, &program_metadata)
+        create_local_command(args, &program_dir, &program_metadata)?
     };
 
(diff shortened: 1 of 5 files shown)


leruaa:

PR #2361, opened 2025-07-14
fix(network): Correct request_proof body and add robust error handling

Alphena-EK:
This PR addresses two main issues in the NetworkClient: a critical compilation bug and fragile error handling.

Problem:

The request_proof function would fail to compile when the sepolia feature was not enabled, as the RequestProofRequestBody was missing several required fields in the else block.

The NetworkClient::new constructor would panic if provided with an invalid private key, making the client less robust.

Changes:

Bug Fix: The else branch in request_proof now correctly populates all required fields for RequestProofRequestBody, ensuring the code compiles correctly with or without the sepolia feature.

Refactor: NetworkClient::new now returns a Result<Self> instead of panicking. This allows callers to gracefully handle initialization errors.

Test: The corresponding unit test has been updated to assert is_ok() on the Result returned by new().

### DIFF
--- a/crates/sdk/src/network/client.rs
+++ b/crates/sdk/src/network/client.rs
@@ -9,7 +9,7 @@ use std::{
 };
 
 use alloy_primitives::{Address, B256, U256};
-use anyhow::{Context, Ok, Result};
+use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use k256::ecdsa::SigningKey;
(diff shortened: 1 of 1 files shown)


leruaa:
Closing because the fix in this PR is incorrect, and `sp-sdk` doesn't fail to compile either with `sepolia` feature on or off.

there are 2 different `RequestProofRequestBody` types, with different fields.

PR #2362, opened 2025-07-14
fix(workspace): Align internal crate versions with workspace version

Alphena-EK:
This PR resolves a critical version inconsistency within the main Cargo.toml file.

Problem:

The main workspace package was defined as version 5.0.8, but all internal sp1-* path dependencies were still pinned to 5.0.5. This 
mismatch could lead to several issues, including:

Potential build failures or unexpected behavior with cargo tooling.
The risk of crates using outdated code from their internal dependencies.
Confusion for developers about the project's current and correct version.

Solution:

All sp1-* versions within [workspace.dependencies] have been updated from 5.0.5 to 5.0.8.
This change ensures that all crates in the workspace are consistent with the declared package version, guaranteeing that the project builds with the intended dependencies and maintaining a clean, predictable state.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace.package]
+# Workspace'in ana versiyonu
 version = "5.0.8"
 edition = "2021"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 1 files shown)


nhtyy:
Duplicate of #2360

PR #2363, opened 2025-07-14
Remove strategy.fail-fast from check-verifier-no-std job (no matrix)

Galoretka:
The check-verifier-no-std job in .github/workflows/pr.yml used a `strategy.fail-fast: false` block without defining a matrix. In GitHub Actions, the `strategy` key (and its `fail-fast` option) is only valid when a `matrix` is present. Using `strategy` without a matrix causes a workflow syntax error and prevents the workflow from running.

This commit removes the unnecessary `strategy` block from the check-verifier-no-std job, making the workflow valid and ensuring it runs as expected.

Reference: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -327,8 +327,6 @@ jobs:
     name: Check Verifier `no_std`
     runs-on: [runs-on, runner=16cpu-linux-x64, disk=large, "run-id=${{ github.run_id }}"]
     timeout-minutes: 10
-    strategy:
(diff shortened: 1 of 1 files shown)


leruaa:

Galoretka:
> As mentioned in the pull request template, trivial PRs are generally not accepted to not encourage PR farming.

this is nooooooot sooooooo triviaaaaaaal

PR #2373, opened 2025-07-19
Fix workspace parent directory check in Docker build command

GarmashAlex:
Previously, the code checked whether the workspace root from cargo metadata was a subdirectory of the user-specified workspace directory. This logic was incorrect, as it did not actually verify that the program directory being built was inside the workspace. As a result, the check could pass even if the program was outside the workspace, or fail for valid configurations, leading to potential build errors or unexpected behavior when mounting directories in Docker.
This commit fixes the logic by ensuring that the canonicalized program directory is a subdirectory of the workspace root. The error message was also updated to accurately reflect the directories being compared. With this change, Docker builds will only proceed if the program is truly located within the specified workspace, preventing misconfigurations and potential build failures due to incorrect directory relationships.

### DIFF
--- a/crates/build/src/command/docker.rs
+++ b/crates/build/src/command/docker.rs
@@ -46,10 +46,10 @@ pub(crate) fn create_docker_command(
         .unwrap_or_else(|| program_metadata.workspace_root.clone());
 
     // Ensure the workspace directory is parent of the program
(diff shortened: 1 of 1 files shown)


leruaa:

PR #2376, opened 2025-07-21
refactor: remove deprecated if/else DSL logic

Lil-Duckling-22:
Removed IfBuilder and its associated logic from builder.rs as it is no longer used.

- Remove IfBuilder struct and IfCondition enum
- Remove unused Builder::if_eq() and Builder::if_ne() methods  
- Remove deprecated condition() method with unimplemented! macro


leruaa:

PR #2393, opened 2025-07-26
fix(build): respect existing CARGO_TARGET_DIR if already set

quietBlockchain:
## Motivation

The `CARGO_TARGET_DIR` was being unconditionally set to `program_metadata.target_directory`, which could override user-defined values. This behavior may conflict with workflows that rely on custom build directories or caching strategies (e.g., CI/CD pipelines).

Respecting user-defined environment variables enhances flexibility and aligns with expected Cargo behavior.

## Solution

Updated the logic to first check if `CARGO_TARGET_DIR` is already set via `std::env::var`. If so, that value is used. If not, the fallback is to use the default `program_metadata.target_directory`.

This allows external tooling or users to define their own target directory without being silently overridden.

## PR Checklist

- [x] Added Tests *(not applicable – small behavior tweak in build env setup)*
- [x] Added Documentation *(not needed – code is self-explanatory and localized)*
- [x] Breaking changes *(No)*

### DIFF
--- a/crates/build/src/command/local.rs
+++ b/crates/build/src/command/local.rs
@@ -86,7 +86,11 @@ pub(crate) fn create_local_command(
         .env("CARGO_ENCODED_RUSTFLAGS", get_rust_compiler_flags(args, &parsed_version))
         .env_remove("RUSTC")
         .env("RUSTC", rustc_bin.display().to_string())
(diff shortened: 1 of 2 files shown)


quietBlockchain:
Let me know if anything needs to be adjusted or further clarified. Thanks for maintaining this great repo. 🙏

PR #2394, opened 2025-07-26
Fix Critical Compilation Errors in SP1 Precompile Integration

savvar9991:
### Adding Semicolon in default_syscall_map
In Rust, a semicolon is mandatory to terminate expressions within a function body. Without it, the compiler will throw a syntax error, preventing compilation.
### Replacing Arch::new with Arc::new
Using Arch::new instead of Arc::new causes a compilation error, as Arch is incompatible with the expected type Arc<dyn Syscall>, blocking the registration of the precompile.


leruaa:

PR #2401, opened 2025-07-31
feat: network signer

0xernesto:
## Motivation

Off-chain network services use KMS keys. This introduces a problem for the services that use the SP1 SDK, where they need a non-KMS key to use the SDK. If the SDK supports KMS, then these services will no longer need to have two separate keys.

## Solution

Add `signer` to the network builder, while keeping backwards compatibility with `private_key`.

Builder code logic:
If `signer` is provided, use it, otherwise, use `private_key` to create a local signer.

### Example Usage
Using local signer
```rust
use sp1_sdk::{NetworkSigner, ProverClient};

let private_key = "...";
let signer = NetworkSigner::local(private_key).unwrap();
let prover = ProverClient::builder().network().signer(signer).build();
```

Using KMS signer
```rust
use sp1_sdk::{NetworkSigner, ProverClient};

let kms_key_arn = "arn:aws:kms:us-east-1:123456789:key/key-id";
let signer = NetworkSigner::aws_kms(kms_key_arn).await.unwrap();
let prover = ProverClient::builder().network().signer(signer).build();
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sdk/src/network/signer.rs
+++ b/crates/sdk/src/network/signer.rs
@@ -120,6 +120,18 @@ impl NetworkSigner {
     }
 }
 
+impl From<String> for NetworkSigner {
+    fn from(private_key: String) -> Self {
+        NetworkSigner::local(&private_key).expect("Failed to parse private key")
+    }
+}
+
+impl From<&str> for NetworkSigner {
+    fn from(private_key: &str) -> Self {
+        NetworkSigner::local(private_key).expect("Failed to parse private key")
+    }
+}
+
 /// Extract AWS region from a KMS ARN-formatted string.
 fn extract_region_from_kms_arn(arn: &str) -> Result<String, NetworkSignerError> {
     let parts: Vec<&str> = arn.split(':').collect();
--- a/crates/sdk/src/network/prover.rs
+++ b/crates/sdk/src/network/prover.rs
@@ -54,10 +54,19 @@ impl NetworkProver {
     /// Creates a new [`NetworkProver`] with the given signer.
     ///
     /// # Details
-    /// * `signer`: The network signer to use for signing requests.
+    /// * `signer`: The network signer to use for signing requests. Can be a `NetworkSigner`,
+    ///   private key string, or anything that implements `Into<NetworkSigner>`.
     /// * `rpc_url`: The rpc url to use for the prover network.
     ///
-    /// # Example
+    /// # Examples
+    /// Using a private key string:
+    /// ```rust,no_run
+    /// use sp1_sdk::NetworkProver;
+    ///
+    /// let prover = NetworkProver::new("0x...", "...");
+    /// ```
+    ///
+    /// Using a `NetworkSigner`:
     /// ```rust,no_run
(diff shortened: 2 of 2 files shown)


crates/sdk/src/network/client.rs
```diff
@@ -77,30 +77,9 @@ impl RetryableRpc for NetworkClient {
...
-
 impl NetworkClient {
     /// Creates a new [`NetworkClient`] with the given private key and rpc url.
-    pub fn new(private_key: impl Into<String>, rpc_url: impl Into<String>) -> Self {
-        let pk = private_key.into();
-        let private_key_bytes =
-            hex::decode(pk.strip_prefix("0x").unwrap_or(&pk)).expect("Invalid private key");
-        let signer = SigningKey::from_slice(&private_key_bytes).expect("Invalid private key");
-
+    pub fn new(signer: NetworkSigner, rpc_url: impl Into<String>) -> Self {
```
nhtyy, line 82: Probably not a big deal but this is technically a breaking change. 

I think it would make more sense to make this an `impl Into<NetworkSigner>`

crates/sdk/src/network/signer.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// An error occurred while parsing the KMS ARN.
+    #[error("Invalid KMS ARN format: {0}")]
+    InvalidKmsArn(String),
+}
+
+/// Unified signer that supports both local private keys and AWS KMS.
+#[derive(Clone, Debug)]
+pub enum NetworkSigner {
+    /// Local private key signer.
+    Local(PrivateKeySigner),
```
nhtyy, line 69: Should this be an enum and not just a `Box<dyn alloy_signer::SignerSync>`

You can do something like 
```
pub struct NetworkSigner { inner: Box<dyn SignerSync> }

impl From<S: SignerSync> for NetworkSigner { // box the signer }

impl From<S: Into<String>> for NetworkSigner { // parse pk signer }
```

This would maintain backwards compat entirely I beleive as well give you *all* alloy signers for free

crates/sdk/src/network/signer.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Create a local signer from a private key string.
+    pub fn local(private_key: &str) -> Result<Self, NetworkSignerError> {
+        let signer = private_key
+            .parse::<PrivateKeySigner>()
+            .map_err(|e| NetworkSignerError::Parse(e.to_string()))?;
+        Ok(NetworkSigner::Local(signer))
+    }
+
+    /// Create an AWS KMS signer with automatic region detection from KMS ARN.
+    pub async fn aws_kms(key_id: &str) -> Result<Self, NetworkSignerError> {
```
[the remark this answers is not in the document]
0xernesto: Yes. Their service will need either `AWS_` creds or IAM role with correct perms.

Adding to my TO-DO list

crates/sdk/src/network/client.rs
0xernesto: fixed

PR #2410, opened 2025-08-04
perf: optimize SepticExtension field arithmetic operations

Lil-Duckling-22:
## Motivation

The previous implementation of polynomial multiplication in SepticExtension performed unnecessary clone operations and computed results involving zero coefficients. This introduced avoidable overhead, especially in arithmetic-heavy workloads.

## Solution

Optimized the Mul implementations for both owned and borrowed Polynomial<T>:

- Skips multiplication by zero coefficients, significantly improving performance in sparse polynomials.
- Uses unsafe indexing (get_unchecked_mut) in the inner loop to reduce bounds checks in a performance-critical path. This is safe because index bounds are guaranteed by construction.
- Reduced redundant allocations and cloning of coefficient values.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/stark/src/air/polynomial.rs
+++ b/crates/stark/src/air/polynomial.rs
@@ -197,10 +197,23 @@ impl<T: AbstractField> Mul for Polynomial<T> {
     type Output = Self;
 
     fn mul(self, other: Self) -> Self {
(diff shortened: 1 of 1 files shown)


leruaa:

PR #2420, opened 2025-08-07
feat: introduce `SP1_MOONGATE_SERVER_PORT` env value as port for moongate-server

fyInALT:
## Motivation

In testing sp1 and op-succinct, we found that when using CUDA proving mode, the Moongate server's default port is set to 3000, which can easily cause port conflicts with other services on the machine. Since many proof service implementations do not actively configure this port parameter.

## Solution

we could introduce an environment variable to specify the desired port. For example:

```bash
SP1_MOONGATE_SERVER_PORT=13001 SP1_PROVER=cuda RUST_LOG=debug ./target/release/fibonacci-script
```

This will instruct the Moongate server to use port 13001:

```log
2025-08-07T15:42:59.954343Z DEBUG starting new connection: http://localhost:13001/
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cuda/src/lib.rs
+++ b/crates/cuda/src/lib.rs
@@ -132,7 +132,11 @@ pub enum MoongateServer {
 
 impl Default for MoongateServer {
     fn default() -> Self {
-        Self::Local { visible_device_index: None, port: None }
+        let port = std::env::var("SP1_MOONGATE_SERVER_PORT")
+            .ok()
+            .and_then(|port_str| port_str.parse().ok());
+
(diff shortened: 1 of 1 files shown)


leruaa:
This can already be done with https://docs.rs/sp1-sdk/5.2.1/sp1_sdk/cuda/builder/struct.LocalMoongateServerCudaProverBuilder.html#method.port:

Something like:

```rust
let client = ProverClient::builder()
    .cuda()
    .local()
    .port(13001)
    .build()
```

PR #2427, opened 2025-08-14
refactor: remove redundant constraints in AIR builder

PivasDesant:
Optimize constraint system to reduce gate count without changing semantics:

- Branch/jump: remove redundant assert_bool (implied by sum constraints)
- BabyBear word: combine 3 assert_zero into single constraint  
- CPU: merge transition constraints with logical OR
- Memory: combine address assertions into polynomial constraint

~10 fewer constraints per evaluation cycle. All tests pass.


leruaa:
We can't change the circuit because that would require a major version upgrade.

PR #2435, opened 2025-08-25
baby_bear_poseidon2: preserve ultra-compressed mode across Clone

MozirDmitriy:
Introduce UltraCompressed to BabyBearPoseidon2Type and set it in ultra_compressed(). Update Clone to return the correct constructor for Default, Compressed, and UltraCompressed so that ultra-compressed configs are not downgraded to compressed upon cloning.


leruaa:

PR #2454, opened 2025-09-17
feat: enable verifying compressed gnark groth16 proof

PayneJoe:
Hey, SP1 team.

Following the PR #2447 , for the purpose of reducing on-chain cost, we opened the feature of verifying compressed gnark groth16 proof with SDK API `Groth16Verifier::verify_compressed` or a more internal API `Groth16Verifier::verify_compressed_gnark_proof`. 

In this case, we can:
1. use SP1 generate uncompressed (normal) gnark groth16 proof
2. use exposed API `Groth16Verifier::compress_groth16_proof_from_bytes` to convert the uncompressed proof into a compressed one (this PR mainly focus on this)
3. use above SDK API or internal API to verify this compressed proof in VM.

I also add a test for this, please feel free to add more comments, thank you!


crates/verifier/src/tests.rs
```diff
@@ -73,6 +73,69 @@ fn test_verify_plonk(#[case] elf: &[u8]) {
...
+    new_proof[..4].copy_from_slice(&proof[..4]);
+    new_proof[4..].copy_from_slice(&compressed_proof);
+
+    crate::Groth16Verifier::verify_compressed(
+        &new_proof,
+        &public_inputs,
+        &vkey_hash,
+        &crate::GROTH16_VK_BYTES,
+    )
+    .expect("Groth16 proof is invalid");
```
leruaa, line 128: Would it be possible to integrate this to the `test_groth16_verifier` test instead of a new one?

leruaa (commented):
ty!

Looking good, I left a few comments.

crates/verifier/src/tests.rs
PayneJoe: Updated-)

leruaa (approved):
ty!

PR #2456, opened 2025-09-18
feat(sdk): add NetworkMode

AdvaithNair:
## Motivation

In the current version of the sdk, you'll need to use the network feature flag (and remove the reserved-capacity feature flag) to use the mainnet version of our prover network. This becomes a complication for cases when you want to access both the reserved and mainnet network in the same version of the sdk, due to Cargo.toml limitations.

## Solution

The best solution was to use a dual client architecture, and introduce a new function called `.network_for(NetworkMode mode)`, where you can explicitly define the mode you are looking for. The current .network() function is backwards compatible and automatically infers your network type from your feature flags. The default is Reserved.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sdk/src/client.rs
+++ b/crates/sdk/src/client.rs
@@ -138,15 +138,7 @@ impl ProverClientBuilder {
     #[cfg(feature = "network")]
     #[must_use]
     pub fn network(&self) -> NetworkProverBuilder {
-        let network_mode = {
-            cfg_if::cfg_if! {
-                if #[cfg(feature = "reserved-capacity")] {
-                    NetworkMode::Reserved
-                } else {
-                    NetworkMode::Mainnet
-                }
-            }
-        };
+        let network_mode = NetworkMode::default();
 
         NetworkProverBuilder {
             private_key: None,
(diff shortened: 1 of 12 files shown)


0xernesto (changes requested):

crates/sdk/src/client.rs
```diff
@@ -121,22 +121,69 @@ impl ProverClientBuilder {
...
     /// let prover = ProverClient::builder().network().build();
+    ///
     /// let (pk, vk) = prover.setup(elf);
     /// let proof = prover.prove(&pk, &stdin).compressed().run().unwrap();
     /// ```
     #[cfg(feature = "network")]
     #[must_use]
     pub fn network(&self) -> NetworkProverBuilder {
-        NetworkProverBuilder { private_key: None, signer: None, rpc_url: None, tee_signers: None }
+        let network_mode = {
```
nhtyy, line 143: Should call `Default`

nhtyy (changes requested):

crates/sdk/src/client.rs
```diff
@@ -121,22 +121,69 @@ impl ProverClientBuilder {
...
+    ///
+    /// # Examples
+    /// ```no_run
+    /// use sp1_sdk::{network::NetworkMode, Prover, ProverClient, SP1Stdin};
+    ///
+    /// let elf = &[1, 2, 3];
+    /// let stdin = SP1Stdin::new();
+    ///
+    /// // Explicitly specify network mode
+    /// let prover = ProverClient::builder().network_for(NetworkMode::Mainnet).build();
```
0xernesto, line 172: i think it's sufficient to only show 1 client example so people don't get confused. think of this as an example that should run if someone copied and pasted it.
just keep the mainnet prover client in the example

0xernesto (approved):

PR #2460, opened 2025-09-21
fix: Align min_auction_period default; make status and TEE parsing robust

MozirDmitriy:
- Align min_auction_period default to 1s in NetworkProver Prover impl to match NetworkProveBuilder and its docs. Previously it passed 0, conflicting with the builder’s default and semantics.
- Replace unwraps in process_proof_status with fallible conversions for ExecutionStatus and FulfillmentStatus, returning explicit errors on unknown enum values. This prevents runtime panics if the server introduces new variants.
- Harden TEE verification: add length checks for the tee_proof prefix, parse signature and recovery id with proper error handling, and return verification errors instead of panicking. This avoids crashes on malformed or unexpected TEE responses and aligns with SDK error-handling patterns.

### DIFF
--- a/crates/sdk/src/network/prover.rs
+++ b/crates/sdk/src/network/prover.rs
@@ -327,8 +327,14 @@ impl NetworkProver {
         }
 
         // Get the execution and fulfillment statuses.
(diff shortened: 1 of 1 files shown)


leruaa:

PR #2471, opened 2025-10-09
Add MachineAir Id associated type and default id() implementation

Forostovec:
- Introduce associated type Id (default: String) to MachineAir.
- Add default id() method delegating to name() for backward compatibility.

Enables callers to use stable identifiers without parsing names.


leruaa:

PR #2476, opened 2025-10-17
perf: reduce cloning in polynomial multiplication

hawkadrian:
Optimized polynomial multiplication by removing unnecessary cloning. Changed result[i + j] = result[i + j].clone() + ... to result[i + j] += ... in both Mul implementations. 

This cuts down on redundant operations during field arithmetic, which should help with performance in crypto workloads.


leruaa:

PR #2477, opened 2025-10-19
Preserve UltraCompressed mode across Clone for BabyBearPoseidon2

radik878:
ultra_compressed() is marked as Compressed, but Clone does not preserve “ultra”
In the ultra_compressed() constructor, the config_type field is set to Compressed. The Clone implementation only distinguishes between Default and Compressed; “ultra” is not present there. As a result, cloning an ultra_compressed config silently downgraded it to compressed (log_blowup 3 → 2). This change:
- Introduces a dedicated UltraCompressed enum variant.
- Sets config_type to UltraCompressed in ultra_compressed().
- Extends Clone to return ultra_compressed() when config_type is UltraCompressed.
- Adds a unit test asserting that cloning an ultra config preserves log_blowup == 3.

This ensures consistent behavior when configs are cloned in utilities or helper flows.

### DIFF
--- a/crates/stark/src/bb31_poseidon2.rs
+++ b/crates/stark/src/bb31_poseidon2.rs
@@ -249,6 +249,7 @@ pub mod baby_bear_poseidon2 {
     enum BabyBearPoseidon2Type {
         Default,
         Compressed,
+        UltraCompressed,
     }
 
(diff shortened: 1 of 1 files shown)


leruaa:

PR #2485, opened 2025-10-28
Remove duplicated `assert_bool` call in `IsZeroOperation::eval`

Koukyosyumei:
## Motivation

The `builder.when(is_real.clone()).assert_bool(cols.result);` line was duplicated in the `IsZeroOperation::eval` function. Removing the duplicate avoids unnecessary assertions and keeps the code clean.

## Solution

Deleted the second occurrence of `builder.when(is_real.clone()).assert_bool(cols.result);` in `IsZeroOperation::eval`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


leruaa:
Unfortunately we can't merge this because it would require releasing a new version of the circuits.

PR #2488, opened 2025-10-30
More dependencies are optional and only for network feature

citizen-stig:
Follow up of https://github.com/succinctlabs/sp1/pull/2483


citizen-stig:
@leruaa looking forward for merging this!

Issue #2484, opened 2025-10-28
Clarification: future of `network` feature in sdk crate


citizen-stig:
There's a TODO comment:

https://github.com/succinctlabs/sp1/blob/db2a7b2f6d988024dc8b71ebfe656c96157a78c9/crates/sdk/Cargo.toml#L76-L81

Which has been added in this PR https://github.com/succinctlabs/sp1/pull/850

But specified alloy versions already 1.0, which was updated here https://github.com/succinctlabs/sp1/pull/2309

https://github.com/succinctlabs/sp1/blob/db2a7b2f6d988024dc8b71ebfe656c96157a78c9/crates/sdk/Cargo.toml#L54-L58

**Question is** Would network feature remain?

leruaa:
We have no plan to remove the network feature.

citizen-stig:
Thanks for clarification, I've opened https://github.com/succinctlabs/sp1/pull/2489

PR #2489, opened 2025-10-30
Remove todo comment about future of removal of network feature

citizen-stig:
As described in https://github.com/succinctlabs/sp1/issues/2484


leruaa:
ty!

PR #2492, opened 2025-11-06
perf: remove redundant clone in get_recursion_core_inputs

Forostovec:
This change removes a redundant clone in get_recursion_core_inputs where a Vec built via to_vec() was cloned again before assignment. The second clone caused an extra allocation and element copy for shard proofs with no change in behavior. Using the moved Vec preserves the original semantics, reduces memory churn, and avoids unnecessary work during recursive input preparation.


leruaa:

PR #2496, opened 2025-11-18
perf: avoid byte lookup work and allocation in generate_trace

phrwlk:
This change introduces event_to_row_no_blu to populate BitwiseCols without emitting byte lookup events and switches generate_trace to use it. Previously, generate_trace allocated a Vec per event and event_to_row performed to_le_bytes and per-byte iterations to push ByteLookupEvent items that were immediately dropped. Byte lookup dependencies are already generated in generate_dependencies via a HashMap aggregation, so this work in generate_trace was redundant. The refactor preserves functionality, reduces allocations and CPU overhead during trace building, and confines byte lookup emission to the dependency-generation phase where it is actually consumed.


leruaa:

PR #2508, opened 2025-12-04
fix: prevent possible panic from copy_from_slice in HashableKey::bytes32_raw

kien-rise:
## Motivation

```rust
let vkey_bytes = vkey_digest_bn254.as_canonical_biguint().to_bytes_be();
...
result[1..].copy_from_slice(&vkey_bytes);
```

It's not guaranteed that `vkey_bytes` has 31 bytes.
If it has 30 bytes, or less, `.copy_from_slice` will panic.

## Solution

Use the correct offset.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/prover/src/types.rs
+++ b/crates/prover/src/types.rs
@@ -67,7 +67,8 @@ pub trait HashableKey {
         let vkey_digest_bn254 = self.hash_bn254();
         let vkey_bytes = vkey_digest_bn254.as_canonical_biguint().to_bytes_be();
         let mut result = [0u8; 32];
-        result[1..].copy_from_slice(&vkey_bytes);
+        let start = result.len() - vkey_bytes.len();
+        result[start..].copy_from_slice(&vkey_bytes);
         result
     }
 


peter941221:
Downstream confirmation from `alpenlabs/zkaleido#141`.

We hit the same `HashableKey::bytes32_raw()` edge while deriving `ProgramId` from an SP1 verifying key during Alpen's `evm-ee-stf` prove path.

The reproduced panic was:
`copy_from_slice: source slice length (30) does not match destination slice length (31)`

We changed the downstream caller to stop using `bytes32_raw()` and instead pack `hash_bn254().as_canonical_biguint()` into `[u8; 32]` locally.

We also added boundary tests for:
30-byte digest right-alignment
31-byte digest right-alignment
32-byte digest preservation

That branch passed:
`cargo fmt --all --check`
`cargo test -p zkaleido-sp1-host biguint_to_program_id -- --nocapture`
`cargo test -p zkaleido-sp1-host ensure_clean_exit -- --nocapture`
`cargo check -p zkaleido-sp1-host`

So this upstream PR matches a real downstream failure, not just a theoretical panic path.

PR #2511, opened 2025-12-08
feat: Build rv64 target library in SP1 toolchain

wakabat:
For older toolchains, this might fail since it requires rv64 zkvm target. So we might want to put this on hold for now.


leruaa:
Closing as this change is already in the `dev` branch.

PR #2536, opened 2026-01-30
fix: potential verifier panic in `unzip_and_prefix_sums`

erabinov:
The function `unzip_and_prefix_sums` calls `.last().unwrap()` on an iterator that is not checked to be non-empty. This PR adds a verifier check before the call to `unzip_and_prefix_sums` to make sure that the iterator in `unzip_and_prefix_sums` is non-empty.


tamirhemo (approved):
LGTM

PR #2568, opened 2026-02-07
feat: Reintroduce build-toolchain command

wakabat:
The build-toolchain command was once removed to simplify code. However, it can be quite useful in many situations:

* Help release new SP1 toolchain
* Suit users who prefer to build SP1 from source

This commit brings the build-toolchain command back

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


wakabat:
> just wondering, is this useful for maintainers if we want to release a new toolchain

Yes this is used exactly when we release new SP1 Rust toolchains: https://github.com/succinctlabs/rust/blob/succinct-1.93.0-64bit/.github/workflows/ci.yml#L57

PR #2570, opened 2026-02-07
fix(sdk): Remove dangerous -P flag from tar extraction

huahuayu:
The use of the `-P` flag in the tar command during SDK installation allows absolute paths in the archive, which could lead to arbitrary file overwrites. This PR removes the flag to prevent this vulnerability.


leruaa:

PR #2573, opened 2026-02-09
perf: eliminate redundant perm/hash/compress creation in config constructors

MozirDmitriy:
Config constructors (BabyBearPoseidon2Outer::new, BabyBearPoseidon2Inner::new, BabyBearPoseidon2::new/compressed/ultra_compressed) were calling standalone *_fri_config() functions which internally recreated the full perm → hash → compress → val_mmcs chain from scratch — even though the caller had already built the exact same objects one line above. This doubled the work on every config init, including expensive BN254 round-constant conversion for the
outer config.

Build FriConfig inline from val_mmcs.clone() instead of delegating to separate functions, matching the canonical Plonky3 construction pattern. Standalone *_fri_config() functions are left intact as public API.

### DIFF
--- a/crates/stark/src/bb31_poseidon2.rs
+++ b/crates/stark/src/bb31_poseidon2.rs
@@ -115,8 +115,14 @@ impl BabyBearPoseidon2Inner {
         let hash = InnerHash::new(perm.clone());
         let compress = InnerCompress::new(perm.clone());
         let val_mmcs = InnerValMmcs::new(hash, compress);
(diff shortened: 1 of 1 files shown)


MozirDmitriy:
Bench results:
Inner + BabyBearPoseidon2 (sp1-stark)
image
image

Outer (sp1-recursion-core)
image
image

leruaa:
Closing as the related code don't exists anymore following Hypercube release

PR #2586, opened 2026-02-12
fix: Execute program till the end when tracing mode is off.

wakabat:
When tracing mode is off, MinimalExecutor::execute_chunk will always return None. So we cannot use the return value of execute_chunk to test if a program has terminated. This change introduces a new `run_to_end` method, and fix the code so SP1 correctly identifies termination.


wakabat:
A second thought: when tracing is off, the only case when `execute_chunk` returns, is that the program terminates, so current code is actually legit.

PR #2590, opened 2026-02-13
fix: VK_ROOT encoding and Solidity verifier validation

fakedev9999:
## Motivation

V6.0.0 verifier contract artifacts had two issues: (1) VK_ROOT was encoded as an invalid 33-byte Solidity literal, and (2) contract/function name renames in the Groth16 modifier silently fail if gnark's output format changes.

## Solution

**Fix 1 — VK_ROOT 33-byte bug**: Removed the erroneous `0x00` prefix that was prepended to an already-padded 32-byte value, creating an invalid 33-byte hex literal. Added debug_assert to guard against regressions.

**Fix 2 — Solidity modifier validation**: Added pattern validation before each `strings.Replace` in both `modifyPlonkVerifier()` and `modifyGroth16Verifier()`. If gnark's ExportSolidity output format changes, the build now panics with a clear error message instead of silently producing broken contracts.

**Fix 3 — Early validation in verifier templates**: Moved `exitCode`/`vkRoot` validation checks before the `inputs` array allocation in both `SP1VerifierGroth16.txt` and `SP1VerifierPlonk.txt`. This ensures invalid proofs revert before allocating memory (fail-fast), saving gas on revert paths.

## PR Checklist

- [x] Added validation (debug_assert + explicit panic messages)
- [x] Early revert before memory allocation in verifier templates
- [ ] Added Tests (validation is implicitly tested by existing E2E tests)
- [ ] Breaking changes (none)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -190,6 +190,35 @@ jobs:
           RUSTFLAGS: -Copt-level=3 -Coverflow-checks=y -Cdebuginfo=0 -C target-cpu=native
           RUST_BACKTRACE: 1
 
+  contract-check:
+    name: Contract Check
+    runs-on: [runs-on, runner=16cpu-linux-x64, disk=large, spot=false, "run-id=${{ github.run_id }}"]
+    env:
+      CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+    steps:
+      - name: Checkout sources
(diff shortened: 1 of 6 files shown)


yuwen01 (approved):
https://github.com/succinctlabs/sp1/pull/2591

wdyt of this

PR #2591, opened 2026-02-13
contract check ci job

yuwen01:


crates/recursion/gnark-ffi/go/sp1/build.go
```diff
@@ -62,8 +62,6 @@ func modifyGroth16Verifier(file *os.File) {
 		old, new, description string
 	}{
 		{"pragma solidity ^0.8.0;", "pragma solidity ^0.8.20;", "pragma version"},
-		{"contract Verifier {", "contract Groth16Verifier {", "contract name"},
```
yuwen01: i think this ctrl-r replace stuff should be avoided. we can conform the base contract to the new gnark version instead

PR #2698, opened 2026-04-02
feat(sp1-verifier): feature-gate heavy dependencies for WASM-compatible PLONK & GROTH16 verification

dharjeezy:
The sp1-verifier crate currently pulls in sp1-hypercube, sp1-recursion-*, slop-*, and other heavy dependencies unconditionally. These transitively depend on getrandom, which doesn't compile for wasm32-unknown-unknown.  
The PLONK verifier itself doesn't use any of them, they're only needed for Groth16, compressed proofs, and recursion VK computation.
                                                                                                                                                                                                                             
This PR moves those dependencies behind a full feature flag so the PLONK and GROTH16 verifier works out of the box in WASM targets (e.g., Substrate on-chain runtimes, browser environments).                                          
   
**Motivation**                                                                                                                                                                                                                 
                            
We're building an on-chain BEEFY consensus proof verifier for https://github.com/polytope-labs/hyperbridge that runs inside a Substrate runtime (wasm32-unknown-unknown). The verifier needs to check SP1 PLONK and GROTH16 proofs on-chain. The PLONK verification math (BN254 pairing checks) is fully no_std-compatible, but the crate fails to compile for WASM because of transitive getrandom usage from deps that PLONK doesn't need.
                                                                                                                                                                                                                             
**What changed**              

  - Feature-gated behind full: sp1-hypercube, sp1-primitives, sp1-recursion-machine, sp1-recursion-executor, slop-*, bincode, serde, strum, lazy_static, Groth16 verifier, compressed proof module, and recursion VK module  
  - Always available: PLONK verifier, constants, converter, error, utils
  - PLONK_VK_BYTES and GROTH16_VK_BYTES: Changed from lazy_static! to a plain static they just include_bytes!, no lazy init needed                                                                                                           
  - VK_ROOT_BYTES: Added as a precomputed const (available in all contexts). The dynamic VK_ROOT_BYTES_DYNAMIC (computed from recursion VKs) remains available under full                                                    
                                                                                                                                                                                                                             
**Breaking changes**                                                                                                                                                                                                           
                                                                                                                                                                                                                             
None for users who enable full. Without full, only the PLONK verifier is available, Groth16, compressed proofs, and recursion VK types are gated.                                                                         
   
If you'd prefer existing behavior to be preserved without opt-in, default = ["full"] can be added so users must explicitly opt out with default-features = false.


citizen-stig:
That would be great!

fakedev9999:
Hey @dharjeezy, thanks for the original work here. #2877 ports this fix to v6.3.1. Closing this as superseded by #2877. cc @citizen-stig.

PR #2737, opened 2026-04-22
fix(sdk): re-order fulfillment and timeout checks when polling

ross-weir:
## Motivation

`process_proof_status` returns `RequestTimedOut` for fulfilled proofs polled after the deadline

`process_proof_status` checks whether `current_time > status.deadline()` before checking the fulfillment status. This means a proof that was successfully fulfilled on the network will return `RequestTimedOut` if the caller polls it after the deadline has passed — even though the proof bytes are present in the response.

This affects any use case where proof requests are submitted and polled later (e.g. batching multiple submissions before polling results). The proof completes within seconds on the network, but if the poll happens after the deadline, the fulfilled proof is rejected.

## Solution

Moves the fulfillment/execution status checks before the deadline check, so completed proofs are returned regardless of when they're polled.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


tamirhemo (approved):
LGTM!

tamirhemo:
Check seems to fail since it's on a fork. Fix here #2740

PR #2739, opened 2026-04-22
feat: Allow for more columns in zerocheck

Schaeff:
In the context of supporting APCs, we need a lot of `air_idx`.
Currently, the maximum is `1 << 7`.
This PR uses 64 bits for the jagged info instead of 32 bits, increasing the number of possible `air_idx` to `1 << 15`.


tamirhemo (approved):
LGTM

PR #2748, opened 2026-04-23
feat: add struct reflection to access the air column names

Schaeff:
struct reflection enables retrieving the air column names programatically, which is useful for manual analysis of the airs, and in particular understanding autoprecompile compilation.


tamirhemo:
Nice! Some comments: 

1. We should make `struct-reflection` dependency on the workspace toml and then use {workspace = true} on sub-crates (like most other dependencies in the repo)
2. Currently struct reflection is a git dependency. We can't release to crates.io with a git dependency, so we would need a published version

PR #2752, opened 2026-04-26
feat(sdk): sync FulfillmentStatus and treat Reverted/Expired as terminal

fakedev9999:
## Summary

- Syncs SDK `FulfillmentStatus` with wire values the RPC can already return: `REVERTED=5`, `EXPIRED=6`.
- Stops SDK polling from panicking or spinning on those terminal states.
- Maps terminal outcomes to user-safe SDK errors:
  - `REVERTED` -> `Proof request 0x... failed during settlement`
  - `EXPIRED` -> `Proof request 0x... expired before a proof was submitted`
- Keeps HA fallback simple: retry `RequestExpired` like timeout errors; do not retry `RequestReverted` by default.

## Notes

- This intentionally does not do a full proto regen.
- This intentionally does not add Explorer/product display mapping.
- Future unknown enum drift, e.g. fs=7, still hits the existing `try_from(...).unwrap()` path.
- New public `Error` variants can break exhaustive downstream matches, but as a compile error instead of today’s runtime panic.

## Verification

- `cargo +nightly fmt --all -- --check`
- `cargo build --release -p sp1-sdk --features network`
- `cargo test --release -p sp1-sdk --features network --lib` -- 38/38 pass
- `cargo clippy --release -p sp1-sdk --features network --lib --tests -- -D warnings`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/crates/sdk/src/network/proto/mod.rs
+++ b/crates/sdk/src/network/proto/mod.rs
@@ -309,20 +309,4 @@ mod tests {
         );
     }
 
-    #[test]
-    fn fulfillment_status_numeric_values_match_across_trees() {
-        // Both proto trees describe the same wire format. If a future regen drifts
-        // one tree's discriminant out of sync, the runtime selection in `mod.rs`
-        // would produce mismatched semantics — pin the equality directly.
-        assert_eq!(
-            base_types::FulfillmentStatus::Reverted as i32,
-            auction_types::FulfillmentStatus::Reverted as i32,
-        );
-        assert_eq!(
-            base_types::FulfillmentStatus::Expired as i32,
-            auction_types::FulfillmentStatus::Expired as i32,
-        );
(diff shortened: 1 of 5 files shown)


crates/sdk/src/network/proto/mod.rs
```diff
@@ -254,3 +254,75 @@ impl From<auction_types::GetProofRequestParamsResponse> for GetProofRequestParam
         Self::Auction(response)
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Farhad-Shabani, line 259: nit: the test suite could be trimmed a bit. e.g. f`ulfillment_status_numeric_values_match_across_trees` is already implied by the two per-tree wire-value tests, and `fulfilled_returns_proof` just asserts that a `None` passed in comes back as `None`. Let's cut the redundant ones keeps the surface smaller.

Farhad-Shabani (approved):

PR #2738, opened 2026-04-22
feat: Support autoprecompiles in `machine` and `executor` behind a feature

Schaeff:
## Summary

This PR adds end-to-end support for autoprecompiles (APCs) to the core SP1 machine and executor, and wires the resulting APC chips through Hypercube and GPU tracegen/prover plumbing.

At a high level, the change lets us attach dynamically generated APC chips to the RISC-V machine, execute APC candidates during tracing, and account for them correctly in records, reports, shape checking, and trace generation. The base RISC-V AIR is preserved and reused; APC support is layered on top rather than replacing the existing path.

## Main changes

- Split the existing RISC-V AIR into:
  - a base `crates/core/machine/src/riscv/air.rs`
  - an APC-aware wrapper in `crates/core/machine/src/riscv/apc.rs`
- Add APC machine construction via `RiscvAirWithApcs::machine_with_apcs(...)`, which appends runtime-generated `ApcChip`s to the base machine and extends the relevant chip clusters.
- Add the SP1-side autoprecompile machinery in `crates/core/machine/src/autoprecompiles/*`, including:
  - SP1 instruction and bus adapters
  - conversion from SP1 AIRs to symbolic machines
  - APC candidate generation / metadata
  - APC chip evaluation and trace generation
  - a small CLI for compiling APCs
- Thread APC metadata through the executor:
  - `Program` can now carry APC ranges/costs/constraints
  - the VM tracks APC candidates while executing
  - tracing can accept, reject, or abort APC calls based on runtime conditions
  - records can extract APC sub-records and store per-APC events
  - reports and shape checking now account for APC executions and failure reasons
- Add explicit handling for APC failure modes, including:
  - optimistic constraint failure / fallback
  - conflicting APC candidates
  - state bump rejection
  - memory bump rejection
  - segmentation during an in-flight APC
- Extend Hypercube / derive support so runtime-generated chips can participate cleanly:
  - `MachineAir::name()` now returns `&str`
  - `MachineAir` gains `column_names()` and `customize_program()`
  - the `MachineAir` derive macro forwards those methods for enum wrappers
- Plumb APC support through prover / SDK / GPU crates behind the `apc` feature, plus a small GPU compatibility fix for the `MachineAir::name()` signature.

## Testing

- `cargo test -p sp1-core-executor apc -- --nocapture`

This covers:
- APC insertion and successful execution
- loops
- failed APC execution / fallback
- conflicting APC candidates
- branch/PC constraint fallback behavior
- state bump rejection
[...]

PR #2753, opened 2026-04-27
feat: pass machine explicitly everywhere applicable

Schaeff:
The machine to be proven is currently always the same instance of `Machine` created with `RiscvAir::machine()`.
As a result, many places in the code just rebuild it on the fly instead of passing it around.
In the context of APCs however, not all `Machine` instances are the same, as they get created dynamically with some autoprecompile chips.
Therefore, pass the machine around everywhere applicable, in preparation for APCs.

This PR has no APC specific code, it's just a generalisation.


crates/prover/scripts/build_recursion_vks.rs
```diff
@@ -29,7 +29,9 @@ async fn main() {
 
     let maybe_range = start.and_then(|s| end.map(|e| (s..e).collect::<Vec<usize>>()));
     let maybe_either = maybe_range.map(Either::Left);
-    let node = SP1LocalNodeBuilder::from_worker_client_builder(cpu_worker_builder())
+    // TODO: This should take the apcs into account. Now it uses vanilla riscv.
```
Schaeff: remove

crates/prover/src/worker/controller/core.rs
```diff
@@ -167,6 +171,7 @@ pub struct SP1CoreExecutor<A: ArtifactClient, W: WorkerClient> {
     gate: super::ProveShardGate<A, W>,
     minimal_executor_cache: Option<MinimalExecutorCache>,
     cycle_limit: Option<u64>,
+    _machine: Machine<SP1Field, RiscvAir<SP1Field>>,
```
Schaeff, line 178: This is not used now, because we do not customize the executor using the machine so that APCs are executed instead of the software path. The change that introduces the customization is in #2738

sp1-gpu/crates/server/src/server.rs
```diff
@@ -32,7 +32,8 @@ pub struct Server {
 /// The context for a single connection to the server.
 struct ConnectionCtx {
     pk_cache: HashMap<[u8; 32], CachedProgram>,
-    prover: Arc<SP1LocalNode>,
+    prover: tokio::sync::OnceCell<Arc<SP1LocalNode>>,
```
Schaeff, line 35: We only know the machine at setup time, therefore we can't create a prover before that.

crates/prover/src/worker/controller/mod.rs
```diff
@@ -322,6 +323,8 @@ where
             num_deferred_proofs,
             cycle_limit,
             context: context.clone(),
+            // TODO: is this expensive?
```
tamirhemo, line 326: shouldn't be since chips are references

tamirhemo (approved):

PR #2762, opened 2026-04-29
feat(zkvm): make sp1-zkvm + sp1-primitives no_std, forward main's exit code

tamirhemo:
Enable `sp1-zkvm` and `sp1-primitives` to actually be `no_std`. Currently, these crates require std for no good reason. This PR addresses this. 

Main changes:
- Flip `sp1-zkvm` and `sp1-primitives` from \`extern crate alloc\` to actual `#![no_std]`. Mechanical `std::*` → `core::*`/`alloc::*` swaps; no slop crate changes were needed because slop deps' internal \`std::*\` doesn't leak through their public API surface.
- Thread \`main\`'s \`i32\` return value through \`__start\` to \`syscall_halt\` so a guest's \`int main(void) { return code; }\` propagates correctly. The \`entrypoint!\` macro now generates \`extern \"C\" fn main() -> i32 { ZKVM_ENTRY(); 0 }\` — Rust guests still halt with 0 by default; failure paths still go through \`panic! → syscall_halt(1)\` as before.
- Replace \`Mutex<StdRng>\` + \`std::sync::Once\` in \`sys_rand\` with a \`static mut RNG: Option<StdRng>\` since the zkVM is single-threaded.

### DIFF
--- a/crates/primitives/src/consts.rs
+++ b/crates/primitives/src/consts.rs
@@ -1,3 +1,7 @@
+use alloc::{
+    string::{String, ToString},
+    vec::Vec,
+};
 use elf::abi::{PF_NONE, PF_R, PF_W, PF_X};
 
 /// The maximum size of the memory in bytes.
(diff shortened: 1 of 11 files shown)


erabinov (approved):
LGTM!

PR #2764, opened 2026-04-29
chore: bugfixes in VEIL

rdalal:
3 fixes:
* padding for mul proof updated to match correction in paper
* PCS eval now checks if padding matches expected length
* verifier rejects if there are missing mul constraints and no mul proof.


erabinov (approved):
LGTM!

PR #2769, opened 2026-05-01
chore: improve `replay_shards` workflow

erabinov:


crates/prover/src/worker/prover/recursion.rs
```diff
@@ -1052,15 +1073,34 @@ impl<C: SP1ProverComponents> ShrinkProver<C> {
...
-            })?;
+            let SP1RecursionProof { vk, proof, vk_merkle_proof } = compressed_proof;
+            let shaped_input =
+                SP1ShapedWitnessValues { vks_and_proofs: vec![(vk, proof)], is_complete: true };
+            let shrink_input = self
+                .prover_data
+                .append_merkle_proofs_to_witness(shaped_input, vec![vk_merkle_proof])?;
+
+            if !SHRINK_RECORDS_WRITTEN.load(std::sync::atomic::Ordering::Relaxed)
+                && std::env::var("SP1_RECORD_SHRINK_INPUT")
```
tamirhemo: This adds reading from `env` every time here + an atomic load. The `env` should be read once at setup time for this sruct. 

Also, it's better to condition the atomic load only if writing is necessary. Namely,
```
if let Some(path) = shrink_input_dir {

   if !SHRINK_RECORDS_WRITTEN.load(Relaxed) {
        // write record
     }
}
```

crates/prover/src/worker/config.rs
```diff
@@ -101,7 +101,7 @@ impl SP1WorkerConfig {
             .ok()
             .and_then(|s| s.parse::<usize>().ok())
             .unwrap_or(DEFAULT_NORMALIZE_PROGRAM_CACHE_SIZE);
-        let dump_shard_dir = env::var("SP1_DUMP_SHARD_DIR").ok();
+        let record_write_dir = env::var("SP1_RECORD_WRITE_DIR").ok();
```
tamirhemo: This is a good way to record this but I think a better place for this is in the worker initialization code rather than the config since it's more of a benchmarking setup then an actual config

sp1-gpu/crates/perf/README.md
```diff
@@ -79,34 +66,27 @@ The config file is a JSON array of program entries:
 ]
 ```
 
-The output directory path is printed to stdout on completion.
+### `composed_workflow` — Write-then-replay in one shot
```
tamirhemo, line 69: The composed workflow is very nice!

tamirhemo (approved):
LGTM

PR #2772, opened 2026-05-05
perf(gpu): Implement `BlockAir` for Global.

erabinov:
Currently, all constraint evaluations of the Global chip in zerocheck occur entirely sequentially. In this PR, we implement the `BlockAir` trait for the chip so that the zerocheck prover can know which constraints it can evaluate in parallel blocks.


tamirhemo:
Not needed since DAG transition

PR #2775, opened 2026-05-06
perf(gpu): increase grid size for grind kernel launches as number of bits grows

erabinov:
The grind kernel was calling only a single block with 512 threads. For shrink and wrap (22 bits of grinding), this creates an approximately 100ms latency during the grind phase. This PR scales the grid-size adaptively to the number of bits.


tamirhemo:
@erabinov should we merge this?

tamirhemo (approved):

PR #2778, opened 2026-05-06
fix(prover): close three leak paths in distributed-controller path

tamirhemo:
## Summary

Three complementary leak paths in the distributed-controller artifact handling caused artifacts to survive past their useful lifetime, falling back to the 4-hour redis TTL instead of being explicitly deleted.

1. **Leak A — per-shard `DeferredEvents` wrapper artifact never deleted.** Every shard uploads a wrapper artifact containing its `DeferredEvents` (the metadata describing precompile-event slices). The controller downloads it once in `emit_precompile_shards` and consumes the contents, but the wrapper itself is never `try_delete`d. One zombie wrapper per shard accumulates until the 4 h TTL — 17 B for shards with no precompiles, larger for heavy ones.
2. **Leak B — `add_ref` / `remove_ref` key mismatch on partial-split precompile artifacts.** When a precompile artifact straddles a shard boundary (`count > threshold`), `DeferredEvents::split` calls `add_ref` *before* truncating the partial slice, so the ref key uses the original `end_idx`. The consumer in `prover/core.rs:660-672` then calls `remove_ref` with the truncated `end_idx` from its `TraceData`. Redis `SREM` on a non-existent set member is a silent no-op — the refs set never empties, the artifact never falls through `remove_ref → try_delete`, and the artifact survives until the 4 h TTL. **Also unified the format string from `{:?}_{:?}` to `{}_{}` to match the consumer side** (functionally identical for `usize` but the divergence was a footgun).
3. **Leak C — `vk_artifact` not released after the in-process setup cache stores the vk.** The controller's setup-vkey path creates `vk_artifact`, runs the `SetupVkey` task with it as output, downloads the resulting `SP1VerifyingKey`, and stashes it in `setup_cache`. From that point on the redis copy is unused, but it was never deleted — leaving one ~100 B sidecar per unique program per worker process lifetime. The full-node single-machine path already does the equivalent at `node/full/mod.rs:97`; only the distributed-controller path was missing it.

The fix for B is reordering: truncate the partial-last slice in `artifacts` *before* the `add_ref` loop, so the keys we register match the indices the consumer will see in its `TraceData::Precompile`.

## Why it matters

A and B both scale linearly with shard count; C scales with unique programs (per worker process). They don't blow up redis on their own, but together they reduce the effective admission-gate budget for subsequent proofs, which makes concurrent-large-proof scenarios more likely to trip the 2 min `ADMISSION_MAX_WAIT` and fail mid-flight (and a mid-flight failure skips the controller's bulk `delete_batch`, leaving real proof artifacts pinned for 4 h on top of the leaks). On a low-RAM repro stack (256 MB redis), one RSP proof leaves **245** zombie `artifact_*` keys + **46** orphan `refs:artifact_*` sets — at prod scale (100-1000× larger), that's tens of thousands of keys per proof.

## Test plan

Reproduced + verified on a self-hosted cluster running this branch via the network-gateway path (`sp1-perf-prover` → cluster gateway → cpu-node + gpu0 with redis backed at 256 MB).

| Metric | Before fix | After fix (A+B) | Δ |
|---|---|---|---|
| Keys post-proof (single RSP, 525 M cycles, ~200 shards) | 293 | **3** | −99 % |
| Memory used | 78.57 MB | **3.54 MB** | −95 % |
| `refs:artifact_*` orphans | 46 | **0** | −100 % |
| Zombie wrappers (17 B / 74 B) | 161 | **0** | −100 % |
| Large precompile-chunk leaks (~950 KB) | 5 | **0** | −100 % |
| Peak in-flight memory during proof | 139 MB | 123 MB | −12 % |
| Proof time (sanity) | 150.4 s @ 4.53 Mgas/s | 150.4 s @ 4.59 Mgas/s | unchanged |
| Ref add/remove failures | 0 | 0 | — |
| Admission wait-cap-exceeded | 0 | 0 | — |

After the A+B fix the 3 leftover keys were the proof itself + a small URI sidecar + the gateway's `program_*` vk-hash ELF cache. Tracing the small sidecar revealed it was the unfreed `vk_artifact` — Leak C, fixed in the second commit.

- [x] cargo check -p sp1-prover (both base commits)
- [x] End-to-end RSP run via gateway, 256 MB redis, single proof — A+B verified
- [ ] End-to-end re-run with C fix — pending (cluster rebuild in progress)
- [ ] Sequential same-program × 3 — pending
- [ ] Different programs back-to-back — pending
- [ ] CI

## Notes for reviewers

- All three fixes are local to `crates/prover/src/worker/controller/{precompiles.rs, mod.rs}`. No public API changes.
- The truncation reorder in (B) preserves the semantics of the `if i == artifacts.len() - 1 && count > threshold { break; }` check — the controller-ref skip on partial-last is still intentional, so the artifact survives across the next `split()` call.
- The `try_delete` on the deferred-events wrapper (A) is safe because `append` adds `_controller` refs to all the precompile-chunk artifacts it points at *before* this point; deleting the wrapper does not affect chunk reachability.
- The `try_delete` on `vk_artifact` (C) is safe because `setup_cache` keeps an in-process clone of the `SP1VerifyingKey` for the lifetime of the worker; the redis copy is read once and never again.
- None of the three fixes are on the failure path. Mid-proof failures still skip the bulk `delete_batch` at `controller/mod.rs:526` (separate cluster-side issue, out of scope).


Farhad-Shabani (approved):

tamirhemo:
**Verification complete** with all three fixes (A+B+C) on a self-hosted cluster (256 MB redis, single GPU 4090).

| Scenario | Pre-fix leftovers | Post-fix leftovers |
|---|---|---|
| Single RSP | 293 keys / 78.6 MB | 3 keys / 3.5 MB |
| Same program × 3 sequential | ~870 keys | **4 keys / 5.3 MB** |
| 3 distinct RSP variants sequential | ~880 keys | **9 keys / 11.2 MB** |

Across all 6 proofs in test 1 + test 2:
- 0 `Failed to add ref` / `Failed to remove ref`
- 0 `admission wait cap exceeded`
- 0 `Redis near capacity` warnings
- 0 panics
- Peak in-flight memory: 126.8 MB (49% of 256 MB cap)
- Per-proof time identical across runs (for v6/rsp, regardless of cache state — confirms no regressions)

After the test sequence the only leftovers are exactly what the design specifies: one `program_<vk_hash>` per unique program (no TTL, ELF cache) + one ~800 KB proof artifact per call (4 h TTL) + nothing else.

PR #2781, opened 2026-05-07
feat: APC support in prover and sdk

Schaeff:
## Summary

Wire APC-backed RISC-V machines through the prover and SDK so we can execute, prove, verify, and benchmark against machines generated from autoprecompiles.

This branch makes the prover stack machine-aware end to end, updates recursion sizing for APC machines, and adds the guardrails needed where APC machines still diverge from the standard SNARK-wrap path.

## What changed

- preserve APCs when serializing/deserializing `Machine<SP1Field, RiscvAir<SP1Field>>`
  - replace the placeholder `SerializableRiscvMachine` with a real APC-aware wrapper
  - serialize APCs through JSON so Powdr algebraic expressions round-trip safely

- thread the active machine through prover and verifier construction
  - use `Program::custom(..., machine)` in node execution/setup paths
  - build core/compress verifiers from the active machine instead of assuming the default machine
  - carry the machine into recursion prover config and runtime data

- make recursion sizing APC-aware
  - introduce `machine_has_apcs`
  - use a larger recursion max row count for APC machines
  - select APC-specific compress shapes when appropriate

- align vk verification behavior with APC machines
  - automatically disable recursion VK verification for APC machines under `experimental`
  - mirror that behavior in the light node so full-node and light-node verifier state stay consistent
  - relax `vk_root` checks in experimental verification paths where APC machines use modified VKs

- add explicit SNARK-wrap gating for APC machines
  - reject shrink-wrap / Groth16 / Plonk requests when the active machine uses APCs
  - expose `supports_snark_wrap()` on verifier/node codepaths so callers fail early with a clear error

- extend SDK and perf tooling for APC flows
  - add SDK slow tests that construct APC machines from PGO/autoprecompile output
  - add GPU perf node support for `--autoprecompiles`
  - add APC feature plumbing in perf/server crates

## Why

Before this change, the prover stack could accept a machine in some APIs, but important paths still rebuilt verifier/program state from the default RISC-V machine. This PR allows APC machines everywhere.

## Notes / limitations

- APC machines are supported through execution, setup, core proving, compressed proving, and verification.
- APC machines do **not** yet support shrink-wrap, Groth16, or Plonk wrapping in the worker/node path.
- For APC machines, recursion VK verification is currently disabled under `experimental` until the modified VK path has full verification support.

## Testing

- added SDK slow tests covering:
  - APC-backed fibonacci core/compressed flows
  - APC-backed keccak core flows
- added/updated prover-side APC shape tuning coverage
- updated GPU perf node to exercise APC-backed machines via `--autoprecompiles`


crates/sdk/src/blocking/mock/mod.rs
qwang98, line 1: Note to agent: remove diff.

crates/sdk/src/mock/mod.rs
qwang98, line 1: Note to agent: remove diff.

crates/sdk/src/proof.rs
qwang98, line 1: Note to agent: remove diff.

crates/verifier/src/tests.rs
qwang98, line 1: Note to agent: remove diff.

crates/sdk/src/blocking/mock/mod.rs
qwang98: Claude answer: Done in b61187894 — blank lines removed; file now matches base.

crates/sdk/src/mock/mod.rs
qwang98: Claude answer: Done in b61187894 — blank lines removed; file now matches base.

crates/sdk/src/proof.rs
qwang98: Claude answer: Done in b61187894 — blank lines removed; file now matches base.

crates/verifier/src/tests.rs
qwang98: Claude answer: Done in b61187894 — blank lines removed; file now matches base.

crates/prover-types/src/machine.rs
```diff
@@ -1,23 +1,58 @@
...
-/// wrapper in shared transport types preserves the machine slot in protocols for follow-up work.
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
-pub struct SerializableRiscvMachine;
+/// Only the APCs are serialized; the base machine is reconstructed via
+/// `RiscvAir::machine_with_apcs()` on deserialization.
+#[derive(Debug, Clone)]
+pub struct SerializableRiscvMachine(pub Machine<SP1Field, RiscvAir<SP1Field>>);
+
+impl Serialize for SerializableRiscvMachine {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
```
qwang98, line 19: Note to agent: Where is this used? Is this only for cached APC or debug printing?

qwang98: Claude answer: It's used on two wire protocols, not for caching or debug:

1. **Distributed prover IPC** — `CoreExecuteMetadata` in `crates/prover/src/worker/controller/core.rs:61-66` embeds a `SerializableRiscvMachine` and is bincode-serialized by `MessageSender::send` (`controller/core.rs:55-58`) when the controller dispatches a core-execute task to a remote worker. The conversion happens at `controller/core.rs:109` (`machine: self.machine.into()`).
2. **CUDA/GPU server IPC** — `Request::Setup { elf, machine: SerializableRiscvMachine }` in `crates/cuda/src/api.rs:11` is sent from the SDK client (`crates/cuda/src/client.rs:45`) to the GPU server, which deserializes it at `sp1-gpu/crates/server/src/server.rs:127` to rebuild the same APC-aware machine on the GPU side.

Both paths cross a process boundary, so the APCs have to actually travel — hence the real `Serialize`/`Deserialize` impl (rather than the previous placeholder that just round-tripped to `RiscvAir::machine()`).

crates/prover/scripts/build_recursion_vks.rs
```diff
@@ -29,6 +29,7 @@ async fn main() {
 
     let maybe_range = start.and_then(|s| end.map(|e| (s..e).collect::<Vec<usize>>()));
     let maybe_either = maybe_range.map(Either::Left);
+    // TODO: This should take the apcs into account. Now it uses vanilla riscv.
     let machine = RiscvAir::machine();
```
qwang98, line 33: Note to agent: My understanding is that this is the binary used to build the 200k+ recursion vks in production. This is not currently feasible for arbitrary set of APC and should not be used in tests, as it will build for days. You must check if my understanding is correct or not.

qwang98: Claude answer: Your understanding is correct on all three counts.

- **Yes, this is the production vk-map builder.** `crates/prover/Cargo.toml:85-88` declares it as the `build_recursion_vks` bin. The script calls `SP1LocalNode::build_vks` (`worker/node/full/mod.rs:112`), which dispatches the `UtilVkeyMapController` task that ultimately runs `build_vk_map` (`shapes.rs:488`) over every shape returned by `create_all_input_shapes(machine.shape(), DEFAULT_ARITY=4)`. The output `vk_map.bin` is the same file `include_bytes!`'d at `recursion.rs:57` and `worker/node/full/mod.rs:508`.
- **Yes, "~200k vks" is the right order of magnitude.** The committed `crates/prover/src/vk_map.bin` is 7,434,488 bytes; entries are `([SP1Field; DIGEST_SIZE], usize)` ≈ 40 bytes → ~185k entries. So ~200k is correct rounding.
- **Yes, infeasible for arbitrary APC sets in tests.** `create_all_input_shapes` enumerates over the machine's chip shapes; adding APC chips multiplies that combinatorially, and each shape compiles a full recursive verifier program + setup. Running this binary per-APC-config in CI would take days, matching your assessment.

This PR's diff to the script is only the new TODO comment at line 32 (`// TODO: This should take the apcs into account. Now it uses vanilla riscv.`); the script body still uses `RiscvAir::machine()`. So tests aren't invoking it with APCs — the binary remains vanilla-only, and the TODO just flags that an APC-aware production rebuild is still needed (mirrors the analogous TODO at `recursion.rs:65`).

crates/prover/src/worker/controller/core.rs
qwang98, line 1: Note to agent: check throughout the code base whether program customization with machine is enabled for all proving paths (including but not limited to CPU/GPU, local node/worker). For instances where this is not true, comment back below.

qwang98: Claude answer: Audited all `Program::from` and `Program::custom` call sites. All actual proving paths are correctly wired to `Program::custom(elf, &machine)`. A few `Program::from` call sites remain — none are a correctness bug, but two are worth flagging.

**Proving paths — all correct ✅**

| Path | Site |
|---|---|
| Local node — setup (Light) | `crates/prover/src/worker/node/light.rs:85` |
| Local node — execute | `crates/prover/src/worker/node/core.rs:48` |
| Distributed worker — controller (remote core executor) | `crates/prover/src/worker/controller/core.rs:233` |
| Core prover — per-shard prove | `crates/prover/src/worker/prover/core.rs:288` |
| Core prover — `setup_and_prove_shard` | `crates/prover/src/worker/prover/core.rs:718` |

GPU server (`sp1-gpu/crates/server/src/server.rs:127-184`) routes `Setup`/`ProveWithMode` through `SP1LocalNode` → those same `Program::custom` sites, so the GPU IPC path is fine.

**Remaining `Program::from` sites — categorized**

- *Correct by design — PGO / pre-APC*: `crates/sdk/src/lib.rs:149` (in `test_e2e` helper), `sp1-gpu/crates/perf/src/bin/node.rs:84` (`--autoprecompiles` branch), `crates/core/machine/src/autoprecompiles/mod.rs:95,115`. These intentionally bypass APCs because they're *building* the execution profile that selects which APCs to add.
- *Not on a prove path*: tests under `crates/core/runner/src/tests.rs`, `crates/core/executor/**`, `crates/core/machine/**` (chip-level unit tests), `crates/recursion/circuit/src/shard.rs:466`, `sp1-gpu/crates/jagged_tracegen/src/test_utils.rs:758`, `crates/perf/src/bin/executor.rs:190` (executor benchmark, not prover), `crates/prover/src/worker/prover/execute.rs:367` (`#[cfg(test)]`).

**Two minor flags (not bugs, but worth deciding on)**

1. **`crates/prover/src/utils.rs:74` `get_cycles(elf, stdin)`** — disassembles via `Program::from` and runs `MinimalExecutorRunner`. Called only from `crates/eval/src/lib.rs:172` in `run_evaluation` to populate the `cycles` field of the perf report. With an APC-enabled prover the reported cycles will reflect *vanilla* execution rather than APC-folded execution, which can be confusing in benchmark output. Correctness of the proof itself is unaffected.
2. **`sp1-gpu/crates/perf/src/bin/replay_shards.rs:180,216`** — hard-codes `RiscvAir::machine()` (vanilla) and uses `Program::from`. Internally consistent (no APC machine ⇒ no need for `Program::custom`), but this binary has *no* APC support at all. If APC replay-benchmarking is desired, both the machine constructor and the program disassembly would need to be made APC-aware (mirroring what `node.rs` does).

No fixes pushed since I didn't find an incorrect-by-current-spec site. Let me know if you'd like me to address either of the two flags above (especially `get_cycles` if eval reports are expected to be APC-aware).

crates/prover/src/worker/config.rs
```diff
@@ -141,8 +141,10 @@ impl SP1WorkerConfig {
             .ok()
             .and_then(|s| s.parse::<usize>().ok())
             .unwrap_or(DEFAULT_MAX_COMPOSE_ARITY);
-        let reduce_shape =
-            SP1RecursionProofShape::retrieve_or_compute_reduce_shape(machine, max_compose_arity);
+        let reduce_shape = SP1RecursionProofShape::retrieve_or_compute_reduce_shape(
+            machine.clone(),
```
qwang98: Note to agent: is machine.clone() required here?

qwang98: Claude answer: Yes, the `machine.clone()` at L145 is required — `machine` (taken by value into `SP1WorkerConfig::new`) is consumed by `retrieve_or_compute_reduce_shape` here AND by `SP1RecursionProverConfig::new` further down (was L158). One of the two has to clone; the L145 one is the right place since the second call is the binding's last use.

The clone *at the second call site* (L158) was redundant. Dropped in 5bd26bb64 so we only clone once.

crates/prover/src/recursion.rs
```diff
@@ -35,7 +35,7 @@ use sp1_recursion_compiler::{
     ir::{Builder, DslIrProgram},
 };
 use sp1_recursion_executor::{RecursionProgram, DIGEST_SIZE};
-#[cfg(feature = "mprotect")]
+#[cfg(any(feature = "mprotect", feature = "experimental"))]
 use sp1_verifier::VerifierRecursionVks;
```
qwang98, line 39: Note to agent: Is this import required on experimental feature? Why not the reverse?

qwang98: Claude answer: The import gate matches the gate on the only consumer.

- The import at L38-39 brings in `VerifierRecursionVks` *because* `to_verifier_vks()` at L131-140 is also `#[cfg(any(feature = "mprotect", feature = "experimental"))]`. The compiler needs the type in scope iff that method compiles. Both gates have to use the same predicate, or you get either `unused import` warnings or `type not found` errors depending on feature combo.
- The method itself is invoked only from `worker/node/light.rs:69-73`:
  - On `mprotect` *or* `experimental`: build a fresh `RecursionVks` from the current machine and call `to_verifier_vks()` to materialize a real `VerifierRecursionVks` (real `root`, real `num_keys`).
  - Otherwise: `VerifierRecursionVks::default()` (placeholder values; the production verifier carries the prod root via the statically-embedded `vk_map.bin`).
- "Why not the reverse?" — under `experimental` the machine may carry APCs, which changes the recursion shape set and therefore the vk merkle root. The default placeholder would not match, so we *have* to compute and convert at runtime. The non-feature path can use `default()` because production builds rely on the pre-baked artifacts. Same rationale for `mprotect`: it changes which VKs are admitted, so we can't rely on the static default. Flipping the cfg would leave the actual-machine paths with stale placeholder values and break vk-root checks.

crates/prover/src/shapes.rs
```diff
@@ -186,15 +187,27 @@ impl SP1RecursionProofShape {
...
+            false => {
+                tracing::info!("Reduce shape computation started, because the machine has apcs. Ideally, this should be loaded from disk.");
+                // compute the shape.
+                // TODO: we should cache it the same way the vanilla one is cached
+                assert_eq!(
+                    max_arity, DEFAULT_ARITY,
+                    "DEFAULT_ARITY is used a lot in the code so we require max_arity to match it"
+                );
+                block_on(Self::compute_compress_shape(machine))
+            }
```
qwang98: Note to agent: My understanding is that compress shape here is used for sizing trace gen, and will panic if we don't dynamically compute it for each APC enabled machine. This still doesn't allow vk verification, which requires computing the entire 200k+ set of vks. You must check if my understanding is correct or not.

crates/sdk/src/lib.rs
```diff
@@ -83,15 +83,109 @@ pub use utils::setup_logger;
...
+        use rand::{distributions::Distribution, Rng, SeedableRng};
+        let mut rng = rand::rngs::StdRng::seed_from_u64(seed);
+
+        (0..count)
+            .map(|_| {
+                let actual_len = rand::distributions::Uniform::new(0_usize, len).sample(&mut rng);
+                (0..actual_len).map(|_| rng.gen::<u8>()).collect::<Vec<u8>>()
+            })
+            .collect()
+    }
```
qwang98, line 120: Note to agent: This API might be available else where in the code base. Check if we can import rather than creating one here.

crates/prover/src/shapes.rs
qwang98: Claude answer: Both parts of your understanding are correct.

**Compress shape ⇒ trace-gen sizing (would panic otherwise).** `reduce_shape` is consumed in `worker/prover/recursion.rs`:
- `reduce_shape.shape` is *stamped onto* every compose and deferred program (`recursion.rs:501` and `:529`: `program.shape = Some(reduce_shape.shape.clone())`).
- It's also passed into the deferred witness builder at `recursion.rs:523` and to `compose_input` shapes at `:491`.

That `program.shape` is what the recursion trace generator uses to determine column heights and pad rows; an APC-enabled machine has more chips in the core verifier and produces a different recursion shape, so a stale cached shape (from `compress_proof_shape_from_arity`, which only knows about the vanilla `RiscvAir::machine()`) won't fit the actual program → would panic or fail downstream constraint checks. The new `block_on(Self::compute_compress_shape(machine))` at `shapes.rs:205` is exactly that recompute. (Cost cost-asterisk per the TODO at `shapes.rs:200`: it's run at config time, not cached to disk yet.)

**Doesn't enable vk verification — that's gated separately.** `vk_verification` is a sibling field on `SP1RecursionProverConfig` (`recursion.rs:77`) and on `RecursionVks` (`recursion.rs:479`). It's flipped to `false` for APC machines under `#[cfg(feature = "experimental")]` at `worker/builder.rs:55-59`:

```rust
#[cfg(feature = "experimental")]
if crate::components::machine_has_apcs(&machine) {
    config.prover_config.recursion_prover_config =
        config.prover_config.recursion_prover_config.without_vk_verification();
}
```

The accompanying comment (`builder.rs:51-54`) states this explicitly: "APC machines use a modified vk that the production vk_root constant doesn't cover, so vk verification is skipped... TODO: Change this once vk verification is implemented for the modified vk." Enabling vk_verification would require rebuilding the full ~200k recursion vk map, which is what `crates/prover/scripts/build_recursion_vks.rs` does — and which (per the earlier note we discussed at `build_recursion_vks.rs:33`) is infeasible per-APC-config in tests/CI.

So the two are deliberately split: the compress shape gets recomputed per-machine (cheap-ish), the vk map doesn't (expensive).

sp1-gpu/crates/perf/src/bin/node.rs
```diff
@@ -73,7 +80,28 @@ async fn main() {
 
     // Get the program and input.
     let (elf, stdin) = get_program_and_input(args.program.clone(), args.param);
-    let machine = RiscvAir::machine();
+    let machine = if args.autoprecompiles > 0 {
+        let program = Arc::new(Program::from(elf.as_slice()).expect("failed to parse elf"));
+        // Use the exact same proving input for APC PGO.
+        let execution_profile = execution_profile_from_program(program, stdin.clone());
+        let config = sp1_powdr_config(args.autoprecompiles, 0);
+        let pgo_config = PgoConfig::Instruction(execution_profile);
```
qwang98: Note to agent: This should be configurable when autoprecompiles > 0, not hardcoded to instruction mode. What would be the best arg set up for this?

crates/sdk/src/lib.rs
qwang98: Claude answer: There is a similar helper at `patch-testing/sp1-test/src/lib.rs:87-98` (`random_preimages_with_bounded_len(count: u8, len: usize)`), but the differences are deliberate and importing it would actually be the wrong move here:

| Property | `sp1-test` helper | New SDK helper |
|---|---|---|
| Randomness | `rand::thread_rng()` — non-deterministic | `StdRng::seed_from_u64(seed)` — reproducible |
| Count type | `u8` (max 255) | `usize` |
| Workspace | `patch-testing/` (separate, used by patch-testing crates) | `crates/sdk` |

1. **Determinism matters here.** The SDK e2e test runs the *same* preimages through APC PGO and then through proving (see `test_e2e` at `sdk/src/lib.rs:143-145`). If preimages were non-deterministic, PGO would select APCs based on a different trace than what's proved — tests would be flaky for reasons unrelated to APC correctness. The seed is hardcoded to `1234` in `keccak256_software_stdin` precisely to keep PGO and prove inputs identical across runs.
2. **`u8` is too narrow.** The new tests use `keccak256_software_stdin(100, 10)` and `keccak256_software_stdin(200, 10)` — 200 already pushes the `u8` cap.
3. **Workspace boundary.** `patch-testing/sp1-test` is in the patch-testing workspace and isn't a current dependency of `sp1-sdk`. Pulling it in just for ~10 lines of helper would expand the dependency surface inappropriately.

If you'd prefer to unify, the cleanest path is **the reverse direction** — promote a single seeded helper into a shared test-utils crate that both `sp1-test` and the SDK tests can import. Happy to do that as a follow-up if you want; otherwise I'd leave the local copy and add a comment cross-referencing `sp1-test` so future readers know the precedent.

sp1-gpu/crates/perf/src/bin/node.rs
qwang98: Claude answer: Agreed — hardcoding `PgoConfig::Instruction` is too narrow. Looking at the enum at `crates/core/machine/src/autoprecompiles/mod.rs:130-137`, the three variants are:

- `PgoConfig::Instruction(pgo_data)` — current default in this binary.
- `PgoConfig::Cell(pgo_data, max_total_apc_columns)` — cell-cost-based selection, optional column cap.
- `PgoConfig::None` — no PGO data (presumably selects APCs purely by some other heuristic; would still need APCs > 0 to make sense).

**Suggested CLI shape** (keeps `--autoprecompiles N` as the master switch):

```rust
#[arg(long, default_value_t = 0)]
pub autoprecompiles: u64,

/// PGO selection mode. Ignored when --autoprecompiles=0.
#[arg(long, value_enum, default_value_t = PgoMode::Instruction)]
pub pgo_mode: PgoMode,

/// Maximum total APC columns. Only used when --pgo-mode=cell. Unset = no cap.
#[arg(long)]
pub max_apc_columns: Option<usize>,

#[derive(Copy, Clone, Debug, clap::ValueEnum)]
enum PgoMode { Instruction, Cell, None }
```

Dispatch:

```rust
let pgo_config = match args.pgo_mode {
    PgoMode::Instruction => PgoConfig::Instruction(execution_profile),
    PgoMode::Cell => PgoConfig::Cell(execution_profile, args.max_apc_columns),
    PgoMode::None => PgoConfig::None,
};
```

Rationale:
- `--pgo-mode` is the right abstraction: it maps 1:1 to the enum variants and avoids a flag explosion.
- `--max-apc-columns Option<usize>` mirrors `PgoConfig::Cell`'s tuple signature (the autoprecompiles pipeline already asserts `!matches!(pgo_config, PgoConfig::Cell(_, Some(_)))` is unsupported at `autoprecompiles/mod.rs:122` — that constraint would surface naturally if you pass it with the wrong mode).
- `PgoMode::None` is included for completeness but probably useless unless you want a control benchmark; can be omitted if you'd rather force users to supply real PGO data.

Want me to push this directly, or do you have a preferred clap idiom (subcommands? separate `--pgo-instruction`/`--pgo-cell` flag pairs?) before I commit?

qwang98:
Note to agent: 

Audit the entire diff in this PR where machine is added as a new parameter:
- Is this absolutely required? Can this be moved further downstream (introduced later, so we have fewer diffs).
- When passed as a new parameter, can we pass as reference rather than clone?
- Is machine indeed consumed some where in the call chain downstream?

Mark any instance where ANY of these three rules are violated.

sp1-gpu/crates/perf/src/bin/node.rs
qwang98: @Schaeff What's your thoughts on this before we proceed? iirc, cell mode can be prohibitively slow in some guests, so this might be intentional?

qwang98:
Claude answer: Audit of every site in this PR where `machine` is added/changed against your three rules (required / by-ref-not-clone / actually consumed). Two genuine violations + a few worth knowing about.

### 🔴 Violation 1 — duplicate machine storage in recursion config + data

`SP1RecursionProverConfig.machine: Machine<...>` (`crates/prover/src/worker/prover/recursion.rs:673`) and `RecursionProverData<C>.machine: Machine<...>` (`recursion.rs:945-948`) both store the same value. The data field is populated at L732 via `machine: config.machine.clone()` and is only used by:

- L691 (`let machine = self.prover_data.machine.clone();`) — intermediate-verify spawn for compose proofs
- L702 (same pattern for deferred proofs)

Both sites clone `prover_data.machine` to hand into a `spawn_blocking` closure. Both could equally clone `config.machine` (or use a reference into the surrounding `Arc<SP1RecursionProverWorker>`), since the recursion worker holds both. **Recommendation: drop `RecursionProverData<C>.machine` and source those clones from `config.machine` instead** — one fewer copy of `Machine` baked into every shared `Arc<RecursionProverData>`, fewer files touched.

### 🟡 Violation 2 — `compute_compress_shape` takes `Machine` by value but never moves it

`shapes.rs:211`:
```rust
async fn compute_compress_shape(machine: Machine<SP1Field, RiscvAir<SP1Field>>) -> SP1RecursionProofShape {
    let client = SP1LightNode::new_with_machine(machine.clone()).await;  // L216 — clone
    // ... rest: machine.shape(), &machine, &machine (all borrows) ...
}
```

The function only ever borrows `machine` (L219, L223, L228) and clones once for `SP1LightNode`. Taking `&Machine` is a strict improvement: zero behavior change, eliminates the call-site move at `shapes.rs:205` (`block_on(Self::compute_compress_shape(machine))` which then needs owned). Caller (`retrieve_or_compute_reduce_shape`) already only inspects `machine.chips()` before this branch, so it can hand a reference too.

### ✅ Audited and OK

| Site | Style | Notes |
|---|---|---|
| `CoreProver::verifier(&Machine)` (`components.rs`) | ref ✅ | clones once into `ShardVerifier` — necessary |
| `RecursionProver::verifier(&Machine)` (NEW param) | ref ✅ | only used for `recursion_max_log_row_count(machine)` — borrow only |
| `SP1ProverComponents::core_verifier(&Machine)` | ref ✅ | **improved** — was `Machine` by value |
| `SP1ProverComponents::compress_verifier(&Machine)` (NEW param) | ref ✅ | borrow only |
| `machine_has_apcs` / `recursion_max_log_row_count` / `recursion_supports_snark_wrap` helpers (new) | ref ✅ | borrow only |
| `SerializableRiscvMachine::From<Machine>` and reverse | owned | required by `From` trait + stored |
| `SP1RecursionProverConfig.machine` field (new) | owned | stored, needed at config-time before `prover_data` exists (e.g. `run_groth16` guard at L755 reads `&self.wrap_prover_init.config.machine`) — can't be removed |
| `SP1RecursionProverConfig::new(machine, …)` (new param) | owned | feeds the above field |
| `dummy_compose_input(… &Machine)` (new param at recursion.rs:1027) | ref ✅ | borrow into `C::compress_verifier(machine)` |
| `recursion_prover_and_verifier(scope, &Machine)` (sp1-gpu builder.rs:1181) | ref ✅ | borrow only |
| `SP1CoreExecutor.machine` field + `new(machine)` | owned | rename of pre-existing `_machine` placeholder; now actually used for `Program::custom(&self.machine)` at L233 |

### Bonus observation (not strictly in the audit scope)

The `compress_verifier(machine: &Machine)` trait method only uses `machine` to compute the `recursion_max_log_row_count` (i.e., `21 vs 22`). You could move the machine-vs-row-count decision out of the trait by adding a separate constant resolver and have the verifier take just `max_log_row_count: usize` — that would let `compress_verifier` drop the `&Machine` param entirely, shrinking the API surface. Same applies to `RecursionProver::verifier`. Worth considering if you want to minimize this PR's diff footprint at the trait layer.

qwang98:
Note to agent: Good findings re `machine` instances. Fix violations 1 and 2.

qwang98:
Claude answer: 

**Violation 2 — fixed in 6d37d8ba9.** Changed `retrieve_or_compute_reduce_shape` and `compute_compress_shape` to take `&Machine`, and dropped the now-redundant `.clone()` at four call sites: `worker/config.rs:144`, `shapes.rs:547`, `recursion_bench.rs:147`, `replay_shards.rs:192`. `cargo check -p sp1-prover` clean.

**Violation 1 — I need to walk this back.** On closer inspection my original analysis was sloppy: removing `RecursionProverData.machine` would actually *regress*, not improve. Here's the actual situation in `worker/prover/recursion.rs`:

| Storage | Cost |
|---|---|
| `RecursionProverData.machine` | **One** backing `Machine`, shared across all consumers via `Arc<RecursionProverData>`. Per-spawn-task clones at L324/L353 are full clones, but they're sourced from the single Arc-backed copy. |
| `SP1RecursionProverConfig.machine` | Full `Machine` clone at L609 (`ShrinkProver::new(config.clone())`), L615 (`WrapProverInit { config: config.clone(), ... }`), and L681 (the inner wrap-prover build). **Three** full clones at construction. |

Total: `prover_data.machine` is 1 backing copy; `config.machine` is 3+ copies due to `config.clone()` being heavy. My audit pointed at the wrong field.

Also: `RecursionProverWorker` (the struct that consumes `prover_data.machine` at L324/L353) **does not hold a `config` field** — it only holds `prover_data: Arc<RecursionProverData<C>>`. So my proposed "source the clones from config.machine" wouldn't compile without also adding a config field to the worker, which would cost a *new* heavy clone per worker.

**Two real options if you still want to consolidate** (just say which, I'll implement):

**(A)** Remove `SP1RecursionProverConfig.machine` and re-route the four borrow sites (`recursion.rs:555`, plus `706`/`745`/`755`/`765` which all read `&self.wrap_prover_init.config.machine`) to go through `prover_data.machine` instead. Requires giving `WrapProverInit` a handle to `prover_data` (or moving the SNARK-wrap guard checks somewhere that already has `prover_data`). Saves the 3 full-Machine clones from `config.clone()` calls. Net: ~3 `Machine` allocations saved across the recursion-prover lifecycle.

**(B)** Wrap `Machine` in `Arc` at one of the layers (probably `config.machine: Arc<Machine<...>>`). Then `config.clone()` becomes cheap. Less invasive than (A) but introduces a new type to the codebase. ~1 file touched plus the field reads.

Sorry about the original audit pointing the wrong direction — please tell me which (if either) you'd like and I'll push it.

sp1-gpu/crates/perf/src/bin/node.rs
Schaeff: This is fine as is for now.

crates/prover/src/worker/node/full/mod.rs
```diff
@@ -432,7 +437,6 @@ mod tests {
 
     #[tokio::test]
     #[serial]
-    #[ignore]
```
Schaeff, line 443: Should not be ignored in target

PR #2782, opened 2026-05-08
fix: no-gas execute public-value digest extraction (#2780)

### LINKED ISSUE #2780: execute().calculate_gas(false) returns "Failed to extract public value digest" on a clean guest run (v6.1.0)
### Component

sp1-sdk

- [x] SP1 SDK
- [x] cargo prove CLI/sp1up

6.1.0

cargo-prove sp1 (d454975 2026-04-11T01:53:23.933509082Z)

### Operating System

Linux (Arch)

### Describe the bug

# `execute().calculate_gas(false)` returns "Failed to extract public value digest" on a clean guest run (v6.1.0)

**Affects:** `sp1-sdk` 6.1.0 / `sp1-prover` 6.1.0
**Path:** Local CPU execute (both async and `blocking`)

## Summary

`ProverClient::execute(elf, stdin).calculate_gas(false).run()` returns `Err(Other("Failed to extract public value digest"))` even when the guest runs to completion and successfully calls `sp1_zkvm::io::commit_slice`. The default path (`calculate_gas(true)`) works correctly. This makes the documented "skip gas accounting for faster execute" knob unusable as a fast iteration path during local testing.

## Reproducer

A minimal guest that commits any non-empty public values:

```rust
// guest/src/main.rs
#![no_main]
sp1_zkvm::entrypoint!(main);

pub fn main() {
    sp1_zkvm::io::commit_slice(&[1u8, 2, 3, 4]);
}
```

Host:

```rust
use sp1_sdk::blocking::{Elf, Prover, ProverClient, SP1Stdin};

fn main() {
    let elf_bytes = std::fs::read("path/to/guest.elf").unwrap();
    let elf = Elf::from(elf_bytes);
    let stdin = SP1Stdin::new();

    let prover = ProverClient::builder().cpu().build();

    // Default — works.
    let (pv, _) = prover.execute(elf.clone(), stdin.clone()).run().unwrap();
    println!("default: pv bytes = {}", pv.as_slice().len());

    // calculate_gas(false) — fails with "Failed to extract public value digest".
    let result = prover.execute(elf, stdin).calculate_gas(false).run();
    println!("no-gas: {:?}", result);
}
```

Output:

```
default: pv bytes = 4
no-gas: Err(Other("Failed to extract public value digest"))
```

Reproduced locally on `sp1-sdk = "6.1.0"`, Linux aarch64 in the `rust:1.91-bookworm` container described under "Environment" below; both branches are exercised by an in-repo regression test that asserts the default path returns non-empty `public_values` and the `calculate_gas(false)` path returns an `Err` containing `"Failed to extract public value digest"`.

## Expected behavior

`calculate_gas(false)` skips gas/AIR-trace estimation but still returns the committed public values bytes. The caller can then `abi_decode` / inspect them as usual.

## Actual behavior

`Err(Other("Failed to extract public value digest"))` regardless of guest correctness, because the post-execute `final_vm_state` is never populated on the no-gas path.

## Root cause

`sp1-prover-6.1.0/src/worker/prover/execute.rs`

`GasExecutingWorker::call` (lines 92–110):

```rust
async fn call(&self, input: GasExecutingTask) -> Result<ExecutionReport, ExecutionError> {
    let GasExecutingTask { chunk, final_vm_state } = input;
    if !self.calculate_gas {
        return Ok(ExecutionReport::default());     // <-- early return at L94-96
    }
    let mut gas_estimating_vm =
        GasEstimatingVM::new(&chunk, self.program.clone(), self.nonce, self.opts.clone());
    let report = gas_estimating_vm.execute()?;

    // If the VM has completed execution, set the final state.
    if gas_estimating_vm.core.is_done() {
        let final_state = FinalVmState::new(&gas_estimating_vm.core);
        final_vm_state.set(final_state).map_err(|e| {
            ExecutionError::Other(format!("failed to set final vm state: {}", e))
        })?;
    }
    Ok(report)
}
```

Then in `execute_with_options` at lines 310–321:

```rust
let public_value_digest: [u8; 32] = final_vm_state
    .get()
    .map(|state| { /* state.public_value_digest -> [u8;32] */ })
    .ok_or(anyhow::anyhow!("Failed to extract public value digest"))?;
```

`final_vm_state` is a shared lock that the gas worker is the only populator of. When `calculate_gas=false` the worker returns immediately without touching it, so `final_vm_state.get()` is `None` and extraction always errors.

The `MinimalExecutorRunner` *does* still run on the no-gas path (line 167 / spawned at line 229) and produces the committed `public_values` (line 250). The bug is purely that the digest extraction site insists on reading from `final_vm_state` rather than from those public values.

## Suggested fix

Compute the digest from the public values when `final_vm_state` is absent — the in-tree `tests::test_execute_with_optional_gas` (line 348 of the same file) already documents this equivalence:

```rust
assert!(pv.hash() == digest.to_vec() || pv.blake3_hash() == digest.to_vec());
```

So the digest extraction at lines 310–321 can fall back to `SP1PublicValues::hash()` (or `blake3_hash()`, matching whatever hash the runtime expects) when `calculate_gas=false`:

```rust
let public_value_digest: [u8; 32] = match final_vm_state.get() {
    Some(state) => {
        let mut d = [0u8; 32];
        state.public_value_digest.iter().enumerate().for_each(|(i, w)| {
            d[i * 4..(i + 1) * 4].copy_from_slice(&w.to_le_bytes());
        });
        d
    }
    None => public_values.hash().try_into()
        .map_err(|_| anyhow::anyhow!("public values hash is not 32 bytes"))?,
};
```

This avoids running a second VM and keeps the no-gas path's "fast execute" promise intact. The alternative — populating `final_vm_state` from the `MinimalExecutorRunner` instead of from `GasEstimatingVM` — is a deeper refactor (the runner doesn't expose the `core` the way `GasEstimatingVM` does) and isn't strictly needed if the goal is just to keep `public_values` working.

## Why this matters in practice

In a project that runs `cargo test` against integration fixtures, the default gas path takes ~7 minutes per fixture (debug profile, single small guest binary). `calculate_gas(false)` was the documented path to bring this down to seconds for routine path-correctness testing where gas estimation isn't needed. With this bug, users have to either (a) live with the long debug-profile runtime, (b) compile in release to recover ~5–10x, or (c) skip the host-side execute test entirely in favor of guest-only unit tests. None of those are bad, but the documented `calculate_gas(false)` knob currently doesn't deliver its promised behavior.

## Workaround

Use `cargo test --release` (sp1-sdk itself optimized) and leave `calculate_gas` at its default (`true`). For our use case this brings 6 fixtures from ~45 min total down to ~5–10 min, which is acceptable.

## Environment

- `sp1-sdk = "6.1.0"`
- `sp1-zkvm = "6.1.0"` (guest)
- `cargo-prove sp1 (d454975 2026-04-11T01:53:23.933509082Z)` — SP1 v6.1.0 release asset
- Linux aarch64 (Docker `rust:1.91-bookworm`, succinct toolchain `succinct-1.91.1-64bit-v2`)
- Reproduces with both async (`prover.execute(...).run().await`) and blocking (`prover.execute(...).run()`) APIs.

1sgtpepper:
fix: no-gas execute public-value digest extraction (#2780)

## Summary

Fixes `execute().calculate_gas(false)` returning `Failed to extract public value digest` after a guest successfully commits public values.

The no-gas path still runs the minimal executor and produces public values, but it skips gas replay. Before this change, digest extraction always required `FinalVmState`, which is only populated by the gas-executing worker. With gas disabled, `FinalVmState` stays unset and execution fails after the guest has completed successfully.

This change records the public-value digest words emitted by guest `COMMIT` syscalls during minimal execution and uses those words when gas replay is disabled. The gas-enabled path continues to use `FinalVmState`.

## What Changed

- Track committed public-value digest words in the minimal executor.
- Share `PV_DIGEST_NUM_WORDS` from `sp1-primitives::consts` and keep the existing `sp1_hypercube::air` re-export.
- Thread the digest through portable, x86_64/JIT, native runner, and runner-binary paths.
- Use `FinalVmState` for gas-enabled execution and the minimal-execution digest for no-gas execution.
- Avoid panicking on malformed minimal-executor `COMMIT` digest indices by checking the write before updating the digest array.
- Expand `test_execute_with_optional_gas` to compare gas/no-gas behavior for both SHA and BLAKE3 public-value digest modes.

## Why

`calculate_gas(false)` is intended to skip gas accounting while still returning committed public values. On the affected path, the guest can successfully commit public values, but the host returns an error because digest extraction depends on state that the skipped gas worker would have populated.

## Validation

Local branch is based on current upstream `main`:

- base: `succinctlabs/sp1@5c7f4e50759e482359c3b74316d1712cf0726ad7`
- branch: `codex/fix-no-gas-public-value-digest`

Checks run locally on this branch:

```console
git diff --check
cargo check -p sp1-jit --target x86_64-unknown-linux-gnu
cargo check -p sp1-core-executor --target x86_64-unknown-linux-gnu
```

## Notes For Maintainers

This intentionally does not recompute the digest from `SP1PublicValues` on the host. The guest/runtime already commits digest words through `COMMIT`; preserving those words keeps SHA and BLAKE3 public-value behavior aligned with the gas-enabled path.

Closes #2780.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7120,6 +7120,7 @@ dependencies = [
  "serde",
  "sp1-core-executor",
  "sp1-jit",
+ "sp1-primitives",
  "tracing-subscriber 0.3.23",
 ]
 
--- a/crates/core/executor/src/minimal.rs
+++ b/crates/core/executor/src/minimal.rs
@@ -1,6 +1,8 @@
 #![allow(clippy::items_after_statements)]
 use std::sync::Arc;
 
+use sp1_primitives::consts::PV_DIGEST_NUM_WORDS;
+
 use crate::{Program, SupervisorMode, UserMode};
 pub use arch::*;
 pub use postprocess::chunked_memory_init_events;
@@ -145,6 +147,15 @@ impl MinimalExecutorEnum {
         match self {
             Self::Supervisor(e) => e.public_values_stream(),
             Self::User(e) => e.public_values_stream(),
+        }
+    }
+
+    /// Calls `public_value_digest` on the active `MinimalExecutor`.
+    #[must_use]
+    pub fn public_value_digest(&self) -> &[u32; PV_DIGEST_NUM_WORDS] {
+        match self {
+            Self::Supervisor(e) => e.public_value_digest(),
+            Self::User(e) => e.public_value_digest(),
         }
     }
 
--- a/crates/core/executor/src/minimal/arch/portable/mod.rs
+++ b/crates/core/executor/src/minimal/arch/portable/mod.rs
@@ -9,7 +9,7 @@ use sp1_jit::{
 };
 use sp1_primitives::consts::{
     LOG_PAGE_SIZE, PROT_EXEC, PROT_FAILURE_EXEC, PROT_FAILURE_READ, PROT_FAILURE_WRITE, PROT_READ,
-    PROT_WRITE,
+    PROT_WRITE, PV_DIGEST_NUM_WORDS,
 };
 
 use std::{
@@ -63,6 +63,7 @@ pub struct MinimalExecutor<M: ExecutionMode> {
     exit_code: u32,
     max_trace_size: Option<u64>,
     public_values_stream: Vec<u8>,
+    public_value_digest: [u32; PV_DIGEST_NUM_WORDS],
     hints: Vec<(u64, Vec<u8>)>,
     maybe_unconstrained: Option<UnconstrainedCtx>,
     debug_sender: Option<mpsc::SyncSender<Option<debug::State>>>,
@@ -254,6 +255,12 @@ impl<M: ExecutionMode> SyscallContext for MinimalExecutor<M> {
         &mut self.public_values_stream
     }
 
+    fn commit_public_value_digest_word(&mut self, word_idx: usize, word: u32) {
+        if let Some(slot) = self.public_value_digest.get_mut(word_idx) {
+            *slot = word;
+        }
+    }
+
     fn enter_unconstrained(&mut self) -> io::Result<()> {
         assert!(
             self.maybe_unconstrained.is_none(),
@@ -439,6 +446,7 @@ impl<M: ExecutionMode> MinimalExecutor<M> {
             traces: None,
             max_trace_size,
             public_values_stream: Vec::new(),
+            public_value_digest: [0; PV_DIGEST_NUM_WORDS],
             hints: Vec::new(),
             maybe_unconstrained: None,
             debug_sender: None,
@@ -578,6 +586,12 @@ impl<M: ExecutionMode> MinimalExecutor<M> {
     #[must_use]
     pub fn public_values_stream(&self) -> &Vec<u8> {
         &self.public_values_stream
+    }
+
+    /// Get the public-value digest committed by the guest.
+    #[must_use]
+    pub fn public_value_digest(&self) -> &[u32; PV_DIGEST_NUM_WORDS] {
+        &self.public_value_digest
     }
 
     /// Consume self, and return the public values stream.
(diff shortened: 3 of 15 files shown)


1sgtpepper:
Closing this as superseded by #2783, which fixed #2780 and has now been merged.

For context, this PR used the same core direction: capture the public-value digest emitted by COMMIT during minimal execution and use it when gas calculation is disabled. Thanks for taking it over and getting the fix merged.

PR #2789, opened 2026-05-13
feat: Add stdin_private field to controller/executor task inputs

bergkvist:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/prover/src/worker/controller/mod.rs
+++ b/crates/prover/src/worker/controller/mod.rs
@@ -20,6 +20,7 @@ use lru::LruCache;
 
 use slop_algebra::PrimeField32;
 
+use serde::{Deserialize, Serialize};
 use sp1_core_executor::SP1CoreOpts;
 use sp1_core_executor_runner::MinimalExecutorRunner;
 use sp1_core_machine::{executor::ExecutionOutput, io::SP1Stdin};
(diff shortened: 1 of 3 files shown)


0xernesto (approved):

crates/prover/src/worker/controller/mod.rs
```diff
@@ -566,3 +564,65 @@ async fn collect_core_proofs(
...
+    type Error = TaskError;
+
+    fn try_from(inputs: &[Artifact]) -> Result<Self, Self::Error> {
+        let elf = inputs[0].clone();
+        let stdin_artifact = inputs[1].clone();
+        let mode = {
+            let parsed =
+                inputs[2].clone().to_id().parse::<i32>().map_err(|e| TaskError::Fatal(e.into()))?;
+            ProofMode::try_from(parsed).map_err(|e| TaskError::Fatal(e.into()))?
+        };
```
Farhad-Shabani: Since this TryFrom is already fallible, should inputs[0..2] use checked access too? (e.g. inputs.get(0).ok_or(TaskError::Fatal(...))) so a short slice returns Fatal instead of panicking?

crates/prover/src/worker/controller/mod.rs
```diff
@@ -566,3 +564,65 @@ async fn collect_core_proofs(
 
     Ok(())
 }
+
+#[derive(Clone)]
+pub struct ControllerInputs {
```
Farhad-Shabani, line 569: Liked this!

Farhad-Shabani (commented):
On the delete-side (worker/node/full/mod.rs:221) should try_delete use the resolved `ArtifactType` rather than hardcoded `Stdin`?

bergkvist:
> On the delete-side (worker/node/full/mod.rs:221) should try_delete use the resolved `ArtifactType` rather than hardcoded `Stdin`?

Since this is a local only implementation, and the upload is also explicitly `ArtifactType::Stdin` rather than `ArtifactType::PrivateStdin`, keeping it hard coded here should be fine.

PR #2795, opened 2026-05-16
feat(gpu): DAG-native zerocheck prover

tamirhemo:
## Summary

This PR introduces a DAG-native, fused-kernel zerocheck prover for the
GPU and makes it the zerocheck implementation, replacing the previous
SSA-tape path.

The previous prover compiled each chip's constraints into an SSA
instruction tape, relied on hand-written per-chip block decomposition,
and re-derived + re-uploaded bytecode for every shard. The new
implementation instead:

- builds each chip's constraints as a shared DAG with explicit
  cross-constraint sharing,
- derives the partition (chunker) and execution strategy (lowering)
  automatically from a register-pressure budget,
- compiles to **machine-stable** flat bytecode that is uploaded to the
  GPU once per machine and reused across every shard and cluster,
- and interprets it in fused kernels that handle every chunk of every
  chip in a round in a single launch.

Because the bytecode depends only on each chip's AIR and not on the
cluster it lands in, per-shard work scales with the shard's chips
rather than the size of the machine.

## Changes

- Add the DAG-native constraint IR (`sp1-gpu-air::ir`): builder,
  analysis, chunker, lowerings, and bytecode.
- Add the fused-kernel zerocheck prover and its CUDA kernels.
- Remove the previous SSA-tape zerocheck: the symbolic-folder
  constraint codegen, the instruction-tape optimizer, `BlockAir`, and
  the old CUDA kernels.
- Trim now-unused dependencies and dead code from the affected crates.

## Test plan

- [x] Unit tests pass for the `air`, `zerocheck`, and `shard-prover`
  crates.
- [x] `cargo fmt` and `cargo clippy --all-targets` are clean for the
  affected crates.
- [x] End-to-end proving verified across multiple RSP workloads in
  compressed mode.


sp1-gpu/crates/air/src/ir/dag.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::{EF, F};
+
+/// A handle into `ConstraintDag::nodes`. Allocated monotonically by the builder.
+pub type NodeId = u32;
+
+/// Which slice of trace a leaf comes from.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum TraceSource {
+    PreprocessedLocal,
+    PreprocessedNext,
```
erabinov: nit: not needed in current architecture

sp1-gpu/crates/air/src/ir/dag.rs
```diff
@@ -0,0 +1,140 @@
...
+    MainLocal,
+    MainNext,
+}
+
+/// Whether a constraint's assertion is over the base field or the extension field.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ConstraintField {
+    Base,
+    Extension,
```
erabinov: nit: We don't have constraints over the extension field.

sp1-gpu/crates/air/src/ir/dag.rs
tamirhemo: I think the issue is that the builder API allows it so we might want to leave it for now. But I will look into it more

sp1-gpu/crates/sys/lib/zerocheck/sequential.cu
```diff
@@ -0,0 +1,272 @@
...
+    const K* __restrict__ trace_data,
+    const felt_t* __restrict__ public_values,
+    const ext_t* __restrict__ powers_of_alpha,
+    const ext_t* __restrict__ partial_lagrange,
+    const ext_t* __restrict__ powers_of_lambda,
+    const ext_t* __restrict__ gkr_powers,
+    uint32_t rest_point_dim,
+    ext_t* __restrict__ partials
+) {
+    // The eval-point axis lives on grid_z (matches v1's design): each thread
```
tamirhemo: nit: fix this remark to just include what is done by this kernel (refrain from alluding to the earlier version that's deleted)

sp1-gpu/crates/air/examples/bytecode.rs
```diff
@@ -0,0 +1,140 @@
+#![allow(
```
tamirhemo: why are all these allowed here?

sp1-gpu/crates/air/examples/chunker.rs
```diff
@@ -0,0 +1,143 @@
+#![allow(
```
tamirhemo: why are all these allowed here?

sp1-gpu/crates/air/examples/dump_chip_layout.rs
```diff
@@ -0,0 +1,37 @@
+#![allow(
```
tamirhemo: why are all these allowed here?

sp1-gpu/crates/sys/lib/zerocheck/sequential.cu
```diff
@@ -0,0 +1,272 @@
...
+    // tripled register pressure and forced heavy local-memory spilling
+    // (profiling showed 5.9M local loads + 3.3M stores per launch). With
+    // grid_z=3 the same total work runs across 3x more blocks, each block
+    // has 1/3 the per-thread state, and the GPU schedules them concurrently.
+    K regs[MAX_REGS];
+    ext_t thread_acc = ext_t::zero();
+
+    const uint32_t stride = blockDim.x * gridDim.x;
+    const uint32_t row_limit = 1u << rest_point_dim;
+    const int e = blockIdx.z;
```
tamirhemo: Seems like we are launching one block per point. This seems wasteful especially on smaller rounds. We should be able to do it dyanmically on multiple per block (might require fixing the blockreduce issues)

sp1-gpu/crates/air/src/ir/analysis.rs
```diff
@@ -0,0 +1,216 @@
...
+    if leaves_seen == *column_leaves {
+        ConstraintShape::LinearWeightedSum
+    } else {
+        ConstraintShape::General
+    }
+}
+
+/// Recurses into an Add-chain; each leaf of the chain must be a product
+/// of a coefficient (constant / public) and a single column leaf.
+fn walk_linear_sum(
```
erabinov, line 173: Very interesting way to detect the GKR corrections!

sp1-gpu/crates/sys/lib/zerocheck/column_tile.cu
```diff
@@ -0,0 +1,134 @@
...
+        // source: 2 = preprocessed, 4 = main (local row only).
+        size_t base = (leaf.source == 4) ? main_ptr : preprocessed_ptr;
+        K z = K::load(trace_data, base + leaf.col * height + (row << 1));
+        K o = K::load(trace_data, base + leaf.col * height + (row << 1 | 1));
+        K diff = o - z;
+
+        ext_t alpha = ext_t::load(powers_of_alpha, t.alpha_idx);
+
+        K v0 = z;
+        K v1 = z + eval_point(1) * diff;
```
erabinov: This doesn't seem to be using the `diff + diff` trick.

erabinov: You need to allocate another K value, which might make some register pressure, but it is more efficient in terms of arithmetic.

sp1-gpu/crates/sys/lib/zerocheck/sequential.cu
```diff
@@ -0,0 +1,277 @@
...
+                LeafRef leaf = cm.leaves[instr.a];
+                // source: 2 = preprocessed, 4 = main (local row only).
+                size_t base = (leaf.source == 4) ? cm.main_ptr : cm.preprocessed_ptr;
+                K z = K::load(trace_data, base + leaf.col * cm.height + (row_idx << 1));
+                if (e == 0) {
+                    regs[instr.out] = z;
+                } else {
+                    K o = K::load(trace_data,
+                                  base + leaf.col * cm.height + (row_idx << 1 | 1));
+                    regs[instr.out] = z + ep_v * (o - z);
```
erabinov: Again, could benefit from the `2x = x+x` and `4x = 2x + 2x` tricks.

sp1-gpu/crates/sys/lib/zerocheck/sequential.cu
```diff
@@ -0,0 +1,277 @@
...
+                } else {
+                    K o = K::load(trace_data,
+                                  cm.preprocessed_ptr + i * cm.height + (row_idx << 1 | 1));
+                    v = z + ep_v * (o - z);
+                }
+                ext_t bp = ext_t::load(gkr_powers, cm.gkr_main_width + i);
+                acc += bp * v;
+            }
+
+            // Geq correction — subtract `geq(row, eval_pt) * padded_row_adjustment`
```
erabinov: I think this would be way more efficient to do once per chip maybe in post-processing instead of per row.

Or at least I think there's more opportunity to make these field ops more efficient or bit-ops/conditional assignments.

sp1-gpu/crates/zerocheck/src/prover.rs
```diff
@@ -0,0 +1,1473 @@
...
+// ============================================================================
+
+pub struct ZeroCheckJaggedPoly<'b, K: Field> {
+    pub data: Cow<'b, JaggedTraceMle<K, TaskScope>>,
+    /// This shard's chips, as pointer-views into `machine_bytecode`.
+    pub compiled: Vec<CompiledChipDevice>,
+    /// The machine-wide flat bytecode the `compiled` views point into. Held
+    /// here so the device buffers outlive every round of this shard.
+    pub machine_bytecode: Arc<MachineBytecode>,
+    pub virtual_geq: Vec<VirtualGeq<Ext>>,
```
erabinov: Seems unnecessary given the way the kernel currently handles the `geq_adjustment`. I think this is a relic from before an optimization that moved the computations to device.

erabinov: Never mind, I think this is necessary, but I still thing there might be some room to squeeze in the `geq` computation on device.

PR #2796, opened 2026-05-16
fix(sp1-cuda): avoid destructor panic when cleanup runs outside a Tokio runtime

peter941221:
Fixes #2673

### Summary

`sp1-cuda` currently schedules cleanup work from `Drop` using `tokio::spawn`.

That is fine while a Tokio runtime is still active, but it panics if the CUDA proving key or client is dropped after the runtime has already been torn down. Because the panic happens in a destructor, the process then aborts during cleanup.

This patch keeps the existing behavior when a runtime is active, but falls back to a small dedicated cleanup thread with its own current-thread Tokio runtime when cleanup happens outside Tokio runtime context.

### What changed

- added `spawn_cleanup_task` in `crates/cuda/src/client.rs`
- `CudaClientInner::drop()` now uses that helper instead of calling `tokio::spawn` directly
- `SessionKey::drop()` now uses the same helper for async `Destroy`
- added unit coverage for cleanup scheduling with and without an active Tokio runtime

### Why this is the right scope

This PR is intentionally narrow.

It only addresses the destructor-time panic / abort path reported in `#2673`. It does not change proving logic, proof generation, server protocol, or lower level CUDA task execution behavior.

### Repro / verification

I verified this locally on a Linux WSL CUDA setup with an RTX 5090.

Before the patch, a minimal raw CUDA repro that:

1. creates a raw `sp1_cuda::CudaProver`
2. runs setup successfully
3. drops the Tokio runtime
4. then drops the proving key / client

would panic with:

- `there is no reactor running`
- followed by `panic in a destructor during cleanup`

After this patch, the same local repro completes cleanup without panic.

I also ran:

```bash
cargo test -p sp1-cuda spawn_cleanup_task -- --nocapture
```

### Notes

While reproducing on current `main`, I also hit some environment/bootstrap noise that is separate from the reported cleanup bug:

- slow first-run download of the prebuilt `sp1-gpu-server` asset
- missing `libcudart.so.12` in the local WSL runtime until `libcudart12` was installed

I am not including any changes for that here because they are orthogonal to the cleanup panic itself, and I wanted to keep this PR small and reviewable.

### DIFF
--- a/crates/cuda/src/client.rs
+++ b/crates/cuda/src/client.rs
@@ -7,6 +7,7 @@ use sp1_prover::worker::ProofFromNetwork;
 use sp1_prover_types::network_base_types::ProofMode;
 use std::{
     collections::HashMap,
+    future::Future,
     path::{Path, PathBuf},
     sync::{Arc, LazyLock, Weak},
     time::Duration,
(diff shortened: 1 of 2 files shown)


peter941221:
Small follow-up: this is still the same narrow destructor-time cleanup fix for sp1-cuda, with the local repro path unchanged. If you want any part split further, I can keep the follow-up scoped to that cleanup path only.

PR #2798, opened 2026-05-17
perf(gpu): seed-and-grow zerocheck chunker

tamirhemo:
Stacked on #2797 (which is stacked on #2795 `feat/gpu-zerocheck-v2`). The diff is one file — `sp1-gpu/crates/air/src/ir/chunker.rs`.

## Motivation

#2797's `chunk_efficiency` example measured that the chunker emits **9.4× more column loads** than the distinct-column floor at the default `CHUNKER_MAX_LEAFSET=64`. The chunker's objective is to minimise total column-load traffic — `Σ over chunks of |chunk.leafset|` — because a column shared by K chunks is fetched from memory K times, and column reads are the dominant memory traffic of the zerocheck kernel (see the zerocheck ablation work).

## Change

Replace first-fit-decreasing with **seed-and-grow**: open a chunk on the largest unplaced constraint, then repeatedly pull in the unplaced constraint that adds the fewest new column leaves until the budget is full, then move on. Growing each chunk to fullness before opening the next concentrates column overlap far better than placing each constraint into the best-so-far chunk.

Machine-wide reload factor on `RiscvAir` (`cargo run --release --example chunk_efficiency -p sp1-gpu-air`):

| `max_leafset` | FFD (old) | seed-and-grow (new) |
|---|---|---|
| 64 (default) | 9.40× | 8.95× |
| 128 | 4.94× | 3.63× |
| 256 | 4.42× | **1.56×** |

A/B'd against the old packer and a locality-sorted variant — seed-and-grow wins at every budget and never loses.

## Safety

- **Correctness-preserving**: chunking is only a *grouping* of constraints; the machine-stable bytecode is identical for any budget-respecting grouping. The distinct-column union is unchanged (53,338 before and after).
- **Deterministic**: stable sort + strict tie-break — required for machine-stable bytecode.
- Same leafset budget → same `MAX_REGS` template tier, no kernel-side change.
- All `sp1-gpu-zerocheck` tests pass, including `test_zerocheck_real_traces` and `test_machine_bytecode_scaling`.

## Honest scoping

At the default `max_leafset=64` the win is small (9.40 → 8.95×) — there, the **budget**, not the packer, is the wall: the wide field-arithmetic precompiles (Bls12381/Bn254/Ed) have individual constraints near or above 64 leaves, so no packer can pair them. The larger value of this PR is that it makes the leafset budget *effective* — at 128/256 a good packer recovers 2–3× more than FFD does. That sets up a budget retune (which cross-tiers `MAX_REGS` and needs a GPU wall-time sweep — out of scope here). A GPU A/B is recommended before claiming a kernel speedup from this PR alone.

### DIFF
--- a/sp1-gpu/crates/air/src/ir/chunker.rs
+++ b/sp1-gpu/crates/air/src/ir/chunker.rs
@@ -1,11 +1,22 @@
-//! Greedy first-fit-decreasing chunker.
+//! Seed-and-grow chunker.
 //!
 //! Given the per-constraint analysis, packs constraints into chunks bounded
-//! by a register-pressure
-//! budget (max leafset per chunk → downstream `max_reg` → MAX_REGS template
-//! tier). Heuristic: sort constraints by descending leafset size, then for
-//! each, pick the chunk that minimizes the number of NEW unique leaves
-//! introduced.
+//! by a register-pressure budget (max leafset per chunk → downstream
+//! `max_reg` → MAX_REGS template tier). The objective is to minimise the
+//! total column-load traffic: `Σ over chunks of |chunk.leafset|`. A column
+//! shared by K chunks is fetched from memory K times, so a poor packing
+//! amplifies column loads (the dominant memory traffic of the zerocheck
+//! kernel).
+//!
+//! Heuristic: open a chunk on the largest unplaced constraint, then
+//! repeatedly pull in the unplaced constraint that adds the *fewest new
+//! leaves* until the budget is exhausted, then move on to the next chunk.
+//! Growing one chunk to fullness before opening the next concentrates
+//! column overlap far better than placing each constraint into the
+//! best-so-far chunk (the earlier first-fit-decreasing heuristic): measured
+//! on `RiscvAir`, seed-and-grow cuts the machine-wide column-load factor
+//! from 4.4x to 1.6x at `max_leafset = 256`, and is never worse at any
+//! budget.
 //!
 //! If a single constraint already exceeds the budget, it becomes its own
 //! chunk — flagged for the scheduler to route to the escape-valve lowering.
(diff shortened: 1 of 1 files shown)


tamirhemo:
## Measured base vs this branch — the reload-proxy improvement does not translate to speed

Benched both branches (base = `feat/gpu-zerocheck-v2`). **Converting to draft — this should not merge as-is.**

### zerocheck microbenchmark (`cargo bench -p sp1-gpu-zerocheck --bench zerocheck`)

| trace | base (FFD) | seed-and-grow | Δ |
|---|---|---|---|
| `random:25,cluster=core` | 22.55 ms | 21.31 ms | **−5.5%** |
| `random:28,cluster=all-chips` | 126.5 ms | 134.6 ms | **+6.4%** |
| `random:28,cluster=all-chips`, `CHUNKER_MAX_LEAFSET=256` | 267 ms | 509 ms | **+90%** |

### e2e (`node --program local-rsp 21000000 --mode core`, 2 iters)

| | base | seed-and-grow |
|---|---|---|
| core proof time | 36.26 / 36.13 s | 35.75 / 35.76 s |

e2e is ~neutral (−1.2%, within noise — zerocheck is only ~6.6 s of the 36 s core proof).

### Why the proxy didn't translate

`reload×` (column-load count) is **not** what the zerocheck kernel is bound on — it's register/occupancy/latency-bound. The chunker change trades chunk count against per-chunk register pressure:

- **Small-constraint chips** (core RISC-V, leafset p50 ≈ 3–7): denser packing → fewer chunks, register pressure stays low → **−5.5%**, a real win.
- **Wide field precompiles** (Bls12381/Bn254/Ed, constraints near/above the 64 budget): filling chunks to the budget raises `max_reg` and cross-tiers the `MAX_REGS` template. `max_reg` max went 162 (base) → **286** (this branch). At `CHUNKER_MAX_LEAFSET=256` it is catastrophic (+90%) — chunks fill to 256 leaves → `MAX_REGS=512` template, heavy spilling.

RSP shards are a mix of both, so e2e nets to ~neutral.

### Conclusion

Not a merge. The useful output is the finding: **`reload×` is not a valid standalone optimization target** — a chunker change has to be judged on the `max_reg` distribution + GPU wall-time. A register-pressure-aware packer (pack densely only while staying in the same `MAX_REGS` tier) is the real direction.

PR #2803, opened 2026-05-19
feat: ecmul syscall/precompile

rdalal:
Adding an secp256k1 ECMUL precompile and syscall. 
* (IMPORTANT:) needed to refactor how the shape checker computed costs for multi-chip syscall designs. 
- idea: syscall associated to controller chip, shape checker has a list of all controlled chips for each controller
* updated syscall adding instructions (claude) 
* changed all the needed files for adding syscalls (including overwriting ecmul implementation in zkvm/lib for secp256k1). 
* added channels to the execution record so chips can receive messages from other chips during tracegen
* new precompile is in precompiles/weierstrass/weierstrass_mul. It's a multi-chip design. 
* for multi-chip coordination, added two new interactions

(note: tests failing due to vkey issues)


rdalal:
moved to private repo

PR #2805, opened 2026-05-21
Track upstream PgoAdapter API split

georgwiese:
powdr-labs/powdr#3752 removes `PgoAdapter::filter_blocks_and_create_apcs_with_pgo` and exposes `detect_blocks` + `generate_apcs` + `select_apcs` instead, so the build cache survives changes to `autoprecompiles` / `skip`. Inline the three steps here.

Behavior preserved by setting `apc_candidates = autoprecompiles + skip` when unset — same final selection as the old fused call.

Requires bumping the powdr pin in Cargo.toml once #3752 lands.


georgwiese:
Superseded by #2830

PR #2831, opened 2026-06-02
feat(sdk): default PGU pricing to buffered market-based rates

Farhad-Shabani:


crates/sdk/src/network/prover.rs
```diff
@@ -886,10 +886,43 @@ impl NetworkProver {
...
                         } else {
-                            auction_params
-                                .max_price_per_pgu
-                                .parse::<u64>()
-                                .expect("invalid max_price_per_pgu")
+                            // Default to the network's current market wei/PGU plus the
+                            // configured buffer so the effective USD ceiling tracks
+                            // PROVE moves. If the RPC is unavailable, use the
+                            // server-supplied `auction_params.max_price_per_pgu`.
+                            match self.client.get_market_price_per_pgu().await {
```
0xernesto: Can we add tests for the new default `max_price_per_pgu` selection? I think the cases we care about are explicit max still wins, fresh `GetMarketPricePerPgu` applies the 120% buffer, `Unimplemented`/RPC failure falls back to `auction_params.max_price_per_pgu`, and malformed or overflowing market prices do not panic.

0xernesto (approved):

crates/sdk/src/network/prover.rs
Farhad-Shabani: [27e7bda](https://github.com/succinctlabs/sp1/pull/2831/commits/27e7bdac8d37b1ed33f59cccd7fd4ebb282c6460), [89cde90](https://github.com/succinctlabs/sp1/pull/2831/commits/89cde90828dc3e2670ae8fd4483e481d98a08e5b)

Farhad-Shabani: I ended up dropping the market RPC from the auto-default path entirely: `GetProofRequestParams.max_price_per_pgu` is already the safety-vetted version (server applies its own staleness gate + ceiling clamp before publishing it), so the SDK just buffers that value directly. Most of the branches you wanted tests for collapse into "use what the server vouches for".

Issue #2817, opened 2026-05-25
MockProver accepts the same mock proof for different expected exit statuses


1sgtpepper:
### Component

`sp1-sdk`

### Have you ensured that all of these are up to date?

- SP1 SDK: checked against `main` at `150e6294959f40dbc3ba42eb21c8eccc14c95bc5`
- cargo prove CLI/sp1up: not used by this reproduction

### What version of SP1 SDK are you on?

`main` at `150e6294959f40dbc3ba42eb21c8eccc14c95bc5`

### What version of the cargo prove CLI are you on?

Not relevant; this uses only the SDK mock prover path.

### Operating System

Linux (Ubuntu). Reproduced in GitHub Codespaces on Ubuntu 24.04.4 LTS:

```text
rustc 1.95.0 (59807616e 2026-04-14)
cargo 1.95.0 (f2d3ce0bd 2026-03-21)
```

### Describe the bug

`MockProver::verify` accepts `status_code: Option<StatusCode>` through the shared `Prover` trait, but does not enforce it.

The default `Prover::verify` path treats `None` as `StatusCode::SUCCESS` and checks the verifier-facing exit code for real proof modes:

- contract/default: `crates/sdk/src/prover.rs:69-78`, `crates/sdk/src/prover.rs:143-150`
- real proof checks: `crates/sdk/src/prover.rs:166-174`, `197-205`, `224-232`, `241-249`

The mock path weakens that contract:

- async mock verifier ignores `_status_code`: `crates/sdk/src/mock/mod.rs:88-108`
- blocking mock verifier does the same: `crates/sdk/src/blocking/mock/mod.rs:78-92`, `112-120`
- mock Plonk/Groth16 proof construction hard-codes `public_inputs[2]` to `"0"`: `crates/sdk/src/proof.rs:241-285`
- mock Plonk/Groth16 verification checks vkey and public-values hashes, but not the requested status: `crates/sdk/src/proof.rs:287-336`

So the same mock proof can verify as both `SUCCESS` and `PANIC`, even though the real verifier API distinguishes those expected statuses.

#### Minimal reproduction

From a clean SP1 checkout at `150e6294959f40dbc3ba42eb21c8eccc14c95bc5`, replace `/path/to/sp1` below and run:

This uses `SP1ProofWithPublicValues::create_mock_proof` directly to avoid guest toolchain setup; `MockProveRequest::into_future` creates mock proofs through the same constructor after execution.

```sh
[...]

PR #2839, opened 2026-06-07
fix(sdk): enforce expected exit status in MockProver::verify

MavenRain:
## Motivation

`MockProver::verify` takes a `status_code: Option<StatusCode>` argument (part of the shared `Prover::verify` contract) but ignored it, so the mock verifier did not enforce the program's expected exit status the way the real verifier (`verify_proof`) does.

As reported in #2817, this means a single mock proof verifies under *any* expected status, e.g. a successful program's proof verifies as both `SUCCESS` and `PANIC`.  Separately, mock Plonk/Groth16 proofs hardcoded the encoded exit code (`public_inputs[2]`) to `"0"` regardless of the real execution result, so even an enforced check would not have reflected a panicking program.

## Solution

- `create_mock_proof` now records the real execution exit code into the mock Plonk/Groth16 `public_inputs[2]` (previously hardcoded `"0"`).  The exit code comes from the `ExecutionReport` that mock proving already produces but was discarding.
- Both the async (`mock`) and blocking (`blocking::mock`) `MockProver::verify` now enforce `status_code.unwrap_or(StatusCode::SUCCESS)` against the encoded exit code via a shared `verify_mock_exit_code` helper that mirrors `verify_proof`'s Plonk/Groth16 arms exactly (same `BigUint` parse + `StatusCode::is_accepted_code` check + `UnexpectedExitCode` error).
- Added regression tests (async + blocking, Plonk + Groth16) asserting that a successful program's mock proof verifies under `None`/`SUCCESS` but is rejected under `PANIC`, so one proof can no longer satisfy two different expected statuses.

### Scope

This PR covers the Plonk and Groth16 mock modes, which carry the exit code in `public_inputs[2]` (the slot the real verifier reads).  Mock `Core`/`Compressed` proofs are intentionally hollow (`Core(vec![])` / a dummy recursion proof with empty public values), so enforcing status for those modes requires giving them an exit-code-bearing field and is left as a focused follow-up.  **Towards #2817.**

### Note

`create_mock_proof` gains an `exit_code` parameter (the execution exit code to encode).  This is a small breaking change to that public helper; the two in-tree callers (async + blocking mock provers) and the doc example are updated.

### Verification

Locally, on `main`:
- `cargo +nightly fmt -p sp1-sdk -- --check` — clean
- `cargo check -p sp1-sdk --features blocking --tests` — clean
- `cargo test -p sp1-sdk --features blocking` — 6 mock tests pass (4 new + the existing `test_mock_proof_all_types` for both async and blocking)
- `cargo clippy -p sp1-sdk --features blocking` — clean

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/sdk/src/blocking/mock/mod.rs
+++ b/crates/sdk/src/blocking/mock/mod.rs
@@ -19,7 +19,7 @@ use crate::{
         cpu::CPUProverError,
         prover::{BaseProveRequest, ProveRequest, Prover},
     },
-    proof::verify_mock_public_inputs,
+    proof::{verify_mock_exit_code, verify_mock_public_inputs},
     SP1Proof, SP1ProofWithPublicValues, SP1ProvingKey, SP1VerificationError, StatusCode,
 };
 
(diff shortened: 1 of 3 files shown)


MavenRain:
Closing as a duplicate of #2820, which predates this (opened 2026-05-27) and is the more complete fix: it enforces the expected exit status across all four proof modes (including Core/Compressed), and does so without a breaking change to `create_mock_proof`. I missed #2820 before opening this, and apologize for the overlap.  I'm happy to help review #2820 if that'd be a help.

PR #2847, opened 2026-06-16
feat: add `without_recursion` builder option for core-only workloads

qwang98:
## Summary

Adds `without_recursion()` to `SP1WorkerBuilder` and `SP1RecursionProverConfig`, mirroring the existing `without_vk_verification` pattern. When set, `SP1RecursionProver::new` skips the eager compose & deferred PK build entirely, saving ~14 GB of GPU VRAM at startup on APC-configured machines.

Default behaviour is unchanged: callers that don't call `.without_recursion()` get the same eager initialization as before. Compressed-mode users keep the same setup timing model. This is **purely additive** — opt-in by callers that know they only need core proofs.

`sp1-gpu-perf node` opts in based on `--mode`:
```rust
let worker_builder = if mode == ProofMode::Core {
    worker_builder.without_recursion()
} else {
    worker_builder
};
```

## Motivation

On RTX 5090 (32 GB), `node --program local-rsp --autoprecompiles N --mode core` OOMs for any `N >= 1` because the eager compose PK build allocates ~14 GB up front using the worst-case `compress_shape_apc.json`, leaving insufficient room for per-shard proving. Core mode never uses those PKs — the allocation is pure waste.

Probes confirm the allocation site:
```
A_before_worker_builder:      used=  539 MB     (CUDA baseline)
B_after_worker_builder:       used=  587 MB     (cuda_worker_builder: +48 MB)
C_after_node_builder_build:   used=15,375 MB    (SP1LocalNodeBuilder::build: +14,788 MB) ⚠️
E_after_setup:                used=18,193 MB
shard=0 0_start:              used=18,193 MB
```

## Why mirror `without_vk_verification`?

- Same `#[cfg(feature = "experimental")]` gating
- Same plumbing pattern (`SP1WorkerBuilder` → `SP1RecursionProverConfig` flag)
- No new env var, no new state-altering toggle in main code paths
- Opt-in safety: callers must explicitly request it

The `apc` feature already implies `experimental`, so the method is callable in all builds that use APCs (which is the only case this matters for).

## Empirical verification

RTX 5090 (32 GB), block 21740136, `--mode core`:

| APC | Before (this branch's eager init) | After (`.without_recursion()`) | Shards | Core time |
|---|---|---|---:|---:|
| 1   | ❌ OOM exit 134 | ✅ pass | 25 | 18.43 s |
| 88  | ❌ OOM exit 134 | ✅ pass | 22 | 118.78 s |

Compressed-mode callers (don't call `.without_recursion()`) retain identical behaviour and timing.

## Diff

```
 crates/prover/src/worker/builder.rs          |  32 ++++++
 crates/prover/src/worker/prover/recursion.rs | 158 ++++++++++++++++++---------
 sp1-gpu/crates/perf/src/bin/node.rs          |   9 ++
 3 files changed, 145 insertions(+), 54 deletions(-)
```

## Test plan

- [x] APC=1 RSP block 21740136 core mode passes
- [x] APC=88 RSP block 21740136 core mode passes
- [ ] Compressed mode at APC>0 still works (eager init runs at construction when `.without_recursion()` not called) — should be unchanged, needs reviewer confirmation

### DIFF
--- a/crates/prover/src/worker/prover/recursion.rs
+++ b/crates/prover/src/worker/prover/recursion.rs
@@ -84,6 +84,12 @@ pub struct SP1RecursionProverConfig {
     machine: Machine<SP1Field, RiscvAir<SP1Field>>,
     /// The reduce shape
     pub reduce_shape: SP1RecursionProofShape,
+    /// When true, `SP1RecursionProver::new` skips the eager compose & deferred PK build
+    /// entirely. The resulting prover can only serve `--mode core` requests; any
+    /// compress/shrink/wrap lookup will fail at the use-site. Should be `false` by
+    /// default; opt in via [`Self::without_recursion`] when the caller knows it will
+    /// only run core mode (e.g. RSP perf benchmarks).
+    skip_recursion_pk_init: bool,
 }
 
 impl SP1RecursionProverConfig {
@@ -113,12 +119,22 @@ impl SP1RecursionProverConfig {
             vk_map_file: None,
             machine,
             reduce_shape,
+            skip_recursion_pk_init: false,
         }
     }
     #[cfg(feature = "experimental")]
     /// Turn off vk verification for recursion proofs.
     pub fn without_vk_verification(self) -> Self {
         Self { vk_verification: false, ..self }
+    }
+
+    #[cfg(feature = "experimental")]
+    /// Skip the eager compose & deferred PK build inside `SP1RecursionProver::new`.
+    /// Use only when the caller knows it will only run `--mode core` — the resulting
+    /// prover cannot serve compress/shrink/wrap requests. Saves ~14 GB of GPU VRAM
+    /// at startup on machines configured with autoprecompiles.
+    pub fn without_recursion(self) -> Self {
+        Self { skip_recursion_pk_init: true, ..self }
     }
 
     #[cfg(feature = "experimental")]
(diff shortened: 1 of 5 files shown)


crates/prover/src/worker/prover/recursion.rs
```diff
@@ -485,7 +485,25 @@ impl<A: ArtifactClient, C: SP1ProverComponents> SP1RecursionProver<A, C> {
...
+            if skip_recursion_pk_init {
+                tracing::warn!(
+                    "SP1_SKIP_RECURSION_PK_INIT set — skipping recursion PK initialization. \
+                     Only --mode core will work; compressed/shrink/wrap modes will fail."
+                );
+            }
+
+            let effective_max_arity =
+                if skip_recursion_pk_init { 0 } else { config.max_compose_arity };
+            for arity in 1..=effective_max_arity {
```
qwang98: Can this be hooked up with `core` proof mode only, and don't trigger it for compress.

Having an extra env var is not a good idea.

qwang98: Claude answer: addressed in f452b5d. Replaced the env var with `std::sync::OnceLock<EagerRecursionPks<C>>` — compose & deferred PKs are now built lazily on first access. `--mode core` never hits the lookup, so the ~14 GB build is skipped automatically. `--mode compressed/shrink/wrap` trigger the build on first key lookup, identical end-state to before. No new env var or builder opt-in; mode-driven exactly as you asked.

Verified: RSP APC=1 and APC=88 in core mode both pass cleanly (25 / 22 shards, 0 panics).

qwang98: Claude answer: revised in 096245b. Replaced the env var with a builder method following the existing `without_vk_verification` pattern — same `experimental` feature gate, same plumbing from `SP1WorkerBuilder` → `SP1RecursionProverConfig`. `sp1-gpu-perf node` opts in via `if mode == ProofMode::Core { .without_recursion() }`, so the skip is mode-driven without an env var. Compressed-mode callers keep the eager init at construction (same timing model as before). Verified: APC=1 / APC=88 RSP core mode both pass cleanly.

tamirhemo (approved):

qwang98:
The latest target branch passes without this feature, thanks to the recursion shape change which is adapted to APC=12 from APC=88 (so we hold less buffer at proving start, and RSP is able to run).

This feature is only needed when we want to run core mode only with greater number of APCs than the recursion mode allows.

In the long run, to allow more APCs run e2e, we need updates from sp1 backend, rather than relying on this feature.

PR #2848, opened 2026-06-18
fix(prover): make shard and reduce tasks idempotent under re-delivery

Farhad-Shabani:
## Problem

The cluster delivers tasks **at-least-once** — the same `task_id` can run again via dead-worker requeue (30s heartbeat timeout) or fair-share preemption requeue. But a prover task deletes its input artifact on success ("no longer needed"). So a re-run downloads an input a prior run already deleted → `NotFound` → `TaskError::Fatal` → `fail_proof` → the whole request goes **Unfulfillable**, even though a valid proof already exists.

Observed on mainnet: a ProveShard uploaded its proof; fair-share preemption requeued the same task a few ms later; the re-run on another worker 404'd on the deleted input and failed the proof. The duplicate-delivery path produced the same failure repeatedly.

## Fix

Two parts, so re-execution is harmless and prompt reclamation is kept:

1. **Idempotent consumers.** On an input-download failure, if the input is **gone** and the task's **output already exists**, a prior run already did the work — report success instead of failing. Applied to ProveShard (`prover/core.rs`) and recursion reduce (`node/full/init.rs`, with `ReduceTaskRequest::already_reduced` / `RangeProofs::any_proof_missing`).

2. **Delete inputs last.** A consumer deletes its input only after *every* output is durable. ProveShard produces two outputs (`output` + `deferred_output`); the input delete was moved after the deferred upload completes, so "input gone" means the task is fully done — not partway. (This also fixes a latent bug: a deferred-upload failure previously returned an error *after* the input was already deleted, so the retry 404'd.)

The eager delete stays (prompt reclamation — important on memory-bounded stores); (1) and (2) make it safe.


tamirhemo:
I see the need for the fix but the implementation seems a bit messy. Maybe there is a cleaner way to do this with some sort of method on the raw proof requests that checks that outputs exist?

crates/prover/src/worker/prover/core.rs
```diff
@@ -675,6 +700,13 @@ where
             deferred_upload_handle.await.map_err(|e| TaskError::Fatal(e.into()))??;
         }
 
+        // Reclaim the input record only after every output (`output` and the
```
tamirhemo, line 677: This change makes sense

tamirhemo (approved):

PR #2864, opened 2026-07-02
perf(powdr): record-in-chip APC witness generation

qwang98:
## Record-in-chip APC witness generation

Stacked on **#2781** (APC support in prover and sdk). Replaces the "extract ApcEvents from the full software trace" path with **record-in-chip**: during tracing the executor *skips* per-opcode event emission inside APC ranges and captures a minimal `ApcInvocation` per block; the APC chip then regenerates that block's witness by re-executing it, caching the generated trace in `generate_dependencies` for reuse in `generate_trace`.

### Why
The speedup is entirely CPU-side (so the GPU proving path benefits too, since trace generation runs there as well):
- **No redundant `ApcEvents` extraction** from the software trace.
- **Trace cached in `generate_dependencies`** and reused in `generate_trace` (no double generation).
- **APC-chip `generate_dependencies` parallelized** (APC chips are commutative, independent byte-lookup producers).
- Leaner capture: zero-copy `reads: Arc<[MemValue]>` shared across a shard's invocations; `ExecutionRecordSnapshot` trimmed to the 3 fields actually read.

### Correctness
A single `current_skip` tracks the in-progress invocation, resolved on range-exit / fresh-start / success / shard-end, so loops and aborts stay correct — an aborted block's skipped prefix is replayed as software (rollback), a successful block is regenerated by the chip (B1).

### Verification (RSP, real `rsp-client` program, block 21740136, APC=12, cuda)
Measured against this same #2781 head with the ApcEvents path (APC=12), 2 clean unloaded runs each:

| metric | #2781 baseline (ApcEvents) | this PR (record-in-chip) |
|---|---|---|
| **core prove** | 49.2 / 53.0 s | **41.5 / 43.1 s** (~**−17%**) |
| peak host RAM | 23.98 / 23.28 GB | 24.16 / 22.08 GB (comparable) |
| result | `rc=0` | `rc=0`, no `CumulativeSumMismatch` |

- The speedup is in **`prove`**, where shard records + traces are generated — that's what record-in-chip changes (record-gen + trace-gen; the GPU proving portion is unchanged). **`execute` time is not affected**: `client.execute` only produces the cycle/gas `ExecutionReport` via `MinimalExecutorRunner`/`GasEstimatingVM` — no shard records, no APC witnesses — so it's identical between the two paths (any delta there is run-to-run variance).
- **Abort rate ~6.6%** of APC invocations (aborted blocks fall back to software; in line with the ~10% seen at APC=88).
- `proof time` is wall-clock (`tokio::Instant`) around `client.prove_with_mode`; single-sample, so expect some run-to-run GPU variance.
- `cargo fmt` + `cargo clippy -D warnings` clean on the touched crates.

### Commits (5, dependency order)
1. `feat(apc)`: record-in-chip data model — `ReplayTrace`, `ApcInvocation`
2. `feat(executor)`: capture record-in-chip APC invocations during tracing
3. `feat(machine)`: regenerate APC chip trace from record-in-chip invocations
4. `perf(hypercube)`: parallelize APC-chip `generate_dependencies`
5. `test(apc)`: record-in-chip prove tests (APC unit + RSP core e2e)


crates/core/executor/src/autoprecompiles/mod.rs
```diff
@@ -1,53 +1,162 @@
...
+    /// Which APC this invocation runs.
+    pub apc_id: usize,
+    /// Register file at block entry: value AND previous-access timestamp.
+    pub pre_registers: [MemoryRecord; 32],
+    /// Program counter at block entry.
+    pub pc_start: u64,
+    /// Clock at block entry.
+    pub clk_start: u64,
+    /// Global clock at block entry.
+    pub global_clk_start: u64,
```
qwang98, line 28: How are these two different?

crates/core/executor/src/autoprecompiles/mod.rs
```diff
@@ -1,53 +1,162 @@
...
+    apc_id: usize,
+    pre_registers: [MemoryRecord; 32],
+    pc_start: u64,
+    clk_start: u64,
+    global_clk_start: u64,
+    mem_reads: Vec<MemValue>,
+    num_instructions: usize,
+}
+
+impl Serialize for ApcInvocation {
```
qwang98, line 61: Do we have to build `Serialize` for `ApcInvocation`?

crates/core/executor/src/autoprecompiles/mod.rs
```diff
@@ -1,53 +1,162 @@
...
+    /// Whether there are no invocations at all.
+    pub fn is_empty(&self) -> bool {
+        self.by_id.values().all(Vec::is_empty)
+    }
+}
+
+// `ApcInvocation` holds an `Arc<[MemValue]>` (shared shard read-oracle) and `MemValue` doesn't
+// implement `DeepSizeOf`, so derive isn't possible; approximate the owned heap of the stored
+// invocations (the shared oracle is counted once by the shard, not per invocation).
+impl DeepSizeOf for ApcInvocations {
```
qwang98, line 144: Is `DeepSizeOf` required for `ApcInvocations`?

crates/core/executor/src/autoprecompiles/mod.rs
```diff
@@ -1,53 +1,162 @@
...
 #[derive(Debug)]
-/// A snapshot of current state of the execution record
-/// The execution record is made of vectors for each type of events, as well as single values
-/// The snapshot stores the size of the vectors and the current values of the single values at the time of snapshotting
-/// Not all data present in the record is present here, because some fields are guaranteed not to change during the execution of an apc.
+/// A snapshot of the parts of the execution record that change while an APC block runs, taken at
+/// the block's entry and exit. `apply_calls` diffs the two event counters across a call, and
+/// `capture_invocations` slices the block's read range from the read-oracle cursor. Everything else
+/// in the record is either unchanged during an APC or regenerated by the APC chip's re-execution,
+/// so it need not be snapshotted.
```
qwang98, line 155: These can be removed once PR is finalized.

crates/core/executor/src/record.rs
```diff
@@ -113,8 +112,11 @@ pub struct ExecutionRecord {
     pub cpu_local_page_prot_access: Vec<PageProtLocalEvent>,
     /// A trace of all the syscall events.
     pub syscall_events: Vec<(SyscallEvent, RTypeRecord)>,
-    /// A trace for all APC events.
-    pub apc_events: ApcEvents,
+    /// The record-in-chip APC invocations, keyed by `apc_id`. Replaces the materialized
+    /// `ApcEvents`: instead of extracting each block's per-opcode events during tracing, the
+    /// executor stores a minimal per-invocation capture here and the APC chip regenerates its
+    /// trace by re-executing these (see `ApcChip::b1_synthetic_record`).
```
qwang98, line 118: Be more concise and don't mention historical design when PR finalized.

crates/core/executor/src/record.rs
```diff
@@ -153,6 +155,14 @@ pub struct ExecutionRecord {
...
     /// Use optimized `generate_dependencies` for global chip.
     pub global_dependencies_opt: bool,
+    /// Read-oracle cursor for the record-in-chip capture: the number of memory reads consumed so
+    /// far by the `CoreVM`. Snapshotted at each APC block's from/to boundaries so the block's read
+    /// slice can be sliced out of the shard read-oracle for re-execution (see `TracingVM`).
+    pub mem_reads_remaining: usize,
+    /// Per-cycle gate that suppresses per-opcode event emission while an APC block's invocation is
+    /// in progress (the "skip" half of record-in-chip). Set by `TracingVM::execute_instruction`;
+    /// checked by every `emit_*`. Local-memory / page-prot / trap plumbing is unaffected.
+    pub skip_writes: bool,
```
qwang98, line 165: Rename to `skip_emit`.

crates/core/executor/src/record.rs
```diff
@@ -535,31 +545,43 @@ impl ExecutionRecord {
...
+    #[must_use]
+    pub fn has_apc_events(&self, apc_id: usize) -> bool {
+        self.apc_invocations.has(apc_id)
     }
 
     /// Get all the local memory events.
+    ///
+    /// Note: with record-in-chip, an APC block's local-memory accesses stay in
+    /// `cpu_local_memory_access` (the skip gate keeps `insert_record` running), so there is no
+    /// separate APC source to chain here.
```
qwang98, line 573: Can remove this once PR finalized.

crates/core/executor/src/tracing.rs
```diff
@@ -49,10 +50,85 @@ pub struct TracingVM<'a, M: ExecutionMode> {
     pub decoded_instruction_events: HashMap<u32, InstructionDecodeEvent>,
     /// The execution record were populating.
     pub record: &'a mut ExecutionRecord,
+    /// Whether the record-in-chip skip gate is active for this shard (any APCs in the program).
+    /// When true, per-opcode events inside an in-progress APC block are suppressed and the block
+    /// is captured as an [`ApcInvocation`] instead (regenerated by the APC chip via re-execution).
+    skip_in_apc: bool,
```
qwang98: Can remove this as it will always be true, now that we removed ApcEvent based record accounting and solely rely on skipping record in APC range during tracing VM?

crates/core/executor/src/tracing.rs
```diff
@@ -49,10 +50,85 @@ pub struct TracingVM<'a, M: ExecutionMode> {
...
+    /// The shard's full memory-read oracle, retained so per-APC blocks can be sliced out for their
+    /// [`ApcInvocation`] (or replayed on abort). Captured once at construction. Empty when there
+    /// are no APCs (skip gate inert) or on a replay VM.
+    shard_reads: Arc<[sp1_jit::MemValue]>,
+    /// Pre-state captured at each APC-start fetch, keyed by `(apc_id, entry cpu_event_count)`.
+    /// `cpu_event_count` is monotonic, so the key uniquely identifies the entry; a successful
+    /// `ApcCall` later looks itself up by `(apc_id, from.cpu_event_count)`.
+    apc_pre_states: HashMap<(usize, u32), CoreEntryState>,
+    /// Set when block-record capture should run (any APCs in the program).
+    capture_invocations: bool,
```
qwang98: Is this needed or always turned on?

crates/core/executor/src/tracing.rs
```diff
@@ -49,10 +50,85 @@ pub struct TracingVM<'a, M: ExecutionMode> {
...
+    fn step_instruction(&mut self) -> Result<CycleResult, ExecutionError> {
+        self.execute_instruction()
+    }
+}
+
+impl StepInstruction for TracingVM<'_, UserMode> {
+    fn step_instruction(&mut self) -> Result<CycleResult, ExecutionError> {
+        self.execute_instruction()
+    }
+}
```
qwang98, line 105: We don't really need this trait wrapper?

crates/core/executor/src/tracing.rs
```diff
@@ -649,10 +774,301 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+            // range without a successful `ApcCall` — i.e. the candidate aborted (bump /
+            // optimistic-constraint failure) or branched out. Its skipped prefix must be emitted
+            // as software. Doing this here, per-invocation, is what makes loops correct: a later
+            // invocation of the same `apc_id` can no longer discard this aborted one's flush.
+            if let Some(cs) = self.current_skip {
+                if apc_id != Some(cs.apc_id) {
+                    self.current_skip = None;
+                    self.flush_invocation(cs);
+                }
+            }
```
qwang98, line 859: Does this mean unsuccessful APC only reexecute software on the next cycle (or if segmentation interrupts it)?

crates/core/executor/src/tracing.rs
```diff
@@ -649,10 +774,301 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+                self.core.program.apcs.apc_indices_by_start_idx.get(&pc_idx).cloned()
+            {
+                // A still-unresolved invocation at a fresh APC start means its candidate aborted
+                // and execution looped straight back into an APC range without ever leaving it, so
+                // the range-exit flush above never fired. Flush it now, *before* its captured
+                // pre-state is overwritten below: while skipping, `cpu_event_count` (the pre-state
+                // key) does not advance, so this start would otherwise clobber the entry.
+                if let Some(prev) = self.current_skip.take() {
+                    self.flush_invocation(prev);
+                }
```
qwang98, line 873: Can this be integrated with the prior chunk? Not sure why it's written here disjoint.

```
                if apc_id != Some(cs.apc_id) {
                    self.current_skip = None;
                    self.flush_invocation(cs);
                }
```

crates/core/executor/src/tracing.rs
```diff
@@ -649,10 +774,301 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+                    registers: *self.core.registers(),
+                    pc: self.core.pc(),
+                    clk: self.core.clk(),
+                    global_clk: self.core.global_clk(),
+                    mem_reads_remaining: self.record.mem_reads_remaining,
+                };
+                let key_clk = self.record.cpu_event_count;
+                for &id in &apc_ids {
+                    self.apc_pre_states.insert((id, key_clk), entry.clone());
+                }
```
qwang98, line 865: Do not clone entry here, as it's only just constructed and not consumed below.

crates/core/executor/src/tracing.rs
```diff
@@ -649,10 +774,301 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+    /// static range contains the current PC, resolves any in-progress skip invocation that just
+    /// left its range (flushing its skipped prefix as software), captures the entry pre-state for
+    /// any APC starting at this PC, and begins this block's skip invocation. Returns the current
+    /// `apc_id` (`None` outside any APC range) for the caller to pass to [`Self::set_skip_gate`].
+    fn begin_cycle(&mut self) -> Option<usize>
+    where
+        for<'b> TracingVM<'b, M>: StepInstruction,
+    {
+        let pc_idx = self.core.pc().wrapping_sub(self.core.program.pc_base) as usize / 4;
+        let apc_id = self.apc_id_by_pc_idx.get(pc_idx).copied().flatten();
```
qwang98, line 827: How is `apc_id_by_pc_idx` different from `apc_indices_by_start_idx`? Why do we query both in this function?

crates/core/executor/src/autoprecompiles/mod.rs
qwang98: Claude answer: `clk_start` is the shard-relative timestamp clock that drives per-access `MemoryRecord` timestamps (+8/cycle); `global_clk_start` is the monotonic global cycle counter (+1/cycle, never resets across shards). Both are captured at block entry so re-execution restores the exact clock state.

crates/core/executor/src/autoprecompiles/mod.rs
qwang98: Claude answer: Only because `ExecutionRecord` derives `Serialize`/`Deserialize` and now holds `ApcInvocations`; `Arc<[MemValue]>` + offsets cannot derive, so the small `ApcInvocationSer` proxy serializes the slice as a `Vec`. It is never exercised on the in-process cuda path, so it can be dropped if `ExecutionRecord` stops deriving `Serialize`.

crates/core/executor/src/autoprecompiles/mod.rs
qwang98: Claude answer: Yes — `ExecutionRecord` derives `DeepSizeOf`, so every field must implement it, and `MemValue`/`Arc<[_]>` do not, hence the manual impl. Only removable if `ExecutionRecord` drops the `DeepSizeOf` derive.

crates/core/executor/src/autoprecompiles/mod.rs
qwang98: Claude answer: Done — the 37 write-only `*_len` fields were removed and the `ExecutionRecordSnapshotWithPc` wrapper collapsed (commits 1dd9e507e, 0763c133c). Only the three fields actually read by `apply_calls`/`capture_invocations` remain: `cpu_event_count`, `global_interaction_event_count`, `mem_reads_remaining`.

crates/core/executor/src/record.rs
qwang98: Claude answer: Will trim before finalizing — drop the `ApcEvents` history and keep it to what the field holds.

crates/core/executor/src/record.rs
qwang98: Claude answer: Will rename `skip_writes` -> `skip_emit`.

crates/core/executor/src/record.rs
qwang98: Claude answer: Agreed — I will condense the explanatory note at finalize. The `has_apc_events` accessor itself stays (it is used by `ApcChip::num_rows`).

crates/core/executor/src/tracing.rs
qwang98: Claude answer: Done — removed `skip_in_apc` in commit 01264ed6a; it is now derived from a single `capturing()` = `!apc_id_by_pc_idx.is_empty()`.

crates/core/executor/src/tracing.rs
qwang98: Claude answer: Always on under APC capture — removed in commit 01264ed6a, now derived via `capturing()` (same source as the old `skip_in_apc`). Skip and capture must agree, so one source of truth makes the invalid skip-without-capture state unrepresentable.

crates/core/executor/src/tracing.rs
qwang98: Claude answer: It is the minimal bridge so the mode-generic replay (`replay_block_into<M>`) can call the mode-specific `execute_instruction` (Supervisor and User differ). Without the trait the generic replay cannot dispatch to the right `execute_instruction`. Open to a cleaner form if you have one in mind.

crates/core/executor/src/tracing.rs
qwang98: Claude answer: Yes — an aborted APC block is re-executed as software when the PC leaves its range, detected at the next `begin_cycle` (the range-exit check), or on a fresh same-`apc_id` start (loop re-entry), or at shard end (segmentation). Deferring by one cycle is safe: nothing in between reads the skipped events.

crates/core/executor/src/tracing.rs
qwang98: Claude answer: They cover different triggers. This chunk (range-exit) fires when the PC leaves the APC range without a successful call. The start-block flush fires when a new APC start is reached while a stale invocation is still in flight — the loop-re-entry case where `cpu_event_count` did not advance, so range-exit never fired — and it must run there, before the pre-state is overwritten. Merging would lose that ordering; I can add a cross-referencing comment.

crates/core/executor/src/tracing.rs
qwang98: Claude answer: `apc_ids` can hold more than one id (overlapping APCs sharing a start pc), so `entry` is inserted once per id and reused across iterations — hence the clone. In the common single-start case it is one clone that could be a move; I will special-case len==1 to drop it.

crates/core/executor/src/tracing.rs
qwang98: Claude answer: `apc_id_by_pc_idx` maps each pc -> the APC whose range contains it (membership), checked every cycle for the skip gate. `apc_indices_by_start_idx` maps a pc -> the APCs that start there, checked only at a block entry to insert pre-states and begin the invocation. begin_cycle needs both: membership for the gate, starts to detect block entry.

crates/core/executor/src/tracing.rs
```diff
@@ -641,18 +739,365 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+            if has_apcs { trace.mem_reads().collect() } else { Arc::from([]) };
+
+        // Bump-resilient APC: under APC capture, let register (rr/rw) and RAM (mr/mw) accesses that
+        // cross a 2^24 timestamp epoch stay APCs instead of aborting to software. Register
+        // crossings are routed to the shared `MemoryBump` chip via `pending_register_bumps`; RAM
+        // crossings are represented natively by `compare_low` (no bump needed). State/pc bumps and
+        // `register_refresh` still abort.
+        let mut core = CoreVM::new(trace, program, opts, proof_nonce);
+        core.apc_register_bump_tolerant = has_apcs;
+        core.apc_ram_bump_tolerant = has_apcs;
```
qwang98, line 778: These can be removed once it's the default method.

crates/core/executor/src/tracing.rs
```diff
@@ -641,18 +739,365 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+            pending_register_bumps: Vec::new(),
+            _mode: PhantomData,
+        }
+    }
+
+    /// Lightweight constructor for record-in-chip block replay ([`Self::reexecute_apc_block_into`]).
+    /// Uses an untracked [`CoreVM`] (no candidate work), an empty APC pc-map (every lookup →
+    /// `None`, so the gate stays inert), and no read-oracle copy. Capture is off. Candidate
+    /// detection never fires during replay, so this is correct and cheap.
+    fn new_replay<T: MinimalTrace>(
```
qwang98, line 801: Note: FYI this is basically the "throwaway" VM used for rollback replay and reexecute from ApcInvocation, so that we don't need to restore main VM state for the former case and don't need to construct everything from scratch for the latter case.

crates/core/executor/src/tracing.rs
```diff
@@ -641,18 +739,365 @@ impl<'a, M: ExecutionMode> TracingVM<'a, M> {
...
+            for (apc_id, apc) in program.apcs.apc_by_index.iter().enumerate() {
+                let start = apc.start_pc_idx();
+                let end = (start + apc.num_cycles()).min(apc_id_by_pc_idx.len());
+                if start < apc_id_by_pc_idx.len() {
+                    for slot in &mut apc_id_by_pc_idx[start..end] {
+                        *slot = Some(apc_id);
+                    }
+                }
+            }
+        }
```
qwang98, line 764: Note: this map is built so that we can do O(1) indexing into a vec to know when we enter or leave an APC range. It's purely for efficiency. It's built once per shard (so we could technically further optimize to build once only, as it's the same program), but it shouldn't affect run time though.

However, one real short coming is that it's a static map that happens to take the largest APC id for each PC. For non overlapping APCs (assuming we don't do super blocks), this is fine. For superblock support, we need a better mechanism to track APC range entry and exit.

crates/core/executor/src/vm.rs
```diff
@@ -72,14 +72,36 @@ pub struct CoreVM<'a, M: ExecutionMode, S> {
     decoded_instruction_cache: HashMap<u32, Instruction>,
     /// The candidate tracker for autoprecompiles.
     pub apc_candidates: ApcCandidates<CoreExecutionState, Apc, S>,
+    /// Whether to run APC-candidate tracking (`try_insert` on fetch, `check_conditions` /
+    /// `extract_calls` on advance). Disabled for record-in-chip block replay
+    /// ([`CoreVM::new_untracked`]), which ignores candidate output: this avoids both the
+    /// per-construction clone of `apc_by_index` and the per-cycle candidate work — the dominant
+    /// cost of block re-execution (hundreds of thousands of calls per run).
+    track_apc_candidates: bool,
```
qwang98, line 80: Note: This is needed when constructing the light weighted VM for rollback software reexecution and in chip record gen.

PR #2869, opened 2026-07-09
feat: (powdr) tolerate memory timestamp epoch change in autoprecompile

qwang98:
Ready for review. Mirrors the tolerate memory timestamp bump solution in #2864, but yanked out to fit the current under review executor (which extracts ApcEvents out of the main record instead of pushing record gen to the chips).

Overall, the solution:
1. No longer abort APC when register/non-register timestamp epoch change happens during APC execution.
2. Still emits such memory bump events to the memory bump chip, so bus is balanced.
3. Correctness is ensured by the fact that memory bump chip straddles epoch change for register timestamp bump, and that non-register epoch change is already natively checked by sp1 memory chip constraints.

Confirmed to work on RSP when this PR is cherry picked to top of #2781. We should merge it to #2738 first, whose branch this PR is created against.


crates/core/executor/src/record.rs
```diff
@@ -962,7 +959,13 @@ impl ExecutionRecord {
...
                         cpu_local_memory_access,
                         syscall_events,
-                        bump_memory_events,
+                        // Do NOT carve `bump_memory_events` into APC block records: the APC chip
+                        // fills its columns only from the block's *instruction* chips (see
+                        // `chip.rs`), so a `MemoryBump` dummy trace from a carved bump is discarded
+                        // and its bus interaction is lost. Every bump stays in the main record,
+                        // whose `MemoryBump` chip emits it — matching the APC's re-anchored
+                        // register receive (`RegisterAccessTimestamp` re-anchors `prev_low -> 0`).
+                        bump_memory_events: Vec::default(),
```
qwang98, line 967: This is the key adaptation of the no bump solution to the current executor. Previously we extract bump events out of the main record into ApcEvents, which ignores it in APC gen deps/trace gen. For the epoch straddling pair of buses to be generated, memory bump events must stay in the main record. Otherwise, memory bump bus interactions aren't generated and buses won't balance.

crates/core/executor/src/tracing.rs
```diff
@@ -2335,22 +2330,28 @@ mod tests {
...
-            // Index 5: ADDI x12, x11, 1 — post-loop: reads x11 (stale from epoch 0)
-            Instruction::new(Opcode::ADDI, 12, 11, 1, false, true),
-            // Index 6: ADDI x13, x12, 1
-            Instruction::new(Opcode::ADDI, 13, 12, 1, false, true),
-            // Index 7: ADDI x14, x13, 1
+            // Index 6: LW x12, [addr] — post-loop: reads RAM `addr` (stale from epoch 0)
+            Instruction::new(Opcode::LW, 12, 0, addr, false, true),
+            // Index 7: ADDI x13, x11, 1 — post-loop: reads x11 (stale from epoch 0)
+            Instruction::new(Opcode::ADDI, 13, 11, 1, false, true),
+            // Index 8: ADDI x14, x13, 1
```
qwang98, line 2354: These diff just create the test program for `test_apc_memory_bump_tolerated`, renamed from `test_apc_memory_bump_error`.

qwang98: Here we need to test both register and RAM access for timestamp epoch crossing.

crates/core/executor/src/tracing.rs
```diff
@@ -2360,35 +2361,53 @@ mod tests {
...
+            "Expected a register bump event emitted to the MemoryBump chip for the crossing"
+        );
+        // ...and that bump must stay in the MAIN record: the APC chip fills its columns only from
+        // block-instruction chips, so a bump carved into the APC block record is silently dropped
+        // (the APC emits no MemoryBump interaction), which would leave the re-anchored register
+        // receive unmatched on the memory bus. Guard that no APC record carries a carved bump.
+        assert!(
+            record.apc_events.events.values().all(|e| e.record.bump_memory_events.is_empty()),
+            "APC block records must not carry carved bump events (they belong in the main record)"
         );
```
qwang98, line 2411: This tests that memory bump event must stay in main record and never ApcEvent record.

crates/core/executor/src/vm.rs
```diff
@@ -107,18 +107,6 @@ impl ExecutionState for CoreExecutionState {
...
-    #[inline]
-    fn abort_if_epoch_changed(
-        candidates: &mut ApcCandidates<CoreExecutionState, Apc, S>,
-        t0: u64,
-        t1: u64,
-    ) {
-        if t0 >> 24 != t1 >> 24 {
-            candidates.abort_in_progress();
-        }
-    }
```
qwang98, line 120: We created this in #2738. Now it's removed, together with the call sites.

crates/core/executor/src/record.rs
```diff
@@ -962,7 +959,13 @@ impl ExecutionRecord {
                         global_memory_finalize_events,
                         cpu_local_memory_access,
                         syscall_events,
-                        bump_memory_events,
+                        // Do NOT carve `bump_memory_events` into APC block records: the APC chip
+                        // fills its columns only from the block's *instruction* chips (see
+                        // `chip.rs`), so a `MemoryBump` dummy trace from a carved bump is discarded
+                        // and its bus interaction is lost. Every bump stays in the main record,
+                        // whose `MemoryBump` chip emits it — matching the APC's re-anchored
+                        // register receive (`RegisterAccessTimestamp` re-anchors `prev_low -> 0`).
```
Schaeff, line 966: This comment could be removed. It is the same reasoning as `byte_lookups` lower down, which does not have a comment: these events are proven by other chips, so they should not be moved to the apc record.

Schaeff (approved):

PR #2872, opened 2026-07-10
feat: (powdr) cache APC trace to skip regeneration in trace generation

qwang98:
This PR pretty much removes the entire generate trace time for APC (though generate trace still takes time but almost entirely for non APC chips), because generate dependencies already creates the same trace, which we simply attach to the APC chip for later use. The only time spent in APC trace gen is copying the cached trace to the trace gen buffer. We can't directly cache APC trace to the trace gen buffer at generate dependencies, because the buffer isn't allocated till later. Copying is fine as it's much faster than running the whole thing again, regardless of how we do generate trace (whether the current record extraction based, or in the future chip record generation based), so this solution should be included in the main branch regardless.


crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -114,6 +134,26 @@ impl<F: PrimeField32> MachineAir<F> for ApcChip<F> {
...
+            .remove(&(input.public_values.proof_nonce, input.initial_timestamp));
+        if let Some(cached) = cached {
+            let n = cached.len();
+            debug_assert_eq!(n, events.count * self.width());
+            // SAFETY: `n == cached.len()`, so the first `n` slots are filled from `cached` and the
+            // rest (padding rows) are zeroed. `MaybeUninit<F>` has the same layout as `F`.
+            unsafe {
+                core::ptr::copy_nonoverlapping(cached.as_ptr(), buffer.as_mut_ptr().cast::<F>(), n);
+                core::ptr::write_bytes(buffer[n..].as_mut_ptr(), 0, buffer.len() - n);
+            }
```
qwang98: This copy from the trace cached in generate dependencies to the trace buffer used for proving is not avoidable.

Trace gen allocates a contiguous buffer before each chip runs `generate_trace_into`, which takes a pointer to an offset in this contiguous buffer as an input param. This allocation is run after generate dependencies, so a previous construct I attempted (which adds the pointer into the contiguous buffer as an APC chip field, so that generate dependencies can directly write into this buffer) couldn't work.

Fortunately, copying is much faster than recomputing the buffer from events, so this construct is definitely still worth it.

crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -67,11 +66,32 @@ pub struct ApcChip<F: PrimeField32> {
...
+    cached_traces: Mutex<HashMap<([u32; 4], u64), Vec<F>>>,
+}
+
+// Manual `Debug` (skips `cached_traces`/`machine`) — a derived impl would dump the whole trace
+// cache, and the cache doesn't participate in a chip's identity.
+impl<F: PrimeField32> std::fmt::Debug for ApcChip<F> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ApcChip").field("id", &self.id).field("name", &self.name).finish()
+    }
 }
```
qwang98: We don't want to derive Debug on the cached trace here as it'll be too large. However there's no flag that turns off Debug derive just for this field (unless we import some other crates), so I decided the best is to manually implement Debug here.

crates/prover/src/shapes.rs
```diff
@@ -784,6 +784,7 @@ pub fn max_count(a: RecursionAirEventCount, b: RecursionAirEventCount) -> Recurs
     }
 }
 
+#[allow(clippy::mutable_key_type)]
```
qwang98, line 787: These, sprinkled across this PR, are linting fixes. Here in this specific fix, `Chip` contains ApcChip, which we added the new mutable cached trace field to. Therefore, the following `BTreeSet` which uses `Chip` as a key requires this allow flag for linting.

In practice, having a mutable flag here is fine, because the `Eq` `Ord` impl of ApcChip is based on its name, rather than the cached trace, which is the only mutable field.

crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -67,11 +66,35 @@ pub struct ApcChip<F: PrimeField32> {
     cached_apc: CachedApc<F>,
     /// A machine to generate traces for the APC. By construction, it will never have apcs itself.
     machine: Machine<F, RiscvAir<F>>,
+    /// Cache of the filled APC trace, keyed by `(proof_nonce, initial_timestamp)`.
+    /// `generate_dependencies` fills the trace to evaluate the byte-bus interactions and stores it
+    /// here so `generate_trace_into` can restore it instead of regenerating the identical trace.
+    /// `Mutex` because the chip is shared via `Arc` across concurrent shard workers.
+    cached_traces: Mutex<HashMap<([u32; 4], u64), Vec<F>>>,
```
qwang98: Cache is by proof nonce and initial timestamp, because:
- Proof nonce is for disambiguating the same Apc across different proofs
- Timestamp is for disambiguating the same Apc across different shards/workers within the same proof

qwang98: Mutex shouldn't be blocking as the lock in practice is only used for removing the cache entries, while the copying of the cache is done later.

crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -111,147 +134,30 @@ impl<F: PrimeField32> MachineAir<F> for ApcChip<F> {
...
+        let cached = self
+            .cached_traces
+            .lock()
+            .unwrap()
+            .remove(&(input.public_values.proof_nonce, input.initial_timestamp))
+            .unwrap_or_else(|| {
+                panic!(
+                    "APC chip {} trace not cached: generate_dependencies must run before generate_trace_into",
+                    self.id
+                )
```
qwang98, line 151: This in practice should never happen, unless we never run generate dependencies for an APC chip before we run generate trace.

crates/core/machine/src/autoprecompiles/chip.rs
Schaeff: Would be cleaner to create a type for the cached trace and implement debug there, then we can keep this derive.

Schaeff:
What about [this change](https://github.com/succinctlabs/sp1/commit/7cfe823c85c9f7c1d34c606e1684e615c57682c6) to simplify and remove the copy? As far as I understand, `generate_trace_into` is only called from `generate_trace` with a buffer allocated there. If we instead implement `generate_trace` directly, we can simply move the buffer out of the cache rather than copy.

qwang98:
> What about [this change](https://github.com/succinctlabs/sp1/commit/7cfe823c85c9f7c1d34c606e1684e615c57682c6) to simplify and remove the copy? As far as I understand, `generate_trace_into` is only called from `generate_trace` with a buffer allocated there. If we instead implement `generate_trace` directly, we can simply move the buffer out of the cache rather than copy.

This solution would work but ONLY for the CPU case. Here's the CPU main trace gen function which calls per AIR `generate_trace`, exactly as you pointed out. It outputs a map from AIR to the AIR's trace.
```
async fn generate_main_traces(
        &self,
        record: A::Record,
        max_log_row_count: usize,
        prover_permits: ProverSemaphore,
    ) -> MainTraceData<F, A, CpuBackend> {
        let airs = self.machine.chips().to_vec();
        let (tx, rx) = oneshot::channel();
        // Spawn a rayon task to generate the traces on the CPU.
        slop_futures::rayon::spawn(move || {
            let chips_and_traces = airs
                .into_par_iter()
                .filter(|air| air.included(&record))
                .map(|air| {
                    let trace = air.generate_trace(&record, &mut A::Record::default());
                    let trace = Mle::from(trace);
                    (air, trace)
                })
                .collect::<BTreeMap<_, _>>();
```

Here's the GPU pipeline for generating trace. It slices into one contiguous buffer with an offset and then calls `generate_trace_into` for each AIR, so we can't really allocate our own per AIR buffer via `generate_trace`. The crux of the issue is that we materialize the trace during generate dependencies, which happens before the buffer allocation for generate trace happens, so there's no way to avoid the copy for the GPU path.
```
let (host_traces_tx, host_traces) = futures::channel::mpsc::unbounded();
            rayon::spawn(move || {
                jobs.into_par_iter().for_each_with(host_traces_tx, |tx, (air, offset)| {
                    tracing::trace_span!(parent: &outer_span, "chip host main tracegen", chip = %air.name()).in_scope(
                        || {
                            let base_ptr = buffer_ptr as *mut MaybeUninit<Felt>;
                            let height = air.num_rows(&record).unwrap();
                            let width = air.width();
                            let trace_len = height * width;
                            let slice: &mut [MaybeUninit<Felt>] = unsafe {
                                std::slice::from_raw_parts_mut(base_ptr.add(offset), trace_len)
                            };
                            air.generate_trace_into(&record, &mut A::Record::default(), slice);
                            let start_pointer = unsafe { base_ptr.add(offset) as usize };
                            // If the receiver dropped (task cancelled), exit gracefully instead of panicking.
                            let _ = tx.unbounded_send((air.name().to_string(), start_pointer, height, width));
                        },
                    );
                });
```

I haven't looked deep into how we could potentially adapt the GPU pipeline to not use one pre-allocated contiguous trace gen buffer, as I assumed we won't want to change that part of the pipeline, which can be more general to sp1.

crates/core/machine/src/autoprecompiles/chip.rs
qwang98: I seem to have missed this comment but it's fixed now!

crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -67,11 +80,19 @@ pub struct ApcChip<F: PrimeField32> {
...
 
 impl<F: PrimeField32> ApcChip<F> {
     pub fn new(apc: Arc<Sp1Apc<F>>, id: usize) -> Self {
-        Self { id, name: format!("APC_{id}"), cached_apc: apc.into(), machine: RiscvAir::machine() }
+        Self {
+            id,
+            name: format!("APC_{id}"),
+            cached_apc: apc.into(),
+            machine: RiscvAir::machine(),
+            cached_traces: CachedTraces(Mutex::new(HashMap::new())),
```
Schaeff:
```suggestion
            cached_traces: CachedTraces::default(),
```

crates/prover/src/shapes.rs
Schaeff: There's an alternative way of fixing this in a centralized way in `clippy.toml`:
```
ignore-interior-mutability = [
    "bytes::Bytes",
    "sp1_core_machine::autoprecompiles::chip::ApcChip",
]
```

PR #2874, opened 2026-07-10
fix(gpu): 16-byte align poseidon2 buffers reinterpreted as FDW_t (Blackwell misaligned-address)

tamirhemo:
## Problem

Every GPU proof — any program, `core` or `compressed` — aborts with:

```
thread 'tokio-rt-worker' panicked at sp1-gpu/crates/cuda/src/stream.rs:340:18:
called `Result::unwrap()` on an `Err` value: Other(CudaRustError: misaligned address)
```

surfacing in the FRI commit phase (`MerkleTreeSingleLayerProver::commit_tensors`
under `FriCudaProver::commit_phase_round`). Reproduced on an RTX 5090
(sm_120 / Blackwell, CUDA 13.3, driver 580) with both `fibonacci --mode core`
and `rsp --mode compressed`, on `main` as well as feature branches — it is not
specific to any one branch.

## Root cause

`poseidon2.cuh` casts several `F_t` buffers to `FDW_t` — a `__align__(16)`
vector type — and performs 128-bit loads/stores through the cast:

- the local `state[WIDTH]` arrays in `Hasher::compress` and `Hasher::hash`, and
- the `data[WIDTH]` member of `struct HasherState` (via `finalize`).

Those buffers were only `alignof(F_t)` (4-byte) / `alignof(size_t)` (8-byte)
aligned, so the vector access is misaligned. Older archs kept the buffers in
registers (or tolerated the access), so it never faulted; on Blackwell the
access goes through local memory and faults with `misaligned address`.

## Fix

Force 16-byte alignment on the three buffers to match `FDW_t`.

## Testing

On an RTX 5090, after the fix:
- `fibonacci --mode core` → proves in ~0.4s
- `rsp --mode compressed` → proves in 84.7s ± 1.2s (5 reps), 525.3M cycles

Both previously aborted with the misaligned-address panic.


erabinov (approved):
LGTM!

PR #2877, opened 2026-07-13
fix(verifier): restore no_std Groth16 and Plonk builds

fakedev9999:
## What
- Keep Groth16 and Plonk verification available without default features.
- Gate compressed proof types and recursion dependencies behind the default `full` feature.
- Precompute the v6.3.1 recursion VK root and test it against the serialized verifier keys.
- Add a `wasm32-unknown-unknown` no-default-features CI check.

## Why
`sp1-verifier` v6.3.1 pulls recursion and slop dependencies unconditionally. Those dependencies pull `getrandom` and prevent WASM runtimes from compiling even when they only use the Groth16 or Plonk verifier.

This is a v6.3.1-compatible follow-up to #2698. It preserves the existing default full API and public lazy-static VK constants.

## Validation
- `cargo +1.88.0 check --target wasm32-unknown-unknown --no-default-features -p sp1-verifier`
- `cargo check -p sp1-verifier --all-features`
- `cargo check -p sp1-sdk -p sp1-prover -p sp1-recursion-gnark-ffi`
- `cargo test -p sp1-verifier test_vk_root_bytes`
- `cargo clippy -p sp1-verifier --all-targets --all-features -- -D warnings`
- `cargo package -p sp1-verifier --allow-dirty --no-verify`
- Avail's `pallet-vector` dependency tree with this patch no longer contains the SP1 recursion/slop/`getrandom` path.

## Notes
A direct standalone `pallet-vector` WASM check progressed past the SP1 dependency issue but stopped in Avail's existing `blst` C build, which does not support the direct `wasm32-unknown-unknown` target on this macOS host.


erabinov (approved):
LGTM!

PR #2878, opened 2026-07-13
feat: (powdr) parallelize APC dependency generation

qwang98:
Current holdup: AIR name based ("APC_" prefix) lookup of APC chips might not be reliable. However, the speed up is read and we might merge in the future.

Small but effective parallelization change that worked. Note that generate_dependencies is entirely sequentially so far. This is the within shard parallelization for generate_dependencies over APC chips, thanks to its clean dependency structure (only produces byte lookup events, so as long as it's run before the byte lookup chip, APC chips can be parallelized however we want).

### DIFF
--- a/crates/hypercube/src/machine.rs
+++ b/crates/hypercube/src/machine.rs
@@ -3,6 +3,7 @@ use slop_algebra::Field;
 use std::collections::BTreeSet;
 
 use crate::{air::MachineAir, Chip, MachineRecord};
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 
 /// A shape for a machine.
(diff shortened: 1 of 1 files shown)


crates/hypercube/src/machine.rs
```diff
@@ -106,7 +107,27 @@ where
...
+                    .map(|chip| {
+                        let mut output = A::Record::default();
+                        chip.generate_dependencies(record_ref, &mut output);
+                        output
+                    })
+                    .collect();
+                for r in &mut apc_records {
+                    record.append(r);
+                }
+            }
```
qwang98, line 130: Run generate_dependencies in parallel (need read only reference access to record). Append to main record after (need write access to record).

PR #2879, opened 2026-07-13
feat: (powdr) accumulate APC byte lookups per chunk instead of per row

qwang98:
Here's another small bite sized PR that will benefit APC regardless of what execution architecture we use eventually, because we will still need to generate trace and side effects. It basically parallelizes over # of CPUs instead of # of rows, thereby reducing the # of hash maps created for side effects accounting from hundreds of thousands to just a handful.

### DIFF
--- a/crates/core/machine/src/autoprecompiles/chip.rs
+++ b/crates/core/machine/src/autoprecompiles/chip.rs
@@ -15,7 +15,7 @@ use slop_algebra::PrimeField32;
 use slop_matrix::Matrix;
 use slop_maybe_rayon::prelude::{
     IndexedParallelIterator, IntoParallelIterator, IntoParallelRefIterator, ParallelIterator,
-    ParallelSliceMut,
+    ParallelSlice, ParallelSliceMut,
 };
 use sp1_core_executor::{
     events::ByteLookupEvent, opcode::ByteOpcode, ApcRange, ExecutionRecord, Program, RiscvAirId,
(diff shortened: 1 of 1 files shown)


crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -369,70 +369,76 @@ impl<F: PrimeField32> MachineAir<F> for ApcChip<F> {
...
-            .zip_eq(dummy_values_by_event.par_iter())
-            .map(|(trace_row, dummy_values_by_instruction)| {
-                for (instruction_index, dummy_slice) in
-                    dummy_values_by_instruction.iter().enumerate()
+            .par_chunks_mut(chunk_size * trace_width)
+            .zip_eq(dummy_values_by_event.par_chunks(chunk_size))
+            .map(|(trace_chunk, dummy_chunk)| {
+                // Store effects in a map of ByteLookupEvent to count to apply after parallel
+                // execution
+                let mut byte_lookup_effect = HashMap::new();
```
qwang98, line 380: Here is the crux of the PR:
- Previously we parallelize over each trace row, calculate the trace, and for the row create a map of byte lookup events. Then after all rows are done, we merge all the per row HashMaps, which is obviously not very efficient.
- Now we parallelize over the # of CPUs for both trace calculation and byte lookup event map, which means that we will have # of HashMaps equal to # CPUs instead of hundreds of thousands of maps which create significant overhead.

crates/core/machine/src/autoprecompiles/chip.rs
```diff
@@ -369,70 +369,76 @@ impl<F: PrimeField32> MachineAir<F> for ApcChip<F> {
...
-                    let map = &dummy_trace_index_to_apc_index_by_instruction[instruction_index];
-                    // By caching `dummy_trace_index_to_apc_index_by_instruction`, we only loop over
-                    // the values that are assigned to the APC instead of all values in the dummy
-                    // trace
-                    for (dummy_index, apc_index) in map.iter() {
-                        trace_row[*apc_index] = dummy_slice[*dummy_index];
+                    for (instruction_index, dummy_slice) in
+                        dummy_values_by_instruction.iter().enumerate()
+                    {
+                        let map = &dummy_trace_index_to_apc_index_by_instruction[instruction_index];
```
Schaeff:
```suggestion
for (dummy_slice, map) in dummy_values_by_instruction
                    .iter()
                    .zip(&dummy_trace_index_to_apc_index_by_instruction)
                {
```

Schaeff (approved):
lgtm, small suggestion not directly related to this PR

crates/core/machine/src/autoprecompiles/chip.rs
qwang98: I applied it.

PR #2889, opened 2026-07-16
fix(cli): surface non-zero child exits in toolchain install/build

### LINKED ISSUE #2856: cargo prove toolchain commands can ignore non-zero child process exits
### Component

cargo prove CLI/sp1up

### Version

Current `main` at `9069dd60f`.

### Operating system

macOS (Apple Silicon)

### Describe the bug

While auditing the CLI/toolchain paths, I noticed two places where child process failures can be treated as success because only process spawning is checked, not the returned exit status.

In `crates/cli/src/commands/install_toolchain.rs`, both of these commands use `.status()?` without checking `ExitStatus::success()`:

- `tar -xzf ...` at lines 137-140
- `rustup toolchain link ...` at lines 151-159

`Command::status()` only returns an `Err` when the process could not be spawned or waited on. A command that runs and exits non-zero still returns `Ok(status)`. That means a failed `tar` extraction or failed `rustup toolchain link` can be ignored, and the code can continue to print `Successfully linked toolchain to rustup.` after a non-zero link exit.

There is a similar pattern in `crates/cli/src/commands/build_toolchain.rs`: `CommandExecutor::run()` calls `.output().map(|_| ())`, so commands such as `git clone`, `python3 x.py build`, `rustup toolchain link`, and `tar` are considered successful as long as the process was spawned, regardless of its exit code.

Minimal local reproduction of the Rust behavior:

```rust
use std::process::Command;

fn main() -> std::io::Result<()> {
    let status = Command::new("sh").arg("-c").arg("exit 42").status()?;
    println!("status_success={}", status.success());
    println!("function_returned_ok=true");
    Ok(())
}
```

Output:

```text
status_success=false
function_returned_ok=true
```

I did not open a PR yet because this changes CLI failure behavior, and I wanted to confirm the desired scope first. If this matches the intended direction, a narrow fix could add a small helper that checks `ExitStatus::success()` and returns an error with command context, then use it for the toolchain install/build commands with a regression test around a command that exits non-zero.

Expected behavior: non-zero child process exits in these toolchain paths should stop the command and surface a clear error.

Actual behavior: non-zero exits can be silently ignored if the process itself was spawned successfully.

MavenRain:
## Motivation

Fixes #2856.

Several `cargo prove` toolchain commands treat child-process failures as success. `Command::status()` returns `Err` only when the process cannot be spawned or waited on — a process that runs and exits **non-zero** still returns `Ok(status)`. The affected paths only checked the `Ok`/`Err` split, never `ExitStatus::success()`:

- `crates/cli/src/commands/install_toolchain.rs` — the `tar -xzf ...` extract and the `rustup toolchain link ...` calls used `.status()?` with no success check, so a failed extraction or link was ignored and the command still printed `Successfully linked toolchain to rustup.`
- `crates/cli/src/commands/build_toolchain.rs` — the shared `CommandExecutor::run()` helper called `.output().map(|_| ())`, so `git clone`, `git submodule update`, `python3 x.py build`, `rustup toolchain link`, and `tar` were all considered successful as long as the process spawned, regardless of exit code.

The net effect is that a broken toolchain install/build can be reported as successful, surfacing later as confusing downstream failures instead of a clear error at the point of failure.

## Solution

Check the exit status wherever these toolchain commands run to completion, and stop with a clear, command-tagged error on a non-zero exit:

- **`build_toolchain.rs`** — `CommandExecutor::run()` now drives the command with `.status()` (stdio inheritance is unchanged) and returns an error unless `status.success()`. Because every command in that module goes through `run()`, this fixes all of them at once.
- **`install_toolchain.rs`** — adds a small private `run_checked(&mut Command)` helper with the same success check, used for the two bare `.status()?` sites (`tar` extract, `rustup toolchain link`).

Intentionally left as-is: the "rustup toolchain remove" paths (which should tolerate "no toolchain installed") and the `rustup --version` / `forge --version` "is it installed?" probes, which legitimately only care whether the process could be spawned. I also audited the rest of `crates/cli`; the one other capturing call site (`new.rs` git clone) already checks `output.status.success()`.

Regression tests (`#[cfg(unix)]`, using `sh -c "exit N"`) assert a non-zero child exit becomes an error and a zero exit returns `Ok` in both modules.

Behavior for the success path is unchanged; only genuinely failing commands now abort with an error instead of being silently ignored.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation (doc comment on the new `run_checked` helper)
- [ ] Breaking changes


fakedev9999 (approved):
Thanks @MavenRain, lgtm. I added a follow-up to exclude credentials from command errors, with tests covering nonzero exits and spawn failures.

PR #2898, opened 2026-07-23
feat: (powdr) update to use latest powdr

qwang98:
Update #2781 to latest powdr as well. Partial update already done for #2738. This PR additionally pick up some other changes from the diff (but not all the same as the base changed): https://github.com/succinctlabs/sp1/compare/29ee613ab60e0de1d2a61016dd8bd084ceae4386...b80388257928b82185bf4bdc002dfabb6deac17f


crates/core/machine/src/autoprecompiles/chip.rs
qwang98, line 1: Ignore this file's diffs entirely. They are from the cached traces which is already reviewed and merged.

crates/core/machine/src/autoprecompiles/mod.rs
qwang98, line 1: Should be reviewed.

qwang98: It's a direct port of the diff here: https://github.com/succinctlabs/sp1/compare/29ee613ab60e0de1d2a61016dd8bd084ceae4386...b80388257928b82185bf4bdc002dfabb6deac17f

crates/prover/src/shapes.rs
qwang98, line 1: Should be reviewed. It's just the call site of the new config API.

crates/sdk/src/lib.rs
qwang98, line 1: Should be reviewed. It's just the call site of the new config API.

sp1-gpu/crates/perf/src/bin/node.rs
qwang98, line 1: Should be reviewed. It's just the call site of the new config API.

Cargo.toml
qwang98, line 1: Should be reviewed. Note that it's pinned to the latest powdr commit, because we don't have release after `0.1.4`.

qwang98:
Stale bump that the base already moved past.

PR #2899, opened 2026-07-23
feat: (powdr) support derived columns in APC trace generation

qwang98:
The purpose of this PR is to support derived columns, including those that read values from removed columns in the dummy trace, with the minimum diff possible. Eventually, I foresee using a shared powdr API between sp1 and powdr-openvm, but that'll be larger refactoring so deferred to the future, after we have full e2e powdr sp1 leanr proof.

Only touches files under `autoprecompiles`.

(merged) Depend on https://github.com/powdr-labs/powdr/pull/3801


qwang98:
GPU CI failure seem network and unrelated to code. Force push was a rebase and code content still the same.

qwang98:
Closed now as #2912 is here. This PR's commit history is a bit too messy.

PR #2904, opened 2026-07-28
fix(prover): drop duplicate shard proofs from CoreExecute redelivery

Farhad-Shabani:
## Summary

The cluster delivers tasks at-least-once, but a re-delivered `CoreExecute` re-runs the full execution and streams a second, identical set of per-shard `ProofData` into the controller. The `CompressTree` assumes disjoint shard ranges, so the duplicates corrupt it and proving wedges silently — no failure, no status write — until the request deadline evicts the proof hours later.

## Incidents
- 2026-07-15, [mainnet request](https://explorer.succinct.xyz/request/0x38a5edaff6dcf00aff3c5a1492bd05b5e71282a4b25a0336ee5a00841546c399) — 45M-cycle proof, executed in 1.5s. A CPU worker's heartbeat arrived 35s late under a 5× traffic spike; the coordinator declared it dead and re-delivered its in-flight `CoreExecute` while the original run was still going. Both runs streamed a full shard set (30 ProveShards for a ~15-shard proof). All shards and reduces completed, then the controller went silent at the reduce→wrap boundary for 3h54m.
- Same day, 1aee2eb0… — identical wedge signature; here the worker genuinely died mid-run (gnark panic), so the redelivery was legitimate. Both flavors — false-positive and real death — poison the tree the same way.
- 2026-07-27 recurrence — same signature on another request from the same customer, ~2h01m customer latency; independently root-caused to the same retry bug.

**Connected PR in sp1-cluster:** https://github.com/succinctlabs/sp1-cluster/pull/163

### DIFF
--- a/crates/prover/src/worker/controller/mod.rs
+++ b/crates/prover/src/worker/controller/mod.rs
@@ -25,7 +25,7 @@ use sp1_core_executor::SP1CoreOpts;
 use sp1_core_executor_runner::MinimalExecutorRunner;
 use sp1_core_machine::{executor::ExecutionOutput, io::SP1Stdin};
 use sp1_hypercube::{
-    air::{PublicValues, PROOF_NONCE_NUM_WORDS},
+    air::{PublicValues, ShardRange, PROOF_NONCE_NUM_WORDS},
     SP1PcsProofInner, SP1VerifyingKey, ShardProof,
 };
 use sp1_primitives::{io::SP1PublicValues, SP1GlobalContext};
@@ -33,7 +33,7 @@ use sp1_prover_types::{
     network_base_types::ProofMode, Artifact, ArtifactClient, ArtifactType, TaskStatus, TaskType,
 };
 use sp1_verifier::{ProofFromNetwork, SP1Proof};
-use std::{borrow::Borrow, sync::Arc};
+use std::{borrow::Borrow, collections::HashSet, sync::Arc};
 use tokio::{
     sync::{oneshot, Mutex, MutexGuard},
     task::JoinSet,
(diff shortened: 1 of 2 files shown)


crates/prover/src/worker/controller/mod.rs
```diff
@@ -140,11 +140,46 @@ where
     /// Execute Risc-V program, and trigger shard proofs for each trace chunk.
     /// Run the core executor and deferred proof emitter for a `CoreExecute` task. Proof shards
     /// are streamed back to the consumer via the task's message channel.
+    ///
+    /// Idempotent under redelivery: if the execution output artifact already exists, a prior
```
0xernesto: This idempotency claim does not cover Core proof mode.

This guard only helps after `execution_output` is uploaded. If redelivery overlaps the original run, duplicate `ProofData` can still reach `collect_core_proofs`, which accepts every message. Should add the same task/range dedupe there before uploading the Core proof artifact

0xernesto (approved):

crates/prover/src/worker/controller/mod.rs
Farhad-Shabani: [042635e](https://github.com/succinctlabs/sp1/pull/2904/commits/042635efdbee8e7b1a3f384a8afa9e7c3550a358)

tamirhemo:
Seems like this mechanism works according to task IDs. What about tasks that failed and get re-submitted as another task? is task id duplication what we are trying to defend against? Not sure what the exact idempodency relation here is. 

Also the filter map seems to contain all tasks in it's state, it could be ok if that's what is needed but it is a breakout from the compress tree original scope

Farhad-Shabani:
> Seems like this mechanism works according to task IDs. What about tasks that failed and get re-submitted as another task? is task id duplication what we are trying to defend against? Not sure what the exact idempodency relation here is.
> 
> Also the filter map seems to contain all tasks in it's state, it could be ok if that's what is needed but it is a breakout from the compress tree original scope

@tamirhemo actually range is the identity. Every ProveShard submit mints a fresh id, so a redelivery re-proves the same ranges under new ids — task id can't catch that. The task id set is for a different case: the coordinator replays buffered messages to reconnecting subscribers, so the same one can land twice.
Your retry case keeps the id.  `fail_task` bumps `task.retries` and re-enqueues the same task, so it never reaches the filter.
On scope, you're right. I did look at folding the task id half into `core_proof_map`, but that drains as proofs get consumed so it can't answer "seen before". Maybe worth moving the whole thing into the tree's ingest — want it to live there?

tamirhemo (approved):

PR #2913, opened 2026-07-29
perf: Cache more in APC chip

Schaeff:
- precompute everything that does not depend on the input / events
- add one indirection to please clippy (`CachedApc -> Arc<CachedApc>`)


qwang98 (approved):
LGTM as in:
1. CachedApc computation doesn't touch anything proof time
2. Cached structs are all used and nothing redundant calculated

PR #2919, opened 2026-08-01
fix: require explicit TEE signer pinning instead of trusting server-reported list

2TheMoom:
## Summary

Fixes #2695

When `tee-2fa` is enabled and no signers are explicitly pinned via
`.tee_signers()`, `NetworkProverBuilder::build()` silently fetched
signers from `network::tee::get_tee_signers()` and trusted them
directly, with no verification or pinning.

`get_tee_signers()`'s own doc comment already says it works by
"trusting the server to honestly report the list of signers." A
compromised or impersonated TEE server could return an
attacker-controlled address, which would then pass
`verify_tee_proof`'s empty-check while still being the wrong signer,
defeating the purpose of TEE 2FA.

## Changes

- `crates/sdk/src/network/builder.rs`: replaced the silent
  `get_tee_signers()` fallback with a hard panic when `tee-2fa` is
  enabled and no signers were explicitly pinned, directing callers to
  source signers from a trusted config or on-chain source instead.
- `crates/sdk/src/network/tee/mod.rs`: added a `# Security` note to
  `get_tee_signers()`'s doc comment clarifying it returns untrusted
  convenience data and should be cross-checked out-of-band before use.

## Testing

Added `test_build_panics_without_pinned_tee_signers`, which confirms
`.build()` panics with the expected message when `tee-2fa` is enabled
and no signers are pinned. Ran the full `sp1-sdk` unit test suite
(`--features network,tee-2fa`) - all `network::builder` tests pass,
including the pre-existing ones. 8 unrelated failures in
`mock`/`env`/`light` tests are pre-existing local environment issues
(shared-memory space limits), confirmed present on unmodified `dev`
with no changes applied.

`cargo +nightly fmt --all -- --check` passes with no diffs.

## Notes

Went with `panic!` rather than a `Result`-returning refactor to stay
consistent with this block's existing pattern (`.expect(...)` was
already used here). Happy to convert to a `Result` if maintainers
prefer that direction - could be a larger follow-up affecting the
builder's broader error-handling approach.


2TheMoom:
Closing - opened this against a stale fork branch by mistake, which
pulled in 79 unrelated commits into the diff. Reopened cleanly from a
branch based directly on upstream/dev: see #<new PR number>

PR #2928, opened 2026-08-05
ci: bind release/suite workflow inputs via env before shell

SashaMIT:
## Summary
- Bind `inputs.dry_run` / `target_branch` / `prerelease` (and release tag output) through `env:` before shell steps in `release.yml`.
- Bind `github.event.inputs.network_concurrent_repeat_count` through `env:` before the `sp1-perf` shell invocation in `suite.yml`.
- Same class as GitHub’s documented script-injection guidance for interpolating workflow inputs into `run:` scripts.

## Test plan
- [ ] Dry-run release still creates draft tags with run-id suffix
- [ ] `gh release create` flags for prerelease/draft unchanged for trusted inputs
- [ ] Network suite still forwards concurrent-repeat-count

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -27,9 +27,11 @@ runs:
   steps:
     - name: Set up git private repo access
       shell: bash
+      env:
+        PULL_TOKEN: ${{ inputs.pull_token }}
       run: |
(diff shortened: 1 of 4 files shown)


SashaMIT:
Follow-up: also bound `zkevm-sdk-build` (`version` / `release_tag`) and setup `pull_token` through `env:` before shell — same class as the release/suite change.

PR #2933, opened 2026-08-07
fix: install circuit artifacts atomically

Farhad-Shabani:
## Summary

- Downloads each circuit archive into a sibling staging directory and publishes it after extraction succeeds.
- Removes failed staging data and discards completed staging data when another installer publishes first.

## Motivation

The current installer creates the final version directory before the download finishes.
A failed download leaves that directory behind. Later starts treat its presence as a complete installation and reuse partial artifacts.


crates/prover/src/build.rs
```diff
@@ -496,15 +496,44 @@ pub async fn try_install_circuit_artifacts(artifacts_type: &str) -> Result<PathB
...
+        .tempdir_in(parent)?;
+
+    // A failure here drops the staging dir, which deletes the partial download.
+    download_circuit_artifacts(staging_dir.path(), artifacts_type).await?;
+
+    let staging_path = staging_dir.keep();
+    if let Err(err) = std::fs::rename(&staging_path, &build_dir) {
+        std::fs::remove_dir_all(&staging_path)?;
+        // If build_dir exists, another process installed the artifacts first.
+        if !build_dir.exists() {
```
0xernesto: An old failed download can leave this directory with missing files. Maybe add a completion file after extraction, and use the directory only when that file exists.

0xernesto (approved):

PR #2938, opened 2026-08-18
fix(prover): confine tarball extraction to the target directory

### LINKED ISSUE #2533: [H] Tar extraction allows arbitrary file overwrite
### Component

sp1-sdk

- [ ] SP1 SDK
- [ ] cargo prove CLI/sp1up

_No response_

_No response_

### Operating System

None

### Describe the bug

**Context**: crates/sdk/src/install.rs

**Severity**: HIGH

### Description

`install_circuit_artifacts` downloads a remote `.tar.gz` and extracts it by invoking `tar` with `-P` (preserve absolute paths) and without any checks to restrict archive entries to `build_dir`. 

A malicious archive can contain absolute paths (e.g., `/etc/cron.d/...`) and/or `../` traversal paths and will be written outside the intended directory when extracted, resulting in arbitrary file overwrite on the machine running the installer.

**Realistic exploit path**:  
- If the S3 bucket/object is compromised (or a malicious artifact is served under the expected name), running any SDK flow that triggers `try_install_circuit_artifacts` will download and extract the attacker-controlled tarball, overwriting user files (e.g., `~/.bashrc`, `~/.ssh/authorized_keys`) or system files when run with elevated privileges, potentially leading to code execution/persistence.

**Impacted code**
```
let mut res = Command::new("tar")
    .args([
        "-Pxzf",
        artifacts_tar_gz_file.path().to_str().unwrap(),
        "-C",
        build_dir.to_str().unwrap(),
    ])
    .spawn()
    .expect("failed to extract tarball");
res.wait().unwrap();
```

### Recommendation

Avoid shelling out to `tar` for untrusted archives. Prefer a Rust tar/gzip library and explicitly validate each entry path (reject absolute paths, reject `..` components, enforce that the final canonicalized output path is within `build_dir`). 
If keeping `tar`, drop `-P`, add flags/logic to prevent path traversal (e.g., `--no-overwrite-dir`, `--no-same-owner`, and validate archive contents before extraction). 

Also consider extracting into a temporary directory and atomically moving validated files into place.

zexoverz:
Fixes #2533.

A note on where this lives: the issue points at `crates/sdk/src/install.rs`, which is now an eight-line re-export. The implementation is in `crates/prover/src/build.rs`.

There are two extraction sites there, not one. `download_circuit_artifacts` pulls the circuit artifacts over HTTPS, and the dev-artifact path pulls from S3 via `aws s3 cp`. Both passed `-P`, so both accepted absolute member names and `..` components. Only the first is mentioned in the issue.

Both now go through a shared `extract_tarball` helper that omits `-P`. Nothing needed it. The tarballs are produced with relative paths by the upload path in this same file (`tar -czf ... -C <dir> .`), and the repository's other two extraction sites, the CUDA server download and the toolchain installer, already extract without `-P`. This brings the outliers in line rather than introducing a new convention.

Two tests build genuinely hostile archives, one with a `../` member and one with an absolute member, and assert neither writes above the extraction directory. I checked they fail with `-P` restored, so they pin the behaviour rather than just passing. The absolute-path test demonstrates the overwrite directly: without the fix, a file outside the extraction directory is reverted to the archive's copy.

The two tars behave differently on the hostile input, so the tests assert only the property that matters. GNU tar strips the offending components and continues; bsdtar refuses the entry and exits non-zero. Both confine the write, which is what is asserted.

I have not attempted to address symlink-based escapes, which are a separate class and would mean moving off the `tar` binary entirely. Happy to look at that separately if you want it.


zexoverz:
Ten days in and no workflow has run on this yet. Could a maintainer approve the run? It is a path traversal fix so I would rather it not sit.

PR #2944, opened 2026-08-23
fix(executor): reject non-canonical Poseidon2 precompile inputs

4waan:
Depends on #2935. Merge that first.

Both PRs add code to `crates/core/executor/src/minimal/tests.rs`. Once #2935 lands I rebase this branch onto main

## Motivation

`SP1Field::from_canonical_u32` enforces its range check with a `debug_assert`, so in release it is `from_wrapped_u32`: the executor silently reduced a guest state word `>= P` and hashed a value the guest never wrote. The AIR does not accept that word (`input_range_checkers`, constrained on `is_real`), so the execution was unprovable, yet it ran to completion with no diagnostic. The only symptom was a constraint failure inside the Poseidon2 chip at proving time, naming neither the syscall nor the guest. Trace generation shares the unguarded conversion,
so it could not catch the case either.

Guests reach this through `Poseidon2State::absorb_field_block_unchecked`, whose contract was unenforced and whose stated consequence ("the proof will not verify") was not what actually happened.

Separately, `RiscvAir::Poseidon2` and both `sp1-lib` hashing surfaces shipped without tests of any kind.

## Solution

- Check the words before converting, so the diagnostic does not depend
  on the `debug_assert` being compiled in.
- Panic rather than return an `Interrupt`: the AIR has no trap
  representation for a non-canonical input, so there is nothing provable
  to return. Same reasoning as the decommissioned decompress handlers.
  The check sits after `read_write_slice_check`, so page protection
  traps behave exactly as before.
- One site covers every executor: the x86_64 backend registers the same
  `sp1_ecall_handler`, and the VM syscall path replays records rather
  than computing. Poseidon2 is the executor's only guest-controlled
  `from_canonical_u32`; the `lift_x` calls in `program.rs` are
  mask-bounded to 24 bits or fewer.
- New coverage: a guest fixture driving both hashing surfaces, host/guest
  agreement tests, a proving test for `RiscvAir::Poseidon2`, AIR
  operation unit tests via a test-only single-row
  `DebugConstraintBuilder`, and koala-bear constant tests.
- The canonicality test passes exactly `P`, which reduces to zero, so the
  old behaviour looked like an ordinary hash of an all-zero block. The
  agreement test beside it absorbs `P-1`, pinning the boundary from both
  sides.

No working behaviour changes: every execution this rejects was already unprovable.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation


4waan:
Trace generation at `.../precompiles/poseidon2/air.rs` still converts the guest word unguarded. It is unreachable now that the executor halts first and guarding it would move the panic into the prover. Happy to record the invariant there in a comment if preferred.

PR #2946, opened 2026-08-24
ci(gpu): guard the CUDA Poseidon2 constants against drift

4waan:
## Motivation

both `poseidon2_kb331_16.cuh` and `poseidon2_bn254_3.cuh` hold duplicate of Poseidon2 constants
the rust side (vs the fact that `StaticHasher` in `poseidon2.cuh` binds those tables directly) already derives.. 

Something had to tie these two copies together. Editing a constant on one side produces no compile error and no test failure. The GPU simply computes a different permutation than the CPU, and the first symptom is a proof that does not verify.

## Solution

Parse the constant tables out of both headers in `sp1-gpu-sys`'s build script and compare them against `slop-koala-bear` and `slop-bn254`. A drifted constant becomes a build failure naming the exact entry:

```
include/poseidon2/poseidon2_bn254_3.cuh: MAT_INTERNAL_DIAG_M1[2] is
[..., 471134084], but Rust has [..., 471134083]. The CUDA and Rust Poseidon2
constants have drifted; fix whichever side changed.
```
### Verification

- `cargo test -p slop-bn254` passes, including two tests pinning the new helpers to `R` and to `[one, one, two]`.
- `cargo check -p sp1-gpu-sys` runs the guard against the real headers on a machine with no CUDA.
- Perturbing one limb of `MAT_INTERNAL_DIAG_M1[2]` fails the build with the message above.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7776,6 +7776,8 @@ dependencies = [
  "cbindgen",
  "cmake",
  "pathdiff",
+ "slop-algebra 6.4.0",
+ "slop-bn254 6.4.0",
  "slop-koala-bear 6.4.0",
  "sp1-core-executor",
  "sp1-core-machine",
--- a/slop/crates/bn254/src/lib.rs
+++ b/slop/crates/bn254/src/lib.rs
@@ -11,12 +11,13 @@
 #![allow(clippy::disallowed_types)]
 use std::marker::PhantomData;
 
+use ff::PrimeField as FFPrimeField;
 pub use p3_bn254_fr::*;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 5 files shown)


4waan:
@erabinov since this builds directly on #2871, id be glad for your review here.. 

The guard compared only round constants. BN254's `MAT_INTERNAL_DIAG_M1` is live, read on every one of the 56 internal rounds.. while KoalaBear's is genuinely dead. checked up with that asymmetry.. with the diagonal read back out of `DiffusionMatrixBN254` rather than transcribed a third time.

I have only exercised the guard with CUDA absent, so a confirmation that it behaves on an actual GPU build would help!

PR #2951, opened 2026-08-27
chore: remove dead code and unused dependencies

bergkvist:
## Motivation

The workspace carried dependency declarations and code that nothing used.

`cargo-shear` found 66 unused dependency declarations across 26 crates. Some cost build time for guest programs. `sp1-verifier` declared `dirs` under `[build-dependencies]` but has no `build.rs`, and several `slop` crates declared `futures` without using it. Those entries pulled `dirs`, the `futures` crates, `redox_users`, and the `windows-sys` crates into every guest program build.

Four items of dead code hid behind `#[allow(dead_code)]`, so the compiler could not report them.

## Solution

Delete the dead code:

- `ForkState` in `sp1-core-executor`. Nothing refers to it.
- `Page<V>` in `sp1-core-executor`. `NewPage<V>` replaced it, and it was the last user of `vec_map` in that crate.
- `default_pcs_verifier` in `slop-veil`.

Then remove the declarations that dead code kept alive, the rest of the unused declarations, 20 dev-dependency entries that repeat an identical `[dependencies]` entry, and 7 unused entries in the root `[workspace.dependencies]`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/core/machine/src/syscall/precompiles/weierstrass/weierstrass_decompress.rs
```diff
@@ -705,107 +705,3 @@ where
         );
     }
 }
-
-// #[cfg(test)]
```
erabinov, line 709: Let's keep these commented-out tests here. Ideally, we reinstate it at some point.

erabinov (approved):
LGTM!

PR #2955, opened 2026-09-03
feat(sdk): derive serde for SP1ProvingKey

Farhad-Shabani:
## Summary

- Add Serialize and Deserialize support to Elf with the same byte-sequence format.
- Return `Elf::Dynamic` after deserialization and test both source variants.
- Derive both serde traits for `SP1ProvingKey`.

### DIFF
--- a/crates/sdk/src/prover/prove.rs
+++ b/crates/sdk/src/prover/prove.rs
@@ -1,11 +1,12 @@
 use super::{IntoSendFutureResult, Prover};
 use crate::{ProvingKey, SP1ProofMode, SP1ProofWithPublicValues, StatusCode};
+use serde::{Deserialize, Serialize};
 use sp1_build::Elf;
 use sp1_core_executor::SP1ContextBuilder;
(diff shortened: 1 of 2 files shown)


crates/sdk/src/prover/prove.rs
```diff
@@ -5,7 +5,7 @@ use sp1_core_executor::SP1ContextBuilder;
 use sp1_core_machine::io::SP1Stdin;
 use sp1_prover::SP1VerifyingKey;
 
-#[derive(Clone)]
+#[derive(Clone, serde::Serialize, serde::Deserialize)]
```
erabinov: nit: import Serialize and Deserialize at top of file.

Farhad-Shabani: [e7ecc4e](https://github.com/succinctlabs/sp1/pull/2955/commits/e7ecc4e7cb329a961c5e16e61143a84639014da1)

erabinov (approved):
LGTM!

PR #2960, opened 2026-09-07
fix(sdk): preserve network simulation errors

Kemperino:
## Motivation

Network proof requests discard the underlying local simulation error and return only `Program simulation failed`. This makes guest panics and executor failures difficult to diagnose.

## Solution

Preserve the original `anyhow::Error` as the source of `SimulationFailed` and include its error chain in the displayed message. Add a unit test covering the resulting message and source.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

`Error::SimulationFailed` changes from a unit variant to a tuple variant containing the source error.

### DIFF
--- a/crates/sdk/src/network/prover.rs
+++ b/crates/sdk/src/network/prover.rs
@@ -890,7 +890,7 @@ impl NetworkProver {
             .node
             .execute(elf, stdin.clone(), SP1Context::builder().calculate_gas(true).build())
             .await
-            .map_err(|_| Error::SimulationFailed)?;
+            .map_err(Error::SimulationFailed)?;
 
         let (_, committed_value_digest, report) = execute_result;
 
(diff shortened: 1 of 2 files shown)


crates/sdk/src/network/prover.rs
```diff
@@ -890,7 +890,7 @@ impl NetworkProver {
             .node
             .execute(elf, stdin.clone(), SP1Context::builder().calculate_gas(true).build())
             .await
-            .map_err(|_| Error::SimulationFailed)?;
+            .map_err(Error::SimulationFailed)?;
```
Farhad-Shabani: Could we keep `SimulationFailed` as a unit variant and use `.context(Error::SimulationFailed)` here instead? That should preserve the error chain without breaking the public API.

Kemperino: yes agree - strictly better considering we have anyhow error chain available.

Farhad-Shabani (approved):

PR #2971, opened 2026-09-14
test(patch-testing): add sha2 0.11.0

sebastianst:
## Why

`revm-precompile` 42 requires `sha2 = "0.11"`. SP1 accelerates SHA-256 by replacing the `sha2`
crate through `[patch.crates-io]`, and a patch only applies to the exact version it names — so
`patch-sha2-0.10.9-sp1-6.2.0` does nothing for a graph that resolves `sha2` to 0.11. Programs that
pull in revm 42 silently drop to the soft backend and stop emitting `SHA_EXTEND` / `SHA_COMPRESS`.

The 0.11 patch itself landed as https://github.com/sp1-patches/RustCrypto-hashes/pull/10 and is
tagged `patch-sha2-0.11.0-sp1-6.2.0`. This PR adds the `patch-testing` coverage for it, so 0.11 is
validated the same way every other `sha2` version already is. Tracking issue on our side:
https://github.com/ethereum-optimism/optimism/issues/22890.

## What

A `sha2_v0_11_0` case mirroring `sha2_v0_10_9`:

- `sha/program/Cargo.toml`: optional `sha2-v0-11-0` dep, `v0-11-0` feature, `[[bin]]`, and the
  `[patch.crates-io]` entry that swaps in the patched crate for the guest.
- `sha/program/bin/sha2.rs`: the `extern crate … as sha2` alias behind `v0-11-0`.
- `sha/build.rs`: build the guest for the new feature.
- `sha/src/lib.rs`: `#[sp1_test("sha2_v0_11_0", syscalls = [SHA_COMPRESS, SHA_EXTEND], gpu, prove)]`.
- `Cargo.toml` / `build-host/Cargo.toml`: `sha2-v0-11-0-patched`, so the patch is also compiled for
  the host target alongside the other patched crates.

Both patch sources point at the released `patch-sha2-0.11.0-sp1-6.2.0` tag, written the same way as
the neighbouring 0.10.9 entries.

`sha2` 0.11 moves to `digest` 0.11, but the surface this test uses is unchanged — `Digest::new`,
`update`, and `finalize().into()` to `[u8; 32]` all still work — so `bin/sha2.rs` needed nothing
beyond the new alias, and the host-side expected-digest helper is shared verbatim with the 0.10.9
case.

## Validation

Run locally against this branch, SP1 toolchain from `sp1up` (`cargo-prove sp1 58c4aea`, `succinct`
rustc 1.96.0-dev), CPU executor, no proving:

```
cargo test --manifest-path patch-testing/Cargo.toml -p sha --release -- --test-threads=1
```

6 passed, 0 failed, as does `cargo check -p build-host --release`. The new case emits both
precompiles and matches the host digests (cycle counts vary slightly run to run, since part of the
corpus is random):

| case | cycles |
| --- | --- |
| `test_sha2_v0_11_0_expected_digest_lte_100_times` | 1,336,295 |
| `test_sha2_v0_10_9_expected_digest_lte_100_times` | 1,347,444 |

An extra ad-hoc run over a corpus covering the empty input, one-block, multi-block, and inputs
sliced at odd offsets (to hit the unaligned-first-block branch of the patch) gave byte-identical
digests to host `sha2` 0.11 and identical syscall counts for 0.11 and 0.10.9:
`SHA_COMPRESS = 117`, `SHA_EXTEND = 117`.

🤖 *Co-created with Claude Fable 5.1*


fakedev9999 (approved):
lgtm

PR #2977, opened 2026-09-18
feat(sdk): support dynamic network signers

wwared:
Allows callers to provide an asynchronous Alloy signer as a network signer, enabling external signing services without changing existing local or AWS KMS behavior.


0xernesto (changes requested):
It seems that this would be a breaking change for users depending on NetworkSigner. Instead, you can do something like:
```rust
#[derive(Clone)]
pub(crate) enum SignerSource {
    Existing(NetworkSigner),
    Dynamic(Arc<dyn Signer + Send + Sync>),
}
```
and use it like this:
```rust
let prover = ProverClient::builder()
    .network()
    .dynamic_signer(Arc::new(custom_signer))
    .build()
    .await;
```
This would keep NetworkSigner and the existing public method signatures unchanged. We can consolidate these changes in a future breaking change (eg, new major release).

wwared:
@0xernesto That's a good point. I've refactored the PR to make it fully backwards compatible. The only new public API change is adding `dynamic_signer` to the builder type, every other change is fully internal.

As a result the PR touches more of the SDK internals, but it's still pretty straightforward. Let me know if this works!

0xernesto (approved):

